Issue #512, opened 2021-02-12
Keep the top-level resource stable within a major version


mikhailshilkov:
Design and implement the mechanism to keep the top-level resources stable within a major version.

According to our plan in https://github.com/pulumi/pulumi-azure-nextgen/issues/169#issuecomment-777458268:

At 1.0, for a given resource, we will likely pick an API version that is the latest. If a later version is published by Microsoft and it contains breaking changes, we will NOT introduce this version to the top-level resource. It will be available under a versioned namespace only.

One obvious way is to stop updating [`provider_list.json`](https://github.com/pulumi/pulumi-azure-nextgen-provider/blob/master/azure-provider-versions/provider_list.json) but let's spend some time considering pros/cons and other options.

Issue #127, opened 2020-08-21
Breaking changes in API


mikhailshilkov:
Let's collect breaking changes in stable API as we upgrade periodically.

The first batch (see #122 for discussion):

- CDN removed properties from a few latest versions [PR in specs](https://github.com/Azure/azure-rest-api-specs/pull/10126). In the description, they claim that those versions weren't released yet, so apparently we need to adopt a notion of release in our pipeline. A working theory is that a version isn't released unless it's published in the "README". And indeed, those versions were not in [here](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/cdn/resource-manager/readme.md) until a week ago. We should validate this across the repo.

- Security renamed their types to avoid the word "whitelist" [PR in specs](https://github.com/Azure/azure-rest-api-specs/pull/10100). They confirm it's breaking but shipped for language-usage reasons.

- Storage cache fixed something about discrimination properties [PR is specs](https://github.com/Azure/azure-rest-api-specs/pull/10392). There's no explanation in the PR, so I asked for one, but it looks like this may be a bug fix in their terms, see this [other PR](https://github.com/Azure/azure-rest-api-specs/pull/10383).

In all three cases, Microsoft's automation detected the changes as breaking but the reviewers went ahead regardless.

Two new breaking changes on 2020-08-24 (see #131):

- C# namespace alias changed from `AAD` to `Aad` this is because [a new spec file](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/domainservices/resource-manager/Microsoft.AAD/stable/2017-06-01/oucontainer.json) was [added](https://github.com/Azure/azure-rest-api-specs/pull/10508) and our codegen took the alias from this new file, not the old file. If this happens after the release, we may want to pin the alias in our code.

- A resource `SubscriptionAlias` is renamed to `Alias`. This is a breaking change and clearly marked so in the [original PR](https://github.com/Azure/azure-rest-api-specs/pull/10486). They let it through because
    > This is newly added version which is added last week. No SDK has been released yet for this version and API is not live yet for real customers, only internal testing users are using it.

    I guess it would be okay for us to release this breaking change with the same note.

Another breaking change on 2020-08-27 (see #140):

[...]

PR #634, opened 2021-03-05
Introduce locked versions for top-level resources

mikhailshilkov:
A very simple approach to fix https://github.com/pulumi/pulumi-azure-native/issues/512

Basically, the suggested default mode of operation:

- When a new API version comes out, we'll update our top-level resources to it automatically (that already works)
- We manually review every update for code-level breaking changes (I already do, see e.g. #127)
- If the review identifies that a new version brought a breaking change, we add an entry to the `lockedTypeVersions` map below and re-run the generation
- We assume that the new API version doesn't bring service-level breaking changes that we would want to avoid. We don't really have a way to test for this, so our options are a) take all new versions and rely on MS not breaking the behavior b) don't take any new versions at all. I feel like (b) would be too detrimental to the user experience.
- Since the review is a manual process, I'm choosing to hard-code the map in code. There's no way to generate it anyway.

We may want to invest in a tool that compares two schemas and detects breaking changes. I'll open an issue once we land this PR.

We'll start applying this policy now and will still have time to learn and iterate if needed before the GA.

### DIFF
--- a/provider/pkg/openapi/versioner.go
+++ b/provider/pkg/openapi/versioner.go
@@ -66,7 +66,7 @@ func newVersioner() (*versioner, error) {
 	return &versioner{lookup: result}, nil
 }
 
-// A manually-maintained list of stable versions where we want to promote a later preview version to be used for
+// A manually-maintained list of stable versions that we want to promote a later preview version to be used for
 // the top-level resource. These versions are "known" to be outdated but no newer stable versions were released yet.
 // However, there's no formal way to derive this from Open API specs, so we have to maintain this manual map.
 var deprecatedProviderVersions = map[string][]string{


lblackstone (approved):

provider/pkg/openapi/versioner.go
```diff
@@ -74,6 +74,13 @@ var deprecatedProviderVersions = map[string][]string{
 	"Sql": {"v20140401", "v20150501"},
 }
 
+// A manually-maintained list of versions where we want to use for the top-level resource. The primary goal is to
```
viveklak, line 77:
```suggestion
// A manually-maintained list of versions that we want to use for the top-level resource. The primary goal is to
```

viveklak (approved):

Issue #643, opened 2021-03-08
Template deployment fails due to auto-location feature


DaRosenberg:
Template deployments using the `Pulumi.AzureNative.Resources.Deployment` resource cannot be done at resource group scope due to the auto-location feature incorrectly adding a `location` property to the deployment.

## Expected behavior

Resource should deploy successfully, as long as I don't add any `DeploymentArgs.Location`property or add a `location` property to the template JSON myself.

## Current behavior

Deployment fails with error message:

```
Diagnostics:
  azure-native:resources:Deployment (hostname-binding-template):
    error: Code="InvalidDeployment" Message="The 'location' property is not allowed for a deployment at resource group scope. Please see https://aka.ms/deploy-to-subscription for usage details."
```

## Steps to reproduce

1. Create a resource group scoped template deployment resource using the new native provider (template and parameter values should be irrelevant as long as they are otherwise valid)
2. Do `pulumi up --skip-preview`
3. Error message appears

## Context (Environment)

There are still things requiring template deployments even with new provider. This issue makes it impossible to use them at all with the new provider, as far as I can tell.

## Affected feature

Azure template deployment resources, at resource group scope.

PR #654, opened 2021-03-10
Do not auto-populate deployment location

mikhailshilkov:
Fix https://github.com/pulumi/pulumi-azure-native/issues/643

### DIFF
--- a/provider/pkg/resources/resources.go
+++ b/provider/pkg/resources/resources.go
@@ -28,11 +28,6 @@ const (
 	AutoNameCopy AutoNameKind = "copy"
 )
 
-type AutoLocationKind string
-const (
-	AutoLocationOff AutoLocationKind = "off"
-)
-
 // AzureAPIProperty represents validation constraints for a single parameter or body property.
 type AzureAPIProperty struct {
 	Type                 string            `json:"type,omitempty"`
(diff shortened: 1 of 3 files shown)


provider/pkg/resources/resources.go
```diff
@@ -87,6 +92,9 @@ type AzureAPIResource struct {
 	ReadMethod string `json:"readMethod,omitempty"`
 	// Path to append to the resource ID to produce the URL for to read resource state.
 	ReadPath string `json:"readPath,omitempty"`
+	// By default, we populate the `location` property of every resource to the location of its resource
+	// group or the configured value. This property can override this default behavior.
+	AutoLocation AutoLocationKind `json:"autoLocation,omitempty"`
```
viveklak, line 97: Nit - switch to `DisableAutoLocation bool` instead of a string/enum since the values are clearly binary?

viveklak (approved):

provider/pkg/resources/resources.go
mikhailshilkov: I started with this but then I thought to keep it closer to auto-naming. I don't see this appeal anymore - so I'll apply your suggestion, thank you.

PR #655, opened 2021-03-11
Add ignoreChange support to provider

### LINKED ISSUE #645: Unable to ignore changes on `enableNodePublicIP` AgentPool property.
Changes to the `enableNodePublicIP` resource property cannot be ignored with `IgnoreChanges` and this appears to be because of the `IP` vs `Ip` casing.

## Expected behavior

I should be able to ignore changes to `enableNodePublicIP` in the same manner as `enableAutoScaling` and other resource properties.

## Current behavior

No matter what I set in `IgnoreChanges` I cannot get pulumi to ignore the property.

## Steps to reproduce

Unfortunately I don't have a reproduction that doesn't require editing the state file, but these edits are representative of a situation a customer is currently in where `enableNodePublicIP` is in their state file but was not set on the resource.

1. `git clone https://gist.github.com/fb0d4d8938600eecab52ff126ee3412a p-azure-native-agentpool-publicip`
1. `cd p-azure-native-agentpool-publicip`
2. `pulumi stack init dev`
3. `pulumi up`
    - this should provision successfully
4. Uncomment `// EnableNodePublicIP: pulumi.Bool(true)` for the `NewAgentPool` resource.
5. `pulumi pre --diff`
    - observe this previews fine without a change to the `enableNodePublicIP` property
6. `pulumi stack export --file export.json`
7. Open `export.json` and add `"enableNodePublicIP": false,` to the `inputs` for the `azure-nextgen:containerservice/v20200701:AgentPool` resource.
7. `pulumi stack import --file export.json`
8. `pulumi pre --diff`
    - observe this shows an update to the `enableNodePublicIP` property, but should not

**Note**: Rather than setting `"enableNodePublicIP": false,` in the state file, set `"enableNodePublicIp": false,` (`Ip` vs `IP`) and the preview will not show a change (expected behavior), hence why I think this could be a case-sensitivity related problem.

## Context (Environment)

```
% pulumi version
v2.21.2
% cat go.mod 
module goaks

go 1.14

require (
        github.com/pulumi/pulumi-azure-nextgen/sdk v0.5.1
        github.com/pulumi/pulumi-azuread/sdk/v2 v2.4.0
        github.com/pulumi/pulumi-kubernetes/sdk/v2 v2.8.2
        github.com/pulumi/pulumi-random/sdk/v2 v2.2.0
        github.com/pulumi/pulumi-tls/sdk/v2 v2.2.0
        github.com/pulumi/pulumi/sdk/v2 v2.21.2
)
```

viveklak:
Fixes #645

The specific change here should only be necessary if there is a discrepancy between `inputs` and `__inputs` in state.

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -674,6 +674,9 @@ func (k *azureNativeProvider) Diff(_ context.Context, req *rpc.DiffRequest) (*rp
 			oldState, newResInputs)
 	}
 
+	// Expunge any modifications for fields marked to be ignored.
+	newResInputs = processIgnoreChanges(newResInputs, oldInputs, req.GetIgnoreChanges())
+
 	// Calculate the difference between old and new inputs.
 	diff := oldInputs.Diff(newResInputs, func(key resource.PropertyKey) bool {
 		return strings.HasPrefix(string(key), "__")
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/provider.go
```diff
@@ -674,6 +674,9 @@ func (k *azureNativeProvider) Diff(_ context.Context, req *rpc.DiffRequest) (*rp
 			oldState, newResInputs)
 	}
 
+	// Expunge any modifications for fields marked to be ignored.
+	newResInputs = processIgnoreChanges(newResInputs, oldInputs, req.GetIgnoreChanges())
```
mikhailshilkov, line 678: So, is the problem in `newResInputs` or `oldInputs`? From what I understood - in the former, and then it feels wrong to do this in the provider?

provider/pkg/provider/provider.go
```diff
@@ -674,6 +674,9 @@ func (k *azureNativeProvider) Diff(_ context.Context, req *rpc.DiffRequest) (*rp
...
+
 	// Calculate the difference between old and new inputs.
 	diff := oldInputs.Diff(newResInputs, func(key resource.PropertyKey) bool {
```
mikhailshilkov, line 681: Should we use this `func` to filter out the ignored changes? This is sort of the purpose of this callback.

mikhailshilkov (commented):
I want to understand what exactly we are fixing... I don't yet.

> if there is a discrepancy between inputs and __inputs in state.

`__inputs` are old inputs and `inputs` are new inputs, so having a discrepancy is normal there. The engine is supposed to control the shape of `inputs`. Also, they should be coming from a program execution, not from the state file.

Were you able to reproduce the issue?

viveklak:
> Were you able to reproduce the issue?
I was able to reproduce the situation locally when there is an errant entry in the `inputs` section of the state file but only if there is a code change or the field figures in `ignoreChanges`. Without either of these the update results in no diffs. As discussed on ticket this seems to be working as designed.

Closing this PR.

Issue #625, opened 2021-03-04
RoleDefinition and RoleAssignment - GUIDs required, auto-naming submits auto name


shawnjones253:
Tested using Pulumi.AzureNative.Authorization.V20180101Preview and Pulumi.AzureNative 0.7.1

When trying to create a `RoleDefinition` or `RoleAssignment`, both resources expect a GUID for their identifiers but AzureNative tries to use it's auto-generated random name for these properties, causing Azure to reject the request with 400.

`RoleAssignment`:
```
var assignment = new RoleAssignment(roleAssignmentName,
                            new RoleAssignmentArgs
                            {
                                // commented line below was the fix, needed to use a pulumi random uuid
                                // RoleAssignmentName = roleAssignmentUuid.Result,
                                Scope = Output.Tuple(storageAccountName, queue, resourceGroupName).Apply(async t =>
                                {
                                    var (storageAccountName, queue, resourceGroupName) = t;
                                    return await GetQueue.InvokeAsync(new GetQueueArgs()
                                    {
                                        AccountName = storageAccountName!,
                                        QueueName = queue!,
                                        ResourceGroupName = resourceGroupName!
                                    });
                                }).Apply(v => v.Id),
                                RoleDefinitionId = definition.Id,
                                PrincipalId = group.ObjectId,
                            }
```

```
azure-native:authorization/v20180101preview:RoleAssignment (assign_alertQ_0_0):
    error: cannot check existence of resource '/%2Fsubscriptions%sub-id%2FresourceGroups%2Frg-name%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Fstorage-acct-name%2FqueueServices%2Fdefault%2Fqueues%2Fqueue-name/providers/Microsoft.Authorization/roleAssignments/assign_alertQ_0_07848abbb': status code 400, {"error":{"code":"InvalidRoleAssignmentId","message":"The role assignment ID 'assign_alertQ_0_07848abbb' is not valid. The role assignment ID must be a GUID."}}
```

`RoleDefinition`:
```
var role = new RoleDefinition(roleDefinitionName,
                    new RoleDefinitionArgs
                    {
                        // commented line below was the fix, needed to use a pulumi random uuid
                        // RoleDefinitionId = readerRoleUuid.Result,
                        // specifying RoleDefinitionId also forced me to manually specify RoleName as below
                        //RoleName = $"{roleDefinitionName}{stackName}",
                        Scope = args.PolicyScope,
[...]

PR #667, opened 2021-03-15
Autoname UUID properties

mikhailshilkov:
Fix https://github.com/pulumi/pulumi-azure-native/issues/625


provider/pkg/gen/schema.go
```diff
@@ -878,6 +878,10 @@ func (m *moduleGenerator) genMethodParameters(parameters []spec.Parameter, ctx *
 		result.parameters = append(result.parameters, apiParameter)
 	}
 
+	// Remove the auto-name property from the list of required specs. This happens after we processed all parameters
+	// to account for several cases where the same name is both a method parameter and a body property.
+	result.requiredSpecs.Delete(autoNamedSpec)
```
viveklak, line 883: You can have multiple autonamed parameters right? Won't this only affect the last one? Should `autonamedSpec` be a slice instead?

mikhailshilkov: Maybe in the future?

Currently, this condition ensures it's the last parameter of a path, so it's a single one: https://github.com/pulumi/pulumi-azure-native/pull/667/files#diff-f6a4524eb41046270bf7a4c1f69fba0b49016aacd16db55b907da28e1cb78c59R228

viveklak (approved):

PR #675, opened 2021-03-18
Add some basic arm2pulumi coverage reporting

viveklak:
Fixes: https://github.com/pulumi/arm2pulumi/issues/17

This adds a scaffold to check coverage on a subset of the Azure quickstart templates and some insight into the most common issues hit during conversions.

This test is only run when explicitly invoked. Running `make arm2pulumi_coverage_report` should currently return the following (without any bug fixes):
```
+------------------+--------+--------+
|      RESULT      | NUMBER |  PERC  |
+------------------+--------+--------+
| No Errors        |     90 | 23.56% |
| Low Sev. Errors  |    229 | 59.95% |
| High Sev. Errors |     59 | 15.45% |
| Fatal            |     50 | 13.09% |
| Total            |    382 |        |
+------------------+--------+--------+
Overall Summary of Conversions


+------------+----------------------------------+
| # OF TIMES |              REASON              |
+------------+----------------------------------+
|          8 | failed to evaluate               |
|            | template expression for          |
|            | azureFirewallResource:           |
|            | missing variable:                |
|            | azureFirewallIpConfigurationsVar |
+------------+----------------------------------+
|          2 | failed to evaluate               |
|            | template expression for          |
|            | sqlVirtualMachineResource:       |
|            | missing variable:                |
|            | dataDisksLUNsVar                 |
+------------+----------------------------------+
|          2 | failed to evaluate               |
|            | template expression for          |
|            | serviceResource: missing         |
|            | variable: modelsVar              |
+------------+----------------------------------+
|          2 | failed to evaluate               |
|            | template expression for          |
|            | publicIPAddressResource0:        |
|            | missing variable:                |
|            | lbPublicIPAddressNameVar         |
+------------+----------------------------------+
|          2 | failed to evaluate               |
|            | template expression for          |
|            | managedClusterResource:          |
|            | missing variable:                |
|            | vnetSubnetIDVar                  |
+------------+----------------------------------+
Top Reasons For Fatal Errors


+------------+-------------------------------------------------------------+
| # OF TIMES |                           REASON                            |
+------------+-------------------------------------------------------------+
|         19 | no metadata found                                           |
|            | for resource type                                           |
|            | 'Microsoft.Web/connections'                                 |
|            | and version                                                 |
|            | '2018-07-01-preview'                                        |
+------------+-------------------------------------------------------------+
|          9 | no metadata found for resource type                         |
|            | 'Microsoft.DataFactory/datafactories/linkedservices'        |
|            | and version '2015-10-01'                                    |
+------------+-------------------------------------------------------------+
|          9 | no metadata found for resource type                         |
|            | 'Microsoft.DataFactory/datafactories/datasets'              |
|            | and version '2015-10-01'                                    |
+------------+-------------------------------------------------------------+
|          7 | no metadata found                                           |
|            | for resource type                                           |
|            | 'Microsoft.Web/sites/config'                                |
|            | and version '2019-08-01'                                    |
+------------+-------------------------------------------------------------+
|          6 | no metadata found for resource type                         |
|            | 'Microsoft.Storage/storageAccounts/blobServices/containers' |
|            | and version '2019-06-01'                                    |
+------------+-------------------------------------------------------------+
```


Makefile
```diff
@@ -50,6 +50,10 @@ generate_docs::
 arm2pulumi::
 	(cd provider && go build -a -o $(WORKING_DIR)/bin/arm2pulumi $(VERSION_FLAGS) $(PROJECT)/provider/cmd/arm2pulumi)
 
+arm2pulumi_coverage_report::
+	(cd provider/pkg/arm2pulumi/internal/testdata && if [ ! -d azure-quickstart-templates ]; then git clone https://github.com/Azure/azure-quickstart-templates; fi)
```
mikhailshilkov, line 54: Is it a git submodule now? If so, we shouldn't probably clone here? Or don't you intend to check these files in?

provider/go.mod
```diff
@@ -24,4 +25,5 @@ require (
 	github.com/tombuildsstuff/giovanni v0.15.1
 	github.com/zclconf/go-cty v1.3.1
 	google.golang.org/grpc v1.29.1
+	modernc.org/sqlite v1.10.0
```
mikhailshilkov, line 28: Does this affect the size/compile time of the provider?

Makefile
viveklak: I didn't use a submodule to avoid affecting the rest of build/test time. This target is entirely self-contained right now and nothing is checked in. 

I don't think this needs to be a regression test - more of an assessment of coverage at a point-of-time. We can consider adding it in as a regression test once we improve coverage.

provider/go.mod
viveklak: Its a small native Go driver for sqlite which is only run from the new integration test (which is only used when `arm2pulumi_coverage_report` target is called). Shouldn't affect the binary at all.

Makefile
mikhailshilkov: I'm confused by this then: https://github.com/pulumi/pulumi-azure-native/pull/675/files#diff-229db5321902ca0aea52dd9c7b76760d0f11540d42d5e35a5d29932949fd23c3

image

viveklak: Yeah I started with a submodule but removed it or at least I thought I did. Seems a submodule reference leaked through. Hopefully fixed now.

mikhailshilkov (approved):
LGTM

joeduffy:
Is there any way to run this in CI/CD automatically, and fail the job if the number regresses?

viveklak:
> Is there any way to run this in CI/CD automatically, and fail the job if the number regresses?

Tracked in https://github.com/pulumi/pulumi-azure-native/issues/703

PR #710, opened 2021-04-01
initial version of CI for arm2pulumi coverage test via GitHub Actions

hiteshboinpally:
Testing if the workflow currently works as intended
- Adding creation of `results.json` file that stores overall results. 
    - Stored in JSON to be easily manipulated in the future (e.g. wanting to measure how the coverage has changed from the previous day)
- For reference: current runtime of coverage test is ~2.5 minutes

### DIFF
--- a/provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
+++ b/provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
@@ -29,11 +29,11 @@ var excluded = map[string]bool{
 	"../testdata/azure-quickstart-templates/101-logic-app-sql-proc/azuredeploy.json":          true,
 	"../testdata/azure-quickstart-templates/101-private-endpoint-sql/azuredeploy.json":        true,
 	// Following templates are removed bc of non-deterministic output: https://github.com/pulumi/arm2pulumi/issues/28
-	"../testdata/azure-quickstart-templates/101-asev2-appservice-sql-vpngw/azuredeploy.json":  true,
-	"../testdata/azure-quickstart-templates/101-functions-managed-identity/azuredeploy.json":  true,
-	"../testdata/azure-quickstart-templates/101-redis-cache/azuredeploy.json": 				   true,
-	"../testdata/azure-quickstart-templates/101-sql-logical-server/azuredeploy.json": 		   true,
-	"../testdata/azure-quickstart-templates/101-custom-rp-with-function/azuredeploy.json":	   true,
+	"../testdata/azure-quickstart-templates/101-asev2-appservice-sql-vpngw/azuredeploy.json": true,
+	"../testdata/azure-quickstart-templates/101-functions-managed-identity/azuredeploy.json": true,
+	"../testdata/azure-quickstart-templates/101-redis-cache/azuredeploy.json":                true,
+	"../testdata/azure-quickstart-templates/101-sql-logical-server/azuredeploy.json":         true,
+	"../testdata/azure-quickstart-templates/101-custom-rp-with-function/azuredeploy.json":    true,
 }
 
 func TestQuickstartTemplateCoverage(t *testing.T) {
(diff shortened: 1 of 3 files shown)


provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
```diff
@@ -162,6 +172,23 @@ CREATE TABLE errors(
...
+		HighSevErrors: Result{highSevErrors, float32(highSevErrors)/float32(numTemplates)*100.0},
+		Fatal: Result{fatalErrors, float32(fatalErrors)/float32(numTemplates)*100.0},
+		Total: numTemplates,
+	}
+
+	file, err := json.MarshalIndent(data, "", "\t")
+	require.NoError(t, err)
+	
+	// Stores JSON result in "results.json" file in current directory.
+	err = ioutil.WriteFile("results.json", file, 0600)
```
viveklak, line 188: How about we create a directory called results and store files with date/time formatted json files. e.g. `results/21-04-06-16:00.json` or something?

viveklak: Perhaps answering my own question - keeping the file name consistent should avoid the possibility of dirtying the worktree in CI - which would also be important. We could rely on git history to get historical state instead.

hiteshboinpally: Yeah I can try to do that. I was just testing this out for the time being and was gonna reach out about how we would like this to be formatted. Does the current actual JSON file format seem okay or would there be any other data worth storing in them?

viveklak: > Yeah I can try to do that.

To be clear, I am suggesting against that and retaining the `results.json` instead to avoid dirtying the worktree.

>  Does the current actual JSON file format seem okay or would there be any other data worth storing in them?
Seems reasonable to start with. There are a few additional checks before we checkin the file:
1. Make sure the test output is stable across multiple runs at the same version. We don't want to have unstable output if the file is to be checked in.
2. I recommend we pin the version of `azure-quickstart-templates` here: https://github.com/pulumi/pulumi-azure-native/blob/master/Makefile#L54 (i.e. add a `git checkout <current revision>`) in the target.

The goal is to make sure the output is stable. That way if we detect a drift in coverage metrics, we can at least triangulate what we changed in our code without having to second guess if the arm templates changed etc.

provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
```diff
@@ -162,6 +172,23 @@ CREATE TABLE errors(
...
+	require.NoError(t, err)
+
+
```
viveklak, line 191:
```suggestion
```

provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
hiteshboinpally: > To be clear, I am suggesting against that and retaining the `results.json` instead to avoid dirtying the worktree.

Ah, I missed the second comment, sounds good.

> 1. Make sure the test output is stable across multiple runs at the same version. We don't want to have unstable output if the file is to be checked in.

Yeah I see what you're saying it. It does seem that the output is not stable right now running it locally, which is interesting. I'm not sure exactly what is causing this, maybe has to do with how errors are being logged or something? Could be related to pinning a version but I didn't think that would change between local runs that quickly.

> 2. I recommend we pin the version of `azure-quickstart-templates` here: https://github.com/pulumi/pulumi-azure-native/blob/master/Makefile#L54 (i.e. add a `git checkout <current revision>`) in the target.
> 
> The goal is to make sure the output is stable. That way if we detect a drift in coverage metrics, we can at least triangulate what we changed in our code without having to second guess if the arm templates changed etc.

Right, gotcha makes sense.

viveklak (approved):

Issue #712, opened 2021-04-02
Stop generating the latest version


mikhailshilkov:
The `latest` modules are deprecated. We should stop generating them before the GA to reduce the size and avoid confusion. Any migration from old versions would have to migrate to 0.7.x/0.8.x before migrating to later versions.

PR #724, opened 2021-04-07
Remove the latest version of all resources

mikhailshilkov:
Resolve https://github.com/pulumi/pulumi-azure-native/issues/712


mikhailshilkov:
There is a changelog entry but GitHub fails to detect it, so I added a label.

viveklak (approved):

PR #740, opened 2021-04-12
Hitesh/arm2pulumi cvg

hiteshboinpally:
- Making test run without actually "failing". Errors are still logged, but in a slightly different format:
- Existing msg:
     ```
     === RUN   TestQuickstartTemplateCoverage/../testdata/azure-quickstart-templates/101-aks-advanced-networking/azuredeploy.json
    quickstart_coverage_test.go:75: 
        	Error Trace:	quickstart_coverage_test.go:75
        	Error:      	Received unexpected error:
        	            	failed to evaluate template expression for managedClusterResource: missing variable: vnetSubnetIDVar
        	Test:       	TestQuickstartTemplateCoverage/../testdata/azure-quickstart-templates/101-aks-advanced-networking/azuredeploy.json
     ```
- New msg:
     ```
     === RUN   TestQuickstartTemplateCoverage/../testdata/azure-quickstart-templates/101-aks-advanced-networking/azuredeploy.json
    quickstart_coverage_test.go:92: failed to evaluate template expression for managedClusterResource: missing variable: vnetSubnetIDVar
     ```
- Could reformat if desired. Overall results are the same as they were before.
- Of note, should be fairly straightforward to change to a threshold based failure now as well, or failing if the coverage actually gets worse between days.

### DIFF
--- a/provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
+++ b/provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
@@ -12,7 +12,6 @@ import (
 	"github.com/olekukonko/tablewriter"
 	"github.com/pulumi/pulumi-azure-native/provider/pkg/arm2pulumi"
 	"github.com/stretchr/testify/require"
-	// "github.com/stretchr/testify/assert"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -88,10 +87,7 @@ func TestQuickstartTemplateCoverage(t *testing.T) {
 					require.NoError(t, err)
 					require.NoError(t, ioutil.WriteFile(filepath.Join(*testOutputDir, template+".ts"), []byte(rendered["nodejs"]), 0600))
 				}
-			} else {
-				t.Log(err) // Silently log the error if this conversion failed
 			}
-
 		})
 	}
 	summarize(t, diagList)
--- a/.github/workflows/arm2pulumi-coverage-report.yml
+++ b/.github/workflows/arm2pulumi-coverage-report.yml
@@ -1,12 +1,8 @@
 name: arm2pulumi-coverage-report
-# "on":
-#   schedule:
-#     - cron: 35 17 * * *
-#   workflow_dispatch: {}
 "on":
-  pull_request:
-    branches:
-      - master
+  schedule:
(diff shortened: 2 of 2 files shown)


provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/olekukonko/tablewriter"
 	"github.com/pulumi/pulumi-azure-native/provider/pkg/arm2pulumi"
 	"github.com/stretchr/testify/require"
+	// "github.com/stretchr/testify/assert"
```
hiteshboinpally, line 15: Need to delete this line

provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go
```diff
@@ -71,23 +72,26 @@ func TestQuickstartTemplateCoverage(t *testing.T) {
...
+						diagList = append(diagList, Diag{Template: template, Diagnostic: d})
+					}
+				}
+				if *retainConverted {
+					rendered, _, err := renderer.RenderPrograms(body, []string{"nodejs"})
+					require.NoError(t, err)
+					require.NoError(t, ioutil.WriteFile(filepath.Join(*testOutputDir, template+".ts"), []byte(rendered["nodejs"]), 0600))
+				}
+			} else {
+				t.Log(err) // Silently log the error if this conversion failed
```
viveklak, line 92: I think we might want to record this as a fatal error instead of swallowing the error? Just create a new diagnostic with fatal and record something relevant to track the affected template.

hiteshboinpally: I thought lines 66-74 was still doing that, but I could misunderstand? If we record this as a fatal error do you mean doing something like `t.Error(err)` -- that seems to cause the overall test to fail. Or are you referring to something like `diagList` as an internal way to store the data?

viveklak: Ah so it is! You can just get rid of the else block IMO. We have what we want.

viveklak (approved):

PR #743, opened 2021-04-13
Registry task upgrade

### LINKED ISSUE #736: containerregistry:Task resource fails to create with InvalidRequestBody error
## Expected behavior

The resources would succeed in creating or, at least, prompt an error about the invalid present configuration if any.

## Current behavior

The resource fails to create and a cryptic message re. not having a body or missing a discriminator is presented. Analyzing the resource properties there is no evident discriminator key so one might assume there is some internal problem inside the provider that is triggering a request with an empty body?

> azure-native:containerregistry:Task SolutionBuildTask **creating failed** error: Code="InvalidRequestBody" Message="The request body is required for this request. Check if the body is not empty. If the request expects a polymorphic type, check if it has the discriminator property and is one of the described values." Target="request"

## Steps to reproduce

1. Create the following resource
    ```typescript
    this.buildTask = new Task("BuildTask", {
            registryName: this.registry.name,
            resourceGroupName: args.resourceGroup.name,
            agentConfiguration: {
                cpu: 2
            },
            status: "Enabled",
            taskName: "buildTask",
            identity: {
                type: "SystemAssigned"
            },
            platform: {
                architecture: "amd64",
                os: "Linux"
            },
            step: {
                contextPath: "/dev/null"
            },
            credentials: {
                customRegistries: {
                    [`${args.registryA}.azurecr.io`]: { identity: "system" },
                    [`${args.registryB}.azurecr.io`]: { identity: "system" }
                }
            },
            timeout: 3600,
            trigger: {
            }
        }, {parent: this})
    ```
    > NOTE: Both RegistryA and RegistryB variables are of type `string`

## Affected feature

`azure-native:containterregistry/Task`

mikhailshilkov:
Use the `2019-06-01-preview` version for the `containerregistry.Task` resource, fix #736

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ CHANGELOG
   `sql.ExtendedDatabaseBlobAuditingPolicy`, `sql.ExtendedServerBlobAuditingPolicy`,
   `sql.LongTermRetentionPolicy`, `sql.ServerBlobAuditingPolicy`, `sql.ServerSecurityAlertPolicy`.
   [#725](https://github.com/pulumi/pulumi-azure-native/issues/725)
+- Use the `2019-06-01-preview` version for the `containerregistry.Task` resource
+  [#736](https://github.com/pulumi/pulumi-azure-native/issues/736)
 
 ---
 


sdk/python/pulumi_azure_native/containerregistry/get_task.py
```diff
@@ -21,10 +21,13 @@ class GetTaskResult:
     The task that has the ARM resource and task properties. 
     The task will have all information to schedule a run against it.
     """
-    def __init__(__self__, agent_configuration=None, creation_date=None, credentials=None, id=None, identity=None, location=None, name=None, platform=None, provisioning_state=None, status=None, step=None, tags=None, timeout=None, trigger=None, type=None):
+    def __init__(__self__, agent_configuration=None, agent_pool_name=None, creation_date=None, credentials=None, id=None, identity=None, is_system_task=None, location=None, log_template=None, name=None, platform=None, provisioning_state=None, status=None, step=None, system_data=None, tags=None, timeout=None, trigger=None, type=None):
```
lblackstone, line 24: Not specifically related to this PR, but this signature change could break positional arg usage. I'm not sure if it's worth appending new parameters rather than alphabetizing to avoid this. (Sacrifices readability in favor of compatibility)

lblackstone (approved):

sdk/python/pulumi_azure_native/containerregistry/get_task.py
mikhailshilkov: Oh, interesting. Do you know how we handle this in TF-based providers?

lblackstone: I'm not sure. My guess is that we're clobbering the existing signature in the same way. Most users are likely to use kwargs given the long list of options on most resources, but this could lead to subtle breaks if not.

mikhailshilkov: Yeah, it looks like we also reorder them somewhat arbitrarily: https://github.com/pulumi/pulumi-gcp/pull/558/files#diff-8bea7c3056bab4675291789291ba2a00711e860f76383317e117686401f61abb

image

komalali: It doesn't look arbitrary to me. If something becomes `Optional` then it has a default value passed in, and so it has to go after the args that don't have default values.

At least, the gcp example doesn't look arbitrary to me. It looks like the example commented on here is sorted by alphabetical order.

justinvp: A couple things worth noting here:

1. `GetTaskResult` is the result of an invoke, which we instantiate internally. It's likely very rare that anyone would be calling its `__init__`.
2. For `TArgs` classes, I've deliberately added a `*` parameter, which forces users to use kwargs to avoid this issue with positional arguments.

komalali: But yeah, I don't know how common it is for folks to use positional args. And in fact, with the new `ResourceArgs` changes in the python SDK I'm not sure it's even possible to use positional args now.

mikhailshilkov: > It doesn't look arbitrary to me

Yes, probably my bad wording. I meant that any argument may get a default value in a minor release and end up changing the position.

But it looks like you have it under control 👍

Issue #858, opened 2021-05-19
Missing C# classes for complete definition of dashboards ("Microsoft.Portal/dashboards" -> "azure-native:portal:Dashboard")


rekv-mews:
The C# source code generated via `pulumi import azure-native:portal:Dashboard` does not contain the definition of metrics used in a particular dashboard.

## Expected behavior
I expected that the C# source code generated via `pulumi import azure-native:portal:Dashboard` would follow the complete dashboard definition contained in `json` ARM template given by `az portal dashboard show --name <... dashboard name ...> --resource-group <... resource group name ...>`.

## Current behavior
1. The C# source code generated via `pulumi import azure-native:portal:Dashboard` does not contain the definition of metrics used in a particular dashboard. 
2. An attempt to set particular metrics according to path (`"lenses" ` -> `"parts"` -> `"<.. chart index ...>"` -> `"settings"` -> `"content"`  -> `"options"` -> `"chart"` -> `"metrics"`) found in `json` ARM template failed. The similar C# class definition path ends with `class MarkdownPartMetadataSettingsArgs` where property `Content` of  `class MarkdownPartMetadataContentArgs` containing only one property `Settings` which is of type `class MarkdownPartMetadataSettingsArgs`. This cyclic definition causes inability to make set up of next properties.

![arm](https://user-images.githubusercontent.com/75578608/118814123-1752f680-b8b0-11eb-9374-b5aaf6a53be3.png)

![cyclic-definition](https://user-images.githubusercontent.com/75578608/118813117-1c637600-b8af-11eb-8336-8a6589020ccc.png)

## Steps to reproduce
1. Do import of dashboard via `pulumi import azure-native:portal:Dashboard`
2. Do export of `json` ARM template via  `az portal dashboard show --name <... dashboard name ...> --resource-group <... resource group name ...>`.
3. Compare content `json` ARM template and C# code.
4. Look at cyclic definition which starts on the path `class Pulumi.AzureNative.Portal.Inputs.DashboardPartsArgs` -> `Pulumi.AzureNative.Portal.Inputs.MarkdownPartMetadataArgs` -> propery `Settings`. There is no way how to continue.

## Context (Environment)
Inability to define dashboards using C# code.

## Affected feature

mikhailshilkov:
[...]

PR #925, opened 2021-06-16
Add missing types for complete definition of dashboards

mikhailshilkov:
Fix https://github.com/pulumi/pulumi-azure-native/issues/858

Other approaches that I tried were all breaking, while this one fixes specifically this one spot.


lblackstone (approved):
That resulting name is a bit of a mouthful, but this seems like a reasonable workaround for a rare edge case.

viveklak (approved):

PR #974, opened 2021-07-07
Update pulumi/pulumi

komalali:


mikhailshilkov:
These changes look massively breaking across all languages (definitely .NET and Node, a bit harder to understand the changes in Python and Go). I understand there's nothing wrong with this PR per se. What would be a good next step to triage those?

mikhailshilkov:
Actually, Node looks fine, I wasn't looking carefully enough. .NET is tracked in https://github.com/pulumi/pulumi/issues/7448

mikhailshilkov:
I'm trying again in https://github.com/pulumi/pulumi-azure-native/pull/986

PR #975, opened 2021-07-07
Remove note about beta from README

lukehoban:


ericrudder:
Thanks!

mikhailshilkov:
It looks like this broke the build - we need to copy to Python README (or build, which does that automatically)

lukehoban:
Fix in https://github.com/pulumi/pulumi-azure-native/pull/977.

Issue #771, opened 2021-04-20
Error changing principalId of RoleAssignment


WillooWisp:
Updating this resource, e.g. changing principalId, throws the error below, but instead the resource should be marked for replacement right?

~ azure-native:authorization:RoleAssignment: (update)
      ~ principalId: "3763b23f-33e4-4db1-b8d9-366bf8dbdff0" => "2acecc98-54a8-4791-8db2-696a69c35460"

error: autorest/azure: Service returned an error. Status=400 Code="RoleAssignmentUpdateNotPermitted" Message="Tenant ID, application ID, principal ID, and scope are not allowed to be updated."
error: update failed

mikhailshilkov:
Yes, this change should probably cause a replacement instead of the in-place update.

Microsoft has annotations to mark a property as non-updatable but they barely use it. We have an additional hard-coded map [here](https://github.com/pulumi/pulumi-azure-native/blob/master/provider/pkg/gen/replacement.go#L9) to add extra annotations. For anyone looking to contribute, adding properties to this map would be a great "first issue" to open a PR for.

morrell:
This should definitely trigger a replace, can we get this bumped, it is a PITA to work around.

mikhailshilkov:
Fixed by https://github.com/pulumi/pulumi-azure-native/pull/1010

PR #1007, opened 2021-07-20
Adds authorization:RoleAssignment replacements

jeffmorrell:
Adds authorization:RoleAssignment overrides for forcing replace
Fix https://github.com/pulumi/pulumi-azure-native/issues/771

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
+- Updates Authorization::RoleAssignment to force replace on principalId or scope change (fixes [771](https://github.com/pulumi/pulumi-azure-native/issues/771))
 
 #### Dependencies
 


provider/pkg/gen/replacement.go
```diff
@@ -7,6 +7,9 @@ import "github.com/pulumi/pulumi/pkg/v3/codegen"
 // forceNewMap is a map of Azure Resource Provider -> Resource Name -> Properties that cause replacements.
 // API Versions are currently ignored.
 var forceNewMap = map[string]map[string]codegen.StringSet{
+	"Authorization": {
+		"RoleAssignment": codegen.NewStringSet("applicationId", "principalId", "scope"),
```
mikhailshilkov, line 11: According to [docs](https://www.pulumi.com/docs/reference/pkg/azure-native/authorization/roleassignment/) this resource has no property called `applicationId`

mikhailshilkov (commented):
Thank you for the PR! Could you please add a short description to `CHANGELOG.md`?

provider/pkg/gen/replacement.go
morrell: Interesting, was going off the error message I receive. I'll remove the applicationId as the other two are the only valid ones reported in the error message.

mikhailshilkov:
Alright, the change LGTM but we need to run `make build` to propagate the changes to the metadata file. Would you be able to do so or should I?

morrell:
> Alright, the change LGTM but we need to run `make build` to propagate the changes to the metadata file. Would you be able to do so or should I?

@mikhailshilkov if you wouldn't mind doing it I would very much appreciate it.

mikhailshilkov:
I'm not allowed to push to your fork so I had to open https://github.com/pulumi/pulumi-azure-native/pull/1010

PR #1033, opened 2021-07-30
Upgrade codegen to pulumi master

t0yv0:
Sorry folks I'm out of the loop on the precise process, I just noticed that pulumi/pulumi master verifications were red on trying to build this repo against the tip of our codegen. The changes are very mechanical so I'm including them here. But care is needed to see if the generated code changes are acceptable.

If there's other work fixing this, disregard.


komalali:
I don't think we can move forward with this until https://github.com/pulumi/pulumi/issues/7595 is fixed

Issue #1043, opened 2021-08-03
Examples/simple fails during `pulumi preview` due to missing resource group


T-Vova:
## Steps to reproduce
Running `pulumi preview` or `pulumi up` in the simple example leads to an unhanded exception as the resource group "Azure-Account-Cleanup" doesn't exist:

```
Diagnostics:
  pulumi:pulumi:Stack (simple-dev):
    error: Running program '/home/vova/pulumi-azure-native/examples/simple' failed with an unhandled exception:
    Error: invocation of azure-native:resources:getResourceGroup returned an error: request failed /subscriptions/e42a364a-2eaa-4d7d-9cd9-51d8b2ac497c/resourcegroups/Azure-Account-Cleanup: autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'Azure-Account-Cleanup' could not be found."
        at Object.callback (/home/vova/pulumi-azure-native/sdk/nodejs/node_modules/@pulumi/pulumi/runtime/invoke.js:140:33)
        at Object.onReceiveStatus (/home/vova/pulumi-azure-native/sdk/nodejs/node_modules/@grpc/grpc-js/src/client.ts:338:26)
        at Object.onReceiveStatus (/home/vova/pulumi-azure-native/sdk/nodejs/node_modules/@grpc/grpc-js/src/client-interceptors.ts:426:34)
        at Object.onReceiveStatus (/home/vova/pulumi-azure-native/sdk/nodejs/node_modules/@grpc/grpc-js/src/client-interceptors.ts:389:48)
        at /home/vova/pulumi-azure-native/sdk/nodejs/node_modules/@grpc/grpc-js/src/call-stream.ts:276:24
        at processTicksAndRejections (node:internal/process/task_queues:78:11)
...
...
...
```

mikhailshilkov:
The resource group does exist in our CI account, I suspect you are logged into a different subscription.

I realize it's not universally useful but I run this example as a smoke test for dev changes, and this particular call makes sure that invokes (functions) work.

nesl247:
I think the point is that examples should work for anyone without any changes, not require any other resources to exist.

viveklak:
> I think the point is that examples should work for anyone without any changes, not require any other resources to exist.

[...]

PR #1044, opened 2021-08-03
Addresses Issue #1043

T-Vova:


examples/simple/index.ts
```diff
@@ -181,7 +181,7 @@ new storage.EncryptionScope("encscope", {
     source: storage.EncryptionScopeSource.Microsoft_Storage,
 });
 
-export const existingRg = resources.getResourceGroup({ resourceGroupName: "Azure-Account-Cleanup" });
```
mikhailshilkov, line 184: I want to keep a function call here because I use it for local dev validation. We could maybe run it inside an `apply` and retrieve the properties of the resource group that is created above. I guess this would achieve both goals.

mikhailshilkov (changes requested):

PR #1045, opened 2021-08-04
Add a link to the docs of SP auth

mikhailshilkov:
Otherwise, we give a confusing error message:

```
error: building auth config: Authenticating using the Azure CLI is only supported as a User (not a Service Principal).
    
    To authenticate to Azure using a Service Principal, you can use the separate 'Authenticate using a Service Principal'
    auth method - instructions for which can be found here:
    
    Alternatively you can authenticate using the Azure CLI by using a User Account.
```

Note the empty space after "can be found here: ".


mikhailshilkov:
@viveklak A friendly nudge on this one ;)

viveklak:
Sorry for the delay, this got burried.

PR #1125, opened 2021-09-07
Update pulumi/pulumi and cut Go comments

mikhailshilkov:
Update to pulumi/pulumi 3.10 and work around the Go SDK size by cutting out all comments (descriptions) in versioned (`vYYMMDD` as opposed to top-level) modules (only in Go). I know it's a lame fix but at least it lets us get to a buildable Go SDK again and use a fresher pu/pu version. The resulting size is 506MB which is still dangerously close, so we need to keep working on a better fix.

Fix https://github.com/pulumi/pulumi-azure-native/issues/1101


provider/cmd/pulumi-gen-azure-native/main.go
```diff
@@ -96,6 +103,71 @@ func main() {
 	}
 }
 
+func removeAllComments(pkgSpec schema.PackageSpec) *schema.PackageSpec {
```
viveklak, line 106: Nit: rename to `removeCommentsFromVersionedPkgs`? The function makes it out to be for all descriptions (versioned or not).

Alternatively should we do this in the example generation phase - i.e. skip example generation for versioned resources? I imagine examples are the largest culprits here?

mikhailshilkov: Indeed, I forgot to rename the function after I narrowed the scope, thank you, will do.

I don't think we put examples into the SDK code, do we?

viveklak: > I don't think we put examples into the SDK code, do we?
Ah forgot about that, yes we didn't.

mikhailshilkov:
Any further comments @viveklak? (I'll rename the function when I rebase)

viveklak:
> Any further comments @viveklak? (I'll rename the function when I rebase)

LGTM

viveklak (approved):

PR #1145, opened 2021-09-15
fix nil pointer deref when using managed identity (fixes #1008)

### LINKED ISSUE #1008: authorization.GetClientConfig panics when using managed identity
Calling `authorization.GetClientConfig` while using MSI causes a nil panic within the provider plugin, which we're encountering when running pulumi within a CI pod. 

The culprit is this line: https://github.com/pulumi/pulumi-azure-native/blob/v1.18.0/provider/pkg/provider/provider.go#L179

It turns out that `GetAuthenticatedObjectID` is a [function property in the `auth` struct](https://github.com/hashicorp/go-azure-helpers/blob/v0.12.0/authentication/config.go#L23) rather than a method, and [it doesn't get set when using the MSI authentication method](https://github.com/hashicorp/go-azure-helpers/blob/v0.12.0/authentication/auth_method_msi.go#L71), so calling it unconditionally panics. 
Perhaps the provider should check that the property is non-nil before calling it?

Cheers

polivbr:
Fix https://github.com/pulumi/pulumi-azure-native/issues/1008

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ CHANGELOG
 - The shape of the `documentdb.Service` resource adjusted to match API expectations
   ([source](https://github.com/Azure/azure-rest-api-specs/pull/15971))
 
+#### Bug Fixes
+
+- Fix `authorization.getClientConfig` when using Managed Identities
+  [#1008](https://github.com/pulumi/pulumi-azure-native/issues/1008).
+
 ---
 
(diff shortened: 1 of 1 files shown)


mikhailshilkov (approved):
LGTM. Thanks a ton for the fix! 🎉

polivbr:
It's not clear to me why the automerge job failed nor how to fix it.

Issue #1176, opened 2021-09-27
Azure SQL missing resource encryptionProtector


bgelens:
Right now the encryptionProtector resource is missing (AFAICT) which makes it difficult to implement Customer Managed Key encryption on the SQL Server level.

Azure classic does have the resource https://www.pulumi.com/docs/reference/pkg/azure/mssql/servertransparentdataencryption/

mikhailshilkov:
It's now available in version 1.33.0, thank you for the heads-up

bgelens:
Thank you for this very quick resolve!

PR #1177, opened 2021-09-27
SQL encryption protector resource

mikhailshilkov:
Resolve https://github.com/pulumi/pulumi-azure-native/issues/1176


mikhailshilkov:
> assuming that we haven't blown past any size limits for Go.

We haven't yet but we are so close :(

PR #1188, opened 2021-10-01
Update to pulumi/pulumi 3.13.2

mikhailshilkov:
We are at 513MB of Go SDK so we need to iterate fast on slimming it down at least a bit further


sdk/go/azure/containerregistry/pulumiTypes.go
```diff
@@ -12976,6 +13027,75 @@ func (o SourceControlAuthInfoResponseOutput) TokenType() pulumi.StringPtrOutput
 	return o.ApplyT(func(v SourceControlAuthInfoResponse) *string { return v.TokenType }).(pulumi.StringPtrOutput)
 }
 
+type SourceControlAuthInfoResponsePtrOutput struct{ *pulumi.OutputState }
```
t0yv0, line 13030: Do we have any ideas around excluding these helper types? Namespace-based patterns in config? We could detect which are not in use in the rest of the code, and use that plus config patterns to decide to omit them.

mikhailshilkov: There's just a handful of these but yes - our next step is likely exclude some types to get us to the yellow zone again

lblackstone (approved):

t0yv0 (approved):

PR #1196, opened 2021-10-05
Increase RAM accessible to TSC

t0yv0:
It's starting to use about ~5 GB. Total RAM on GHA action worker is 14 GB.

Also, by using `yarn run` we get the version of tsc that's pinned in development, developed against. I thought it is better, but if the original behavior is needed for some reason I can revert.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ build_nodejs:: VERSION := $(shell pulumictl get version --language javascript)
 build_nodejs::
 	cd ${PACKDIR}/nodejs/ && \
 	yarn install && \
-	NODE_OPTIONS='--max-old-space-size=13312' yarn run tsc --diagnostics && \
+	NODE_OPTIONS=--max-old-space-size=8192 yarn run tsc --diagnostics && \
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
We do this in azure-native: https://github.com/pulumi/pulumi-azure-native/blob/e5de8462db6aef2373554cc3a55eb4bfa4993286/Makefile#L82

Makefile
```diff
@@ -79,7 +79,7 @@ build_nodejs:: VERSION := $(shell pulumictl get version --language javascript)
 build_nodejs::
 	cd ${PACKDIR}/nodejs/ && \
 	yarn install && \
-	node --max-old-space-size=4096 /usr/local/bin/tsc --diagnostics && \
+	NODE_OPTIONS=--max-old-space-size=8192 yarn run tsc --diagnostics && \
```
mikhailshilkov, line 82: We've originally created this line to avoid OOM in GHA. Has the size increased since then?

t0yv0: https://github.com/pulumi/pulumi/pull/8047 OOMs on downstream check. I seem to need this to push through. Locally #8047 peaks at about 5.5 GB RAM

Issue #1194, opened 2021-10-05
Go SDK is above the max size again


mikhailshilkov:
We need to reduce it again to be able to cut functional releases

PR #1197, opened 2021-10-06
Remove Go comments

mikhailshilkov:
Temporary and dirty fix https://github.com/pulumi/pulumi-azure-native/issues/1194


mikhailshilkov:
Self-merged to get the release out of the door

PR #1269, opened 2021-11-04
Add standard devcontainer

danielrbradley:
Copied from `pulumi/pulumi` - makes it easier to use project in Codespaces and GitPod.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,17 @@
+image:
+  file: .gitpod/Dockerfile
+tasks:
+  - name: Start Docker Daemon
+    command: sudo docker-up &
+  - command: "echo 'export PIP_USER=no' >> ~/.bashrc && export PIP_USER=no"
+    name: Use Pulumi
+ports:
+  - port: 3000
+    onOpen: ignore
+vscode:
+  extensions:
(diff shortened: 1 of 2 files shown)


Frassle:
Wonder if there's an easy way to keep these in sync, as I've pretty much just made a change to the pulumi/pulumi one to fix a build on the latest base image (https://github.com/pulumi/pulumi/commit/f8a98b2a86472773051507a829b18b653fe18673)

danielrbradley:
Not sure on the alternative approaches without knowing how much different projects can vary in their required setup. I guess the ideal would be to have a single image pre-built for each of Codespaces and another for Gitpod as they don't seem to be compatible, then each project points to the latest image. Would it be worth getting this in as-is if it's functional, with an aim to start a wider conversation around standard virtualized development environments?

Frassle (approved):

sanmai-NL:
> Wonder if there's an easy way to keep these in sync, as I've pretty much just made a change to the pulumi/pulumi one to fix a build on the latest base image ([pulumi/pulumi@f8a98b2](https://github.com/pulumi/pulumi/commit/f8a98b2a86472773051507a829b18b653fe18673))

You can use Git submodules and symlink or use as template dev container configurations accross repositories.

PR #1281, opened 2021-11-08
Add go mod tidy

t0yv0:
The changes to Makefile run `go mod tidy` prior to building Go. The idea here is to unblock downstream codegen check failure from pulumi/pulumi. Example:

https://github.com/pulumi/pulumi/runs/4143596111?check_suite_focus=true#step:8:72954

The changes to `go.mod` are from me running `go mod tidy` locally on a Mac:

```
go version go1.17.2 darwin/amd64
```

Would this be reasonable to add? It seems that `go mod tidy` automatically modifies `go.mod` upon inspecting source code; so by adding this to CI we might be hiding issues requiring manual go.mod modifications and perhaps more importantly we may be bypassing human audit on go.mod changes. If this is a concern, let me know and I can try to pursue an approach of modifying `pulumi/action-test-provider-downstream@releases/v7` to suit the need of downstream verification specifically, and not the general Makefile (which can participate in dev&release workflow).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ generate_go::
 
 build_go::
 	# Only building the top level packages and building 1 package at a time to avoid OOMing
-	cd sdk/ && \
+	cd sdk/ && go mod tidy && \
 	GOGC=50 go list github.com/pulumi/pulumi-azure-native/sdk/go/azure/... | grep -v "latest\|\/v.*"$ | xargs -L 1 go build
 
 clean::
--- a/sdk/go.mod
+++ b/sdk/go.mod
@@ -4,10 +4,56 @@ go 1.17
 
 require (
 	github.com/blang/semver v3.5.1+incompatible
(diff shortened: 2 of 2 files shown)


t0yv0:
Hm, https://github.com/pulumi/action-test-provider-downstream/blob/master/src/main.ts#L117 invokes `make only_build` after doing some `go mod` manipulations. I am starting to lean myself to that being the right place to make changes.

t0yv0:
This fails because CI toolchain uses Go 1.16

https://github.com/pulumi/pulumi-azure-native/pull/1282

lblackstone (approved):
I'm fine with this change, but I normally expect the `ensure` target to run `go mod tidy`. If that would solve the same use case, maybe we could move it there instead.

t0yv0:
The underlying problem disappeared after updating the build to go.17 so I will close for now as this change is not without a confusing cost.

Issue #969, opened 2021-07-05
Deletion of DNS zone fails


lkt82:
When deleting a Azure DNS zone, the deletion fails with this error

`error: Error occurred unmarshalling JSON - Error = 'json: cannot unmarshal bool into Go value of type map[string]interface {}' JSON = 'true'`

## Steps to reproduce

1. Create a "Pulumi.AzureNative.Network.Zone" resource
2. Create a few "Pulumi.AzureNative.Network.RecordSet" resources for the zone
2. Destroy the stack

Expected: The zone to be deleted
Actual: Deletion fails

**Note**
If you run a second destroy after the failed destroy it succeeds.

**Context**
Language: C#
Provider version: 1.15.0
Pulumi version: 3.6.0

mikhailshilkov:
Hi @lkt82 thank you for reporting this.

Do you have a simple snippet for me to reproduce this problem?

lkt82:
```
using System.Threading.Tasks;
using Pulumi;
using Pulumi.AzureNative.Network;
using Pulumi.AzureNative.Network.Inputs;
using Pulumi.AzureNative.Resources;
using Deployment = Pulumi.Deployment;

namespace TrafficPlatform
{
    public class TrafficPlatformStack : Stack
    {
        public static async Task Main() => await Deployment.RunAsync<TrafficPlatformStack>();

        private Zone _zone = null!;
        private ResourceGroup _resourceGroup;

        public TrafficPlatformStack()
        {
            _resourceGroup = new ResourceGroup("rg", new ResourceGroupArgs
            {
            });

            CreateZone();

            CreateRecords();
        }

        private void CreateZone()
        {
            _zone = new Zone("dnsz", new ZoneArgs
            {
                ResourceGroupName = _resourceGroup.Name,
                Location = "global",
                ZoneType = ZoneType.Public,
                ZoneName = "pulumitestcase.com"
            }, new()
            {
                Parent = this,
                DeleteBeforeReplace = true
            });
        }

        private void CreateRecords()
        {
            Cname("test1", "test1", 1);
            Cname("test2", "test2", 1);
            Cname("test3", "test3", 1);
        }

        private RecordSet Cname(string name, string alias, double ttl)
        {
            var api = new RecordSet(name, new RecordSetArgs
            {
                ResourceGroupName = _resourceGroup.Name,
                ZoneName = _zone.Name,
                RelativeRecordSetName = name,
                Ttl = ttl,
[...]

PR #1291, opened 2021-11-12
Ignore delete response body

mikhailshilkov:
At the moment, we deserialize the body of DELETE operation responses to JSON. We do that similarly to PUT operations, although, we don't use the result of DELETE response. If the body is misformed (not a JSON object) this will cause a runtime error even though the status code says the deletion was successful.

Fix https://github.com/pulumi/pulumi-azure-native/issues/969

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,17 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+#### Bug fixes:
+
+- Make Deletion of DNS zone more robust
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
Makes sense - the body will still be parsed if there's an error (within the `WithErrorUnlessStatusCode` implementation), but not if the response was okay.

lblackstone (approved):

viveklak (approved):

PR #1334, opened 2021-12-01
Allow star modifier in default state

### LINKED ISSUE #866: Adding a WebAppSwiftVirtualNetworkConnectionSlot fails
I use the WebAppSwiftVirtualNetworkConnection for non web app slots and it works fine, but when I try to use the WebAppSwiftVirtualNetworkConnectionSlot version to create the resource for a web app slot it fails with cannot create already existing resource, but it does not exist and is possible to add through the Azure portal.

Configuration
```
var virtualNetworkConnectionSlot = new WebAppSwiftVirtualNetworkConnectionSlot($"wasvncs-{appServiceName}-{slotConfig.Name}", new WebAppSwiftVirtualNetworkConnectionSlotArgs
{
    Name = webApp.Name,
    ResourceGroupName = m_ResourceGroupName,
    Slot = webAppSlot.Name.Apply(name => name.Split('/').Last()),
    SwiftSupported = true,
    SubnetResourceId = args.VnetSubnetId
}, new CustomResourceOptions());
```

Output from pulumi up
```
+ azure-native:web:WebAppSwiftVirtualNetworkConnectionSlot: (create)
        [urn=urn:pulumi:vnetapptest123::platform::azure-native:web:WebAppSwiftVirtualNetworkConnectionSlot::wasvncs-VnetAppTest123-exp]
        [provider=urn:pulumi:vnetapptest123::platform::pulumi:providers:azure-native::default_1_7_0::2492c4a1-cb5e-4bec-8667-721c2c37d59c]
        name             : "VnetAppTest123"
        resourceGroupName: "VnetAppTest123"
        slot             : "exp"
        subnetResourceId : "/subscriptions/314abb85-78af-4b74-bb84-6e3793add2c9/resourceGroups/VnetAppTest123-Services/providers/Microsoft.Network/virtualNetworks/VnetAppTest123-Services/subnets/appservice"
        swiftSupported   : true
error: cannot create already existing subresource '/subscriptions/314abb85-78af-4b74-bb84-6e3793add2c9/resourceGroups/VnetAppTest123/providers/Microsoft.Web/sites/VnetAppTest123/slots/exp/networkConfig/virtualNetwork'
```

mikhailshilkov:
Fix #866

That resource behaves somewhat uniquely because the API returns a value for `subnetResourceId` property which is a resource ID and can't be preset in our default state map.

I went ahead and added a `*` modified to note to accept any value in the default state. I realize we get closer to having a separate non-comprehensible language here, so I'm happy to consider other ideas.

### DIFF
--- a/provider/pkg/resources/convert.go
+++ b/provider/pkg/resources/convert.go
@@ -261,6 +261,8 @@ func (k *SdkShapeConverter) IsDefaultResponse(putParameters []AzureAPIParameter,
 						return false
 					}
 				default:
+					// `*` default body means that we want to accept any value there.
+					// It's used for values that are determined dynamically by Azure API.
 					if defaultBody[key] != value && defaultBody[key] != "*" {
 						return false
 					}
--- a/provider/pkg/openapi/defaultResourcesState.go
+++ b/provider/pkg/openapi/defaultResourcesState.go
@@ -111,7 +111,7 @@ var defaultResourcesStateRaw = map[string]map[string]interface{}{
 		"swiftSupported": true,
 	},
 	"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}/slots/{slot}/networkConfig/virtualNetwork": {
-		"subnetResourceId": "*",
+		"subnetResourceId": "*", // This is going to be a resource ID, so we choose accept any value here.
 		"swiftSupported":   true,
 	},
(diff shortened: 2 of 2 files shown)


lblackstone (approved):
Seems reasonable to me

provider/pkg/resources/convert.go
```diff
@@ -261,7 +261,7 @@ func (k *SdkShapeConverter) IsDefaultResponse(putParameters []AzureAPIParameter,
 						return false
 					}
 				default:
-					if defaultBody[key] != value {
+					if defaultBody[key] != value && defaultBody[key] != "*" {
```
viveklak, line 264: Nit: add a small note for why "*" is used here?

viveklak (approved):

PR #1372, opened 2021-12-17
Update to pulumi 3.20.0

mikhailshilkov:


danielrbradley (approved):
What's the overall impact here in code size? Looks like it's's a significatnt amount of code removed - is this because of the removal of unreferenced input types?

mikhailshilkov:
Yes, the total Go SDK went from 507 MB to 311 MB because of unreferenced type removal.

PR #1374, opened 2021-12-17
Remove etag input properties

### LINKED ISSUE #1337: pulumi.Import() requires eTag to be present
## Hello!

## Issue details

When importing resources which were created using classic-provider to azure-native provider, running pulumi.Import() on certain resources would require the `eTag` field to be present. I haven't found a way to retrieve the `eTag` value nor was it present in our resource properties via app.pulumi.com.

**Workaround:** A temporary solution is to manually import individual resources using the pulumi cli by running `pulumi import azure-native:network:RecordSet testresource "azure_resource_id"`. On a successful import, the return value would return the value of the `eTag` and then we'll manually rerun pulumi without this `eTag` to remove it.

### Steps to reproduce

1. This is happening with several resources but let's use RecordSet as an example.
2. Deploy a RecordSet using the pulumi classic provider following the instructions [here](https://www.pulumi.com/registry/packages/azure/api-docs/dns/nsrecord/)
3. Import the resource deployed in step two via pulumi.Import(pulumi.ID("azure_resource_id")) where **azure_resource_id** is the unique id of the Azure resource in step 2. The pulumi native-provider resource can be found [here](https://www.pulumi.com/registry/packages/azure-native/api-docs/network/recordset/#create-ns-recordset)
4. Once all of the values are in place, pulumi preview will fail with missing eTag to be presented.

Expected:  We expected a successful import without needing to specify the `eTag` field
Actual:  pulumi.Import() fails until we specify the `eTag` field

mikhailshilkov:
ETags may be defined as inputs to PUT endpoints, but we should never model them as resource inputs because they are a protocol implementation detail rather than a meaningful desired-state property.

This PR removes `etag` properties from resource inputs (not the types - maybe there are valid cases there).

Fix #1337 - we improve the import experience with this change.

### DIFF
--- a/sdk/go/azure/aad/domainService.go
+++ b/sdk/go/azure/aad/domainService.go
@@ -134,8 +134,6 @@ type domainServiceArgs struct {
 	DomainSecuritySettings *DomainSecuritySettings `pulumi:"domainSecuritySettings"`
 	// The name of the domain service.
 	DomainServiceName *string `pulumi:"domainServiceName"`
-	// Resource etag
-	Etag *string `pulumi:"etag"`
 	// Enabled or Disabled flag to turn on Group-based filtered sync
 	FilteredSync *string `pulumi:"filteredSync"`
 	// Secure LDAP Settings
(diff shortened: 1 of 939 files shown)


danielrbradley (commented):
Is the change to the descriptions also expected or is this a different side-effect happening?

mikhailshilkov:
> Is the change to the descriptions also expected or is this a different side-effect happening?

Unfortunately, our example generation isn't deterministic which changes descriptions. We need to hunt those down but didn't prioritize that to date. I think it's https://github.com/pulumi/pulumi/issues/7713 but I'm not totally sure.

danielrbradley (approved):

PR #1382, opened 2021-12-20
Update Deploy with Pulumi button

leezen:


mikhailshilkov (approved):
It's probably an accident we have a button on this one particular example

PR #1414, opened 2022-01-07
Correct case for StorageAccount.VirtualNetworkRule.State

gregingenii:
Reference: https://docs.microsoft.com/en-us/azure/templates/microsoft.storage/storageaccounts?tabs=json#virtualnetworkrule

This was causing a mismatch on a Pulumi refresh. I think I've caught everywhere that needs changing but please advise


mikhailshilkov:
Hi @gregingenii thank you for your PR. 

The `schema.json` that you changed manually is actually automatically generated from Azure Open API specs, so we shouldn't be changing it by hand. I can see that Azure have already fixed the casing in the latest versions of Open API (2021-04-01+), so you can get those changes by using an explicit API version (e.g. `import * as storage from "@pulumi/azure-native/storage/v20210401";`. We will also rev our default API version to the latest when we release 2.0 of the provider later this year.

gregingenii:
@mikhailshilkov Thank you for the response. Sorry but I don't understand your comment: are you saying we shouldn't go ahead with this PR?

mikhailshilkov:
Yes, the fix has to happen in Open API specs - otherwise your code changes will be overridden/discarded on the next code generation run.

gregingenii:
Ah, I think I understand now: is it the 'defaultApiVersion' setting for Microsoft.Storage in pulumi-azure-native/azure-provider-versions/provider_list.json (L54553) that needs changing will resolve this?

If I've understood this correctly: is there a concern that the API version we're getting from Azure is not the one used to generate the package, causing this mismatch?

If you think it's best to close this PR, then that's fine by me.

mikhailshilkov:
> is it the 'defaultApiVersion' setting for Microsoft.Storage in pulumi-azure-native/azure-provider-versions/provider_list.json (L54553) that will be changed in v2.0 that will resolve this?

We actually control the default version ourselves [here](https://github.com/pulumi/pulumi-azure-native/blob/master/provider/pkg/openapi/versioner.go#L162). Our current policy is described in the [Top-Level Modules](https://www.pulumi.com/registry/packages/azure-native/version-guide/#top-level-modules).

> If I've understood this correctly: is there a concern that the API version we're getting from Azure is not the one used to generate the package, causing this mismatch?

The concern is that if we change the default version, we will get some code-level breaking changes in our SDKs, which may be disruptive for users.

> If you think it's best to close this PR, then that's fine by me.

Yes, I think I'll close the PR, as there's no simple way to achieve what you tried here. Once again, the workaround is to use an explicit API version in your code, which should work just fine for this case.

PR #1435, opened 2022-01-18
Avoid duplicate go module names

danielrbradley:
gopls complains at duplicate modules within the same workspace so make fake module names unique. This makes the project openable in VSCode and other editors with no custom settings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,8 @@ build_python::
 	python3 setup.py clean --all 2>/dev/null && \
 	rm -rf ./bin/ ../python.bin/ && cp -R . ../python.bin && mv ../python.bin ./bin && \
 	sed -i.bak -e 's/^VERSION = .*/VERSION = "$(PYPI_VERSION)"/g' -e 's/^PLUGIN_VERSION = .*/PLUGIN_VERSION = "$(VERSION)"/g' ./bin/setup.py && \
-	rm ./bin/setup.py.bak && rm ./bin/go.mod && \
+	rm ./bin/setup.py.bak && \
+	rm ./bin/go.mod && \
 	cd ./bin && python3 setup.py build sdist
 
 generate_dotnet::


Makefile
```diff
@@ -87,6 +87,7 @@ generate_nodejs::
 build_nodejs:: VERSION := $(shell pulumictl get version --language javascript)
 build_nodejs::
 	cd ${PACKDIR}/nodejs/ && \
+	echo "module fake_nodejs_module // Exclude this directory from Go tools\n\ngo 1.16" > go.mod && \
```
mikhailshilkov, line 90: I don't think I love doing this in the build target. I'd often do a `make local_generate` to re-gen SDKs and then commit the result. It will mean my commits wouldn't work after this change. Our CI may rely on a similar thing (we'll see when it builds).

danielrbradley: As in it's an issue if the file modified date has changed? The file content shouldn't change going forward as I've already committed the updated content in this PR.

Would you just prefer it in the clean step instead then?

mikhailshilkov: `make clean` deletes these files, then `make generate` doesn't recreate them - so I guess there will be a git diff. I guess my idea is that `build` should make no committable changes. Keeping it in the `clean` step should work fine, other options may be changing generate steps.

lblackstone (approved):

Makefile
```diff
@@ -102,7 +102,7 @@ build_python::
 	python3 setup.py clean --all 2>/dev/null && \
 	rm -rf ./bin/ ../python.bin/ && cp -R . ../python.bin && mv ../python.bin ./bin && \
 	sed -i.bak -e 's/^VERSION = .*/VERSION = "$(PYPI_VERSION)"/g' -e 's/^PLUGIN_VERSION = .*/PLUGIN_VERSION = "$(VERSION)"/g' ./bin/setup.py && \
-	rm ./bin/setup.py.bak && \
+	rm ./bin/setup.py.bak && rm ./bin/go.mod && \
```
viveklak, line 105:
```suggestion
	rm ./bin/setup.py.bak && \
	rm ./bin/go.mod && \
```

Makefile
```diff
@@ -126,9 +126,9 @@ build_go::
 	GOGC=50 go list github.com/pulumi/pulumi-azure-native/sdk/go/azure/... | grep -v "latest\|\/v.*"$ | xargs -L 1 go build
 
 clean::
-	rm -rf sdk/nodejs && mkdir sdk/nodejs && touch sdk/nodejs/go.mod
-	rm -rf sdk/python && mkdir sdk/python && touch sdk/python/go.mod && cp README.md sdk/python
-	rm -rf sdk/dotnet && mkdir sdk/dotnet && touch sdk/dotnet/go.mod
+	rm -rf $$(find sdk/nodejs -depth 1 -name "*" ! -name "go.mod")
+	rm -rf $$(find sdk/python -depth 1 -name "*" ! -name "go.mod" ! -name "README.md")
+	rm -rf $$(find sdk/dotnet -depth 1 -name "*" ! -name "go.mod")
```
viveklak, line 131: Make sure the find usage works across platforms (at least linux and mac os)

viveklak (approved):

Makefile
danielrbradley: Nice catch - yup - Ubuntu doesn't support `-depth` - only min- and max- 🤷

PR #1474, opened 2022-02-03
Tidy nightly commits

danielrbradley:
- ~Don't push noisy deletions.~
- Split SDK commit from schema to reduce commit diff sizes.

### DIFF
--- a/.github/workflows/nightly-sdk-generation.yml
+++ b/.github/workflows/nightly-sdk-generation.yml
@@ -42,6 +42,10 @@ jobs:
           git config --local user.email "bot@pulumi.com"
           git config --local user.name "pulumi-bot"
           git checkout -b generate-sdk/${{ github.run_id }}-${{ github.run_number }}
+      - name: commit changes
+        run: |
+          git add .
+          git commit -m "Preparing the SDK folder for regeneration"
       - name: Update submodules
         run: make update_submodules
       - name: Build codegen binaries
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-sdk-generation.yml
```diff
@@ -42,11 +42,6 @@ jobs:
           git config --local user.email "bot@pulumi.com"
           git config --local user.name "pulumi-bot"
           git checkout -b generate-sdk/${{ github.run_id }}-${{ github.run_number }}
-      - name: commit changes
-        run: |
-          git add .
-          git commit -m "Preparing the SDK folder for regeneration"
-          git push origin generate-sdk/${{ github.run_id }}-${{ github.run_number }}
```
mikhailshilkov, line 49: I think the reason it's here is when a file casing changes, you need separate delete/add commits, otherwise git won't register those and it will be a mess when you try to checkout locally. @stack72 may know the details.

danielrbradley: Ugh yeah ok that makes sense - been trying to figure out why this was being done. Can put this back if so - perhaps without the push until the end though.

danielrbradley: Ok, from everything I'm reading it's not an issue on linux as it's a case-sensitive filesystem. This approach would only be required on OSX or Windows.

mikhailshilkov: Hmm.. we definitely had issues because of that in the past. Feel free to give it a try though.

Issue #1202, opened 2021-10-07
Unable to destroy an ExtendedServerBlobAuditingPolicy 


thecodetinker:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Trying to set up an Azure SQL Database with auditing to a storage account. I can create the system, but pulumi destroy fails.

### Steps to reproduce

1. Add a `ExtendedServerBlobAuditingPolicy` resource to the Pulumi stack.
2. Run `pulumi up` to create the auditing policy.
3. Verify via the Portal that the Database auditing is successfully applied.
4. Run `pulumi destroy` to remove the stack.

Expected: The Azure resources including DB auditing should be removed.
Actual: The update fails with the following error message:

error: Code="InvalidServerBlobAuditingPolicyCreateRequest" Message="The create server blob auditing policy request does not exist or has no properties object."

This seems to happen regardless of whether I run `pulumi destroy` or simply remove the policy from the stack. The only course of action at this point seems to be to manually delete the database in the portal, but still pulumi doesn't seem great at recognising that the resource has already been deleted and tries to remove it again. It's very frustrating.

I should note that this is set up using a Storage Account behind a firewall, therefore not supplying the storage account key as per [here](https://docs.microsoft.com/en-us/azure/azure-sql/database/audit-write-storage-account-behind-vnet-firewall) and [here](https://github.com/MicrosoftDocs/azure-docs/issues/32068#issuecomment-723177293).

thecodetinker:
Thanks for the reply - the workaround to delete the state from the stack was very helpful. Interesting to see the linked issue as well, I haven't had it fail that bad but will look out for it.

prashantb1311:
> 
> 
> Thank you for reporting this @thecodetinker
> 
> If you deleted a database manually, you should run `pulumi refresh` and then the CLI should detect that it was deleted and remove it from the stack together with all policies.
> 
> As a simpler workaround, you can run `pulumi state delete <id of the policy resource>` to remove the policy from your stack.
> 
> Of course, we should fix the issue in the provider - but I hope the above can get you unblocked.

@mikhailshilkov : I am using Pulumi Automation API and I am not able to see "delete" command for stack while I can use "Destroy" command easily.
So i am stuck in the middle and not able to delete serverBlobAuditPolicy

[...]

PR #1490, opened 2022-02-08
Fix property name in blob auditing policy

### LINKED ISSUE #1481: Unable to destroy a ServerBlobAuditingPolicy
## Hello!

## Issue details

Pulumi destroy unable to delete a ServerBlobAuditingPolicy

Looks to be the same issue that was fix for ExtendedServerBlobAuditingPolicy in https://github.com/pulumi/pulumi-azure-native/issues/1202 is affecting ServerBlobAuditingPolicy

### Steps to reproduce

1. Create a new AzureNative.Sql.ServerBlobAuditingPolicy
2. Pulumi up
3. Ressource gets created
4. Pulumi destroy

Tested using pulumi CLI v3.23.2

Expected: Ressource is deleted
Actual: error: Code="InvalidServerBlobAuditingPolicyCreateRequest" Message="The create server blob auditing policy request does not exist or has no properties object."

Workaround: Using a ExtendedServerBlobAuditingPolicy  instead of a ServerBlobAuditingPolicy

danielrbradley:
Fixes #1481

Same problem as in #1202 which was fixed by #1322


mikhailshilkov:
Huh, so it doesn't look like a single outlier but rather a systemic problem. Do you think `auditingSettings` below may be affected?

I wonder if the property name changed from `status` to `state` at some point? That's the weakness of this approach - we don't know when defaults get out of sync. Would it be hard to add a test for these resources into one of the examples?

You need to change `metadata.json` (generate it locally and commit) to make the build pass.

viveklak (approved):

viveklak:
@danielrbradley you might need to rebase this?

danielrbradley:
Closing as it's pretty out of date and the attemptted fix seemed to have a number of bad side-effects.

PR #1530, opened 2022-02-23
Fix properties prefixed with numbers

### LINKED ISSUE #1528: Go code generation fails for property name starting with number
## Hello!

## Issue details

- Azure have released the `mobilenetwork` v2022-01-01-preview API spec.
- The `simPolicy.json` API defines a `DataNetworkConfiguration` object definition with a property called `5qi`.
- Attempting to generate the go SDK results in non-compilable go code.
- The process fails when trying to format the code before outputting it (and therefore we get no output files).

image

danielrbradley:
Fixes #1528

Following approach from https://github.com/pulumi/pulumi/blob/90913de3fbf1dac0908afda785149c83b27d7424/pkg/codegen/go/utilities.go#L81-L82 as suggested by @mikhailshilkov 

This might be solved in pulumi/pulumi as part of the SDK generation for incompatible names at some point.

### DIFF
--- a/provider/pkg/gen/utilities.go
+++ b/provider/pkg/gen/utilities.go
@@ -49,7 +49,11 @@ func ToLowerCamel(s string) string {
 	}
 	r := []rune(s)
 	if prefix, found := commonPrefixReplacements[r[0]]; found {
-		s = prefix + string(r[1:])
+		if len(r) > 1 {
+			s = prefix + strings.ToUpper(string(r[1])) + string(r[2:])
+		} else {
+			s = prefix
+		}
 	}
 	if r := rune(s[0]); r >= 'A' && r <= 'Z' {
 		s = strings.ToLower(string(r)) + s[1:]
--- a/provider/pkg/gen/utilities_test.go
+++ b/provider/pkg/gen/utilities_test.go
@@ -14,7 +14,7 @@ func TestToLowerCamel(t *testing.T) {
 		{"ID", "id"},
 		{"TTL", "ttl"},
 		{"_ref", "ref"},
-		{"5qi", "fiveqi"},
+		{"5qi", "fiveQi"},
 		{"6", "six"},
 	}
 	for _, i := range cases {


provider/pkg/gen/utilities_test.go
```diff
@@ -14,6 +14,8 @@ func TestToLowerCamel(t *testing.T) {
 		{"ID", "id"},
 		{"TTL", "ttl"},
 		{"_ref", "ref"},
+		{"5qi", "fiveqi"},
```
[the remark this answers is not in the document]
danielrbradley: Hmm, yeah, that looks better

provider/pkg/gen/utilities.go
```diff
@@ -32,8 +47,13 @@ func ToLowerCamel(s string) string {
 	if uppercaseAcronym[s] {
 		s = strings.ToLower(s)
 	}
-	if r := rune(s[0]); r == '_' {
-		s = s[1:]
+	r := []rune(s)
+	if prefix, found := commonPrefixReplacements[r[0]]; found {
+		if len(r) > 1 {
+			s = prefix + strings.ToUpper(string(r[1])) + string(r[2:])
```
mikhailshilkov, line 53:
```suggestion
			s = prefix + strings.ToUpper(string(r[1])) + r[2:]
```
I think this may work

mikhailshilkov (approved):
LGTM

provider/pkg/gen/utilities.go
danielrbradley: Nope .. `invalid operation: mismatched types string and []rune` - think it needs the explicit conversion

mikhailshilkov: How does this work then? https://github.com/pulumi/pulumi-azure-native/blob/95c2dc8b77fc3675ed8328374a1e18ee2af7608f/provider/pkg/gen/utilities.go#L59

mikhailshilkov: Ah, I missed `r := []rune(s)`! I see.

PR #1537, opened 2022-02-24
Fix case where there's no known available versions

### LINKED ISSUE #1535: Solutions namespace versions removed
## Hello!

## Issue details

The latest automated SDK upgrades are intoducing breaking changes to the `solutions` namespace. 

We are pinned at `2019-07-01` for our default version but it's now reverting to `2016-09-01-preview`. This seems to be due to the `az providers list` command now returning different versions:

image

- This no longer includes our pinned version.
- This only includes a preview version.
- There is no specification for the only listed version (`2021-02-01-preview`)

The behaviour is a little odd now there's no cross-over with the listed versions – reverting to the oldest preview version.

danielrbradley:
Fixes #1535

If there's no known available versions for a resource, operate like all versions are known.

### DIFF
--- a/provider/pkg/openapi/versioner.go
+++ b/provider/pkg/openapi/versioner.go
@@ -360,6 +360,7 @@ func (c *versioner) calculateLatestVersions(provider string, versionMap Provider
 	sort.Strings(previews)
 	sort.Strings(stables)
 	versions := append(previews, stables...)
+	availableVersions := codegen.NewStringSet(versions...)
 
 	pathTypeNames := map[string]string{}
 	latestResources = map[string]*ResourceSpec{}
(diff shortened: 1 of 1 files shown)


jkodroff:
I see there is a unit test file: https://github.com/pulumi/pulumi-azure-native/blob/master/provider/pkg/openapi/versioner_test.go

Is it feasible to add unit tests to cover this use case?

stack72 (approved):

danielrbradley:
@jkodroff agree this should be tested. Will aim to follow up with a unit test here - the build cycle time is too high to add it on now though. Have tested manually by re-generating locally and checking the diff.

PR #1559, opened 2022-03-10
Pin network group resource version

danielrbradley:
Avoid upgrading to next preview version (now in the versions list) which has breaking changes.


danielrbradley:
Cancelling build and merging as no actual code changes

PR #1587, opened 2022-03-21
Output default versions

danielrbradley:
Make the creation of the default version more transparent to help with upgrades.

Currently if there's breaking changes in the default version it can be due to a number of reasons:
- A new major version being available.
- A new resource in an existing version.
- An old version not being available.

This should help highlight where breaking changes are coming from with the diff.


stack72 (approved):
<3 this!

viveklak (approved):

PR #1607, opened 2022-03-29
[internal] pin Confluent API version to v20200301 to avoid incoming breaking changes

stack72:


provider/pkg/openapi/versioner.go
```diff
@@ -113,6 +113,7 @@ var cutoffProviderVersions = map[string]string{
 	"cognitiveservices":             "v20170418",
 	"compute":                       "v20210301",
 	"confidentialledger":            "v20201201preview",
+	"confluent": 					 "v20200301",
```
mikhailshilkov, line 116: Go fmt

mikhailshilkov (approved):

danielrbradley (approved):

PR #1610, opened 2022-03-30
Publish arm2pulumi binaries as part of release

viveklak:


stack72:
@viveklak should we do what we do for aws-native and have a separate release mechanism for this?

viveklak:
> @viveklak should we do what we do for aws-native and have a separate release mechanism for this?

Ah I didn't realize we had that in place. Certainly makes sense. Will switch to that mechanism unless you want to take this over.

viveklak:
See #1625 instead.

PR #1643, opened 2022-04-11
Use native automerge for sdk generation and pulumi/pulumi updates

jkisk:
Resolves #https://github.com/pulumi/home/issues/1964


jkodroff (approved):
Approved, but now sure how we are gonna know that an update happened and that we should (probably?) release the provider. (Note that this is not the only workflow with this issue.)

jkisk:
> Approved, but now sure how we are gonna know that an update happened and that we should (probably?) release the provider. (Note that this is not the only workflow with this issue.)

This PR doesn't introduce the concept of automerging, just updates the mechanism.  I am curious though why the old action was set to allow empty PRs, I can carry that forward if needed.

PR #1663, opened 2022-04-20
Add Attach method to Provider

Frassle:
Update to pulumi v3.30.0 and implement the Attach method on the provider.


Frassle:
Done by another PR

PR #1678, opened 2022-04-28
Better timeout error

iwahbe:
Better error if an example hangs while generating

### DIFF
--- a/provider/pkg/gen/examples.go
+++ b/provider/pkg/gen/examples.go
@@ -1,4 +1,4 @@
-// Copyright 2021, Pulumi Corporation.  All rights reserved.
+// Copyright 2022, Pulumi Corporation.  All rights reserved.
 
 package gen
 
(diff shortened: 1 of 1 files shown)


provider/pkg/gen/examples.go
```diff
@@ -262,14 +263,33 @@ func generateExamplePrograms(example resources.AzureAPIExample, body *model.Body
 	return languageExample, nil
 }
 
-func recoverableProgramGen(program *hcl2.Program, fn programGenFn) (files map[string][]byte, err error) {
+// Panic after 3 minutes, unless cancel is called.
+func makeTimeout(timer string) (cancel func()) {
```
viveklak, line 267: Perhaps you could just use `context.WithTimeout` instead? https://pkg.go.dev/context#WithTimeout

iwahbe: I believe that would require either
a. That `GenerateProgram` is `context.Context` aware or
b. a similar channel based mechanism built on top of `context.Context.Done()`. 

For (b), I though that just generating the channel directly was cleaner.

viveklak (approved):

PR #1689, opened 2022-05-05
Substantially reduce memory usage

AaronFriel:
These changes reduce peak memory usage during yaml executions by over 400MiB by reducing the resident set when the schema is requested. The change is from 1038MiB (Azure Native 1.64.0) to 594MiB (this commit).

A few changes are made to accomplish this:

1. A replacement library for encoding/json, github.com/segmentio/encoding/json, is used with its zero copy mode for decoding, reducing the cost to unmarshal the schema on demand.
2. The schema is in the `.rodata` section of the binary via `//go:embed` directives that embed it as a string, placing it in shareable, read-only memory.
3. The schema and metadata are stored in the binary *uncompressed*, ensuring that reading the schema does not require copying it into non-shareable memory.

These changes contribute reducing the runtime memory usage substantially when one or more Azure Native providers is running.

Coupled with changes that reduce the time to run `pulumi preview` on the [azure-app-service](https://github.com/pulumi/pulumi-yaml/tree/main/examples/azure-app-service) from around 12s to 6s, and memory usage by over 700 MiB by further reducing the need to host the plugin during YAML program execution.

Edit:

1. Further improvements are made by partially parsing the metadata file
2. By using a bit of carefully written unsafe around the string constant, we avoid copying the schema or metadata before parsing. (String/[]byte casts copy.)
3. A generic map type is introduced to make accessing the partially parsed values more ergonomic.

### DIFF
--- a/.github/workflows/arm2pulumi-release.yml
+++ b/.github/workflows/arm2pulumi-release.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Install Go 1.17.x
         uses: actions/setup-go@v2
         with:
-          go-version: '1.17.x'
+          go-version: '1.18.x'
       - name: Set Version if Parameter available
         if: github.event.inputs.version != ''
         run: echo "GORELEASER_CURRENT_TAG=v${{ github.event.inputs.message }}" >> $GITHUB_ENV
--- a/.github/workflows/run-acceptance-tests.yml
+++ b/.github/workflows/run-acceptance-tests.yml
@@ -53,7 +53,7 @@ jobs:
       fail-fast: true
       matrix:
         goversion:
-        - 1.17.x
+        - 1.18.x
         dotnetversion:
         - 3.1.301
         pythonversion:
@@ -144,7 +144,7 @@ jobs:
       fail-fast: true
       matrix:
         goversion:
-        - 1.17.x
+        - 1.18.x
         dotnetversion:
         - 3.1.301
         pythonversion:
@@ -235,7 +235,7 @@ jobs:
       fail-fast: true
       matrix:
         goversion:
-        - 1.17.x
+        - 1.18.x
         dotnetversion:
         - 3.1.301
         pythonversion:


provider/cmd/arm2pulumi/main.go
```diff
@@ -104,20 +114,22 @@ func loadMetadata() (*resources.AzureAPIMetadata, error) {
 // generation from schema.go
 func loadSchema() (*schema.PackageSpec, error) {
 	var pkgSpec schema.PackageSpec
-	uncompressed, err := gzip.NewReader(bytes.NewReader(pulumiSchema))
```
mikhailshilkov, line 107: How does this affect the binary size?

AaronFriel: Unpacked, the binary is ~200MiB vs 62MiB. The distributed .tar.gz is approximately the same size.

AaronFriel: 206MiB built on this branch.

AaronFriel: @mikhailshilkov per the conversation on this PR:
- https://github.com/pulumi/pulumi/pull/9686

We could consider moving the schema.json outside of the binary - no embedding - while retaining the performance benefits.

That would reduce the binary size. We'd need to agree on the naming convention for packaging the schema as a separate file.

danielrbradley: @AaronFriel moving the schema.json outside the binary has also come up on a separate thread of conversation our side and I'd be quite keen to move towards doing this – just include the schema.json in the gzip package. The motivator our side was to build a single, identical binary once per commit hash. Then the version number would be read from the schema.json rather than being embedded in the binary, meaning to promote a pre-release to a release would just involve updating the schema.json and re-gzipping. This would approximately half our time to release which can be over 4 hours on some repos.

Personally, this also makes more sense from a gzipping point of view - it gets unpacked on the first install, then it's faster to read on every instantiation of the plugin.

AaronFriel: > moving the schema.json outside the binary has also come up on a separate thread of conversation our side and I'd be quite keen to move towards doing this

Maybe you, @pgavlin & I should chat so we can align on conventions here?

AaronFriel:
@danielrbradley I've put the changes to put the schema file in a separate file on a branch here: https://github.com/pulumi/pulumi-azure-native/pull/new/tradeoffs-sxs

If this PR looks good to you, I'd love to ship this and turn the YAML language integration tests back on for Azure.

danielrbradley (approved):
Given this a test locally and looks good to go! :shipit:

AaronFriel:
Rebasing and merging :)

PR #1635, opened 2022-04-08
Update getDefaultName to use NewUniqueHexV2

Frassle:

PR #1720, opened 2022-05-18
Revert "Update getDefaultName to use NewUniqueHexV2"

stack72:
Reverts pulumi/pulumi-azure-native#1635

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## HEAD (Unreleased)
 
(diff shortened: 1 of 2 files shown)


Zaid-Ajaj:
The build [here](https://github.com/pulumi/pulumi-azure-native/runs/6487046195?check_suite_focus=true) is failing because it is running [out of memory](https://github.com/dotnet/roslyn/issues/39276#issuecomment-541990939), I think it just needs a re-run

Zaid-Ajaj (approved):

Issue #1606, opened 2022-03-29
Nightly Codegen: panic when types are incompatible


stack72:
When new versions of resources are released, there is a chance that there are multiple types generated of the same name that may be incompatible types

We currently override the panic by using the following:

```
if tok == "azure-native:authorization:PrincipalResponse" && len(v.Properties) == 2 ||
```

As this *hides* the panic, we should remove the panic and these overrides and allow the changes to bubble up to the schema so that we can check the difference in the schema

viveklak:
I didn;t get a chance to address this in https://github.com/pulumi/pulumi-azure-native/pull/1739

danielrbradley:
This is largely solved by when we added a type merge operation. No conflicts have happened since.

PR #1739, opened 2022-05-31
Unwedge auto-update

viveklak:
#1606 is still a problem but this unwedges the nightly auto-update for now.


viveklak:
@danielrbradley the specific fix is here: https://github.com/pulumi/pulumi-azure-native/pull/1739/commits/94b147cc625350803856070fa58613df1b82ce88. I did run the update submodules target which perhaps wasn't the right call? I can revert that part and just commit the above change if that helps.

danielrbradley:
Thanks for kicking this off @viveklak 

Looks like there's an issue with an enum and a resource with the same name causing the dotnet build to fail. Would be good to unblock this ahead of pushing ahead with the v2 release. Will see how far I get today.

stack72:
Closed via #1743

PR #1790, opened 2022-06-24
V2 version selection

danielrbradley:
The `v2-config.yaml` is a simple file which is generated once before the version change and is designed to be hand-maintained for the duration of the major version. From the config we can generate the `v2.json` which is akin to a lock file - with the full expanded list of resources included. Once we've created the config, we can re-create the `v2.json` file at any point to pick up minor adjustments. 

Updated process:

1. Read full spec
2. Remove currently deprecated APIs (as these will be remove next version)
3. Find candidate versions - prefer stable versions if we have recent stable versions (i.e. within 1 year of a preview - somewhat arbitary time but seems to work)
4. Find minimal set of API versions which can provide all resources
5. Pick latest version for each resource

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,15 +37,19 @@ pulumi up
 
 Key facts about Azure Versions:
 
-- Azure's APIs have many versions of each service which are published per-service on their own schedule.
-- API versions are in the form of an ISO date (e.g. `2020-01-01`) followed by an optional suffix (e.g. `-preview` or `-privatepreview`).
-- Not all versions of the API contain all resources. Some APIs will only publish the parts of their API which have changed since the previous version.
-- Some APIs haven't had a stable (non-suffixed) release for a number of years - and the 'preview' versions are generally considered stable for every-day use.
-- API version specifications are sometimes release before the API version is deployed.
+- Azure's REST API has many versions of each service which are published per-service on their own schedule.
+- Service versions are in the form of an ISO date (e.g. `2020-01-01`) followed by an optional suffix (e.g. `-preview` or `-privatepreview`).
+- Not all versions of the service contain all resources. Some services will only publish the parts of their API which have changed since the previous version.
+- Some services haven't had a stable (non-suffixed) release for a number of years - and the 'preview' versions are generally considered stable for every-day use.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,107 @@
...
+
+To provide simpler discovery and usability, the Pulumi Azure Native provider adds a single 'default' version for each API which is the combination of one or more API versions. The aim is to include all resources, at the latest available version, where possible.
+
+#### Provider v1
+
+In version 1 of the provider, the default version was recalculated every time a new API version was released. This led to frequent breaking changes which had to be fixed before release of an update.
+
+#### Provider v2
+
+In version 2 of the provider we aim to specify more precisely the versions to combine into the 'default version' to avoid SDK instability. This is specified in the [`v2-config.yaml`](./v2-config.yaml). For each API the config can specify a `Tracking` version and `Additions`.
```
viveklak, line 80: Nit - `s/API/resource provider`?

viveklak (approved):
Generally looks great! Thanks for the tests and the docs in CONTRIBUTING.md - which made it a lot clearer to follow along. Some minor suggestions but otherwise looks good to me.

provider/pkg/versioning/defaultVersion.go
```diff
@@ -0,0 +1,255 @@
...
+	candidateVersions := filterCandidateVersions(versions)
+	candidates := filterVersionResources(versions, candidateVersions)
+	minimalVersionSet := findMinimalVersionSet(candidates)
+	minimalVersions := filterVersionResources(candidates, minimalVersionSet)
+
+	var orderedVersions []openapi.ApiVersion
+	for apiVersion := range minimalVersions {
+		orderedVersions = append(orderedVersions, apiVersion)
+	}
+	sort.Strings(orderedVersions)
```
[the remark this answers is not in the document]
danielrbradley: I was trying to find something just like this! 👍

CONTRIBUTING.md
danielrbradley: Have gone with a half-way change here to just add more clarification around terminology.

viveklak (approved):
Updates LGTM!

PR #1829, opened 2022-07-15
Add Java packages overrides to schema

pawelprazak:
Part of https://github.com/pulumi/pulumi-java/issues/381

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## HEAD (Unreleased)
 
 - Memory usage for most programs is greatly reduced [#1689](https://github.com/pulumi/pulumi-azure-native/pull/1689)
+- All Java packages changed to be more intuitive, unfortunately this is a breaking change for Java version [1829](https://github.com/pulumi/pulumi-azure-native/pull/1829)
 
 ## 1.67.0 (2022-07-12)
 
--- a/provider/pkg/gen/schema.go
+++ b/provider/pkg/gen/schema.go
@@ -184,6 +184,9 @@ func PulumiSchema(providerMap openapi.AzureProviders) (*pschema.PackageSpec, *re
 	csharpNamespaces := map[string]string{
 		"azure-native": "AzureNative",
 	}
+	javaPackages := map[string]string{
+		"azure-native": "azurenative",
+	}
 	pythonModuleNames := map[string]string{}
 	golangImportAliases := map[string]string{}
 
@@ -210,12 +213,14 @@ func PulumiSchema(providerMap openapi.AzureProviders) (*pschema.PackageSpec, *re
 				examples:   exampleMap,
 			}
 
-			// Populate C#, Python and Go module mapping.
+			// Populate C#, Java, Python and Go module mapping.
 			module := gen.providerToModule(providerName)
 			csharpNamespaces[strings.ToLower(providerName)] = providerName
+			javaPackages[module] = strings.ToLower(providerName)
 			if version != "" {
 				csVersion := strings.Title(csharpVersionReplacer.Replace(version))
 				csharpNamespaces[module] = fmt.Sprintf("%s.%s", providerName, csVersion)
+				javaPackages[module] = fmt.Sprintf("%s.%s", strings.ToLower(providerName), version)
 			}
 			pythonModuleNames[module] = module
 			golangImportAliases[filepath.Join(goBasePath, module)] = strings.ToLower(providerName)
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## HEAD (Unreleased)
 
 - Memory usage for most programs is greatly reduced [#1689](https://github.com/pulumi/pulumi-azure-native/pull/1689)
+- All Java packages changed to be more intuitive, unfortunately this is a breaking change for Java version [1829](https://github.com/pulumi/pulumi-azure-native/pull/1829)
```
t0yv0, line 6: This could really use some love - let's make it helpful and give an example of what the user needs to edit in their program after upgrading to the new version of the provider, could we do that? Will this require a user to rewrite imports?

In https://github.com/pulumi/templates/compare/master...pulumi-java/main I see an example for k8s but not for azure-native. 

```
import com.pulumi.kubernetes.apps_v1.Deployment;
import com.pulumi.kubernetes.apps_v1.DeploymentArgs;
import com.pulumi.kubernetes.apps_v1.inputs.DeploymentSpecArgs;
import com.pulumi.kubernetes.core_v1.inputs.ContainerArgs;
import com.pulumi.kubernetes.core_v1.inputs.ContainerPortArgs;
import com.pulumi.kubernetes.core_v1.inputs.PodSpecArgs;
import com.pulumi.kubernetes.core_v1.inputs.PodTemplateSpecArgs;
import com.pulumi.kubernetes.meta_v1.inputs.LabelSelectorArgs;
import com.pulumi.kubernetes.meta_v1.inputs.ObjectMetaArgs;
--->
import com.pulumi.kubernetes.apps.v1.Deployment;
import com.pulumi.kubernetes.apps.v1.DeploymentArgs;
import com.pulumi.kubernetes.apps.v1.inputs.DeploymentSpecArgs;
import com.pulumi.kubernetes.core.v1.inputs.ContainerArgs;
import com.pulumi.kubernetes.core.v1.inputs.ContainerPortArgs;
import com.pulumi.kubernetes.core.v1.inputs.PodSpecArgs;
import com.pulumi.kubernetes.core.v1.inputs.PodTemplateSpecArgs;
import com.pulumi.kubernetes.meta.v1.inputs.LabelSelectorArgs;
import com.pulumi.kubernetes.meta.v1.inputs.ObjectMetaArgs;
```

pawelprazak: would you like me to attempt at creating a changelog you have in mind or you intend to do it yourself, not clear sorry

t0yv0: It's the expectation that the PR owner writes a changelog 🙏 LMK if you need help.

> It is the same kind of change that went into the kubernetes provider.

Well that's unfortunate. The changelogs are user facing and especially valuable for breaking changes or when investigating bugs. If possible to clarify it, it could be a time-saver for users down the line. TYVM.

pawelprazak: I've added more info to the changelog, AFAIK this change only affects full schema, that is not used ATM.

pawelprazak:
Rebased and updated CHANGELOG

CHANGELOG.md
t0yv0: I'm still confused... Suppose I am a user of com.pulumi:azure-native , and I upgrade to a version with this in the CHANGELOG. What do I need to know to modify my programs? If the answer is nothing, we don't need a CHANGELOG. If I need to modify my programs (this is breaking) would be lovely to have an example.

pawelprazak: fixing imports should be the only required action

pawelprazak: but given that in the short schema there were no overrides, this should have no impact in practice, I know confusing, sorry about that

t0yv0: Could we put an example of how an import update looks like into the changelog? That'd be very helpful to users.

pawelprazak: The new provider is not generated, so I don't know specific imports that would change.

Every versioned package will change, but since those are only present in `schema-full` there will be no change, because, we've never generated this provider form ful schema.

Please let me know what more I can say to make this clear.

t0yv0: Ah, perfect, thanks for explaining that. Then we don't need the CHANGELOG entry at all, since the users are not affected. There is a label "impact/no-changelog-required".

t0yv0 (approved):

t0yv0 (approved):

t0yv0:
The "packages" field doc says

```
Overrides for module names to Java package names. Example: “autoscaling/v1” -> “autoscaling.v1”.
```

@stack72 so I think this should go in when possible but not urgent as it seems that no code is yet affected for the user. This is because Java builds from a small (not full) schema that does not have the "v1/v2/v3" APIs. Building from a full schema currently stretches the limits of our Gradle build setup and is not possible (blocked) until we figure out how to build such a massive project.

stack72:
hey @pawelprazak I finally merged this - thank you for all of the work here!!!

stack72:
Closed in https://github.com/pulumi/pulumi-azure-native/commit/6359eec0aae0201ba9cdff4970c07d15a312dd86

pawelprazak:
@stack72 thank you for making this happen! :)

PR #1833, opened 2022-07-15
Represent user assigned identity inputs as string arrays

### LINKED ISSUE #812: Simplify user-assigned identities type
Currently assigning user-assigned identities requires jumping through hoops, e.g.:
In typescript
https://github.com/pulumi/pulumi-azure-native/blob/master/examples/keyvault/index.ts#L33

Assigning user assigned identities is pretty common use-case. 

A replacement of the following form would be much more usable `userAssignedIdentities: {[userIdentity.id]: {}}`.

danielrbradley:
Fixes #812 

- If the object type has no value type defined then it's just a string set.
- Revert genTypeSpec filling missing value types with "any" as we'll handle it outside.
- Add metadata for "IsStringSet" so we know we need to convert it back in the provider.
- We don't need to validate string sets as the value should match the defined type (string array).
- When converting, intercept string sets and convert the array back to a map with empty values.

Summary of affected resources:
```
Resource "azure-native:azurestackhci:Cluster"
Type "azure-native:app:ManagedServiceIdentity"
Type "azure-native:appconfiguration:ResourceIdentity"
Type "azure-native:appplatform:ManagedIdentityProperties"
Type "azure-native:authorization:Identity"
Type "azure-native:automation:Identity"
Type "azure-native:batch:BatchAccountIdentity"
Type "azure-native:batch:BatchPoolIdentity"
Type "azure-native:cache:ManagedServiceIdentity"
Type "azure-native:cognitiveservices:Identity"
Type "azure-native:compute:EncryptionSetIdentity"
Type "azure-native:compute:VirtualMachineIdentity"
Type "azure-native:compute:VirtualMachineScaleSetIdentity"
Type "azure-native:containerinstance:ContainerGroupIdentity"
Type "azure-native:containerservice:ManagedClusterIdentity"
Type "azure-native:dashboard:ManagedServiceIdentity"
Type "azure-native:databox:ResourceIdentity"
Type "azure-native:dbformysql:Identity"
Type "azure-native:devcenter:ManagedServiceIdentity"
Type "azure-native:devcenter:ProjectEnvironmentTypeUpdatePropertiesCreatorRoleAssignment"
Type "azure-native:devcenter:UserRoleAssignment"
Type "azure-native:devices:ArmIdentity"
Type "azure-native:deviceupdate:ManagedServiceIdentity"
Type "azure-native:digitaltwins:DigitalTwinsIdentity"
Type "azure-native:documentdb:ManagedServiceIdentity"
Type "azure-native:eventhub:Identity"
Type "azure-native:fluidrelay:Identity"
Type "azure-native:healthbot:Identity"
Type "azure-native:healthcareapis:ServiceManagedIdentityIdentity"
Type "azure-native:insights:DataCollectionEndpointResourceIdentity"
Type "azure-native:insights:DataCollectionRuleResourceIdentity"
Type "azure-native:insights:WorkbookResourceIdentity"
Type "azure-native:kusto:Identity"
Type "azure-native:loadtestservice:ManagedServiceIdentity"
Type "azure-native:logic:ManagedServiceIdentity"
Type "azure-native:machinelearningservices:ManagedServiceIdentity"
Type "azure-native:media:MediaServiceIdentity"
Type "azure-native:mobilenetwork:ManagedServiceIdentity"
Type "azure-native:netapp:ManagedServiceIdentity"
Type "azure-native:network:ManagedServiceIdentity"
Type "azure-native:operationalinsights:Identity"
Type "azure-native:purview:Identity"
Type "azure-native:recoveryservices:IdentityData"
Type "azure-native:resources:Identity"
Type "azure-native:resources:ManagedServiceIdentity"
Type "azure-native:servicebus:Identity"
Type "azure-native:servicefabric:ManagedIdentity"
Type "azure-native:signalrservice:ManagedIdentity"
Type "azure-native:solutions:Identity"
Type "azure-native:sql:DatabaseIdentity"
Type "azure-native:sql:ResourceIdentity"
Type "azure-native:storage:Identity"
Type "azure-native:storagecache:AmlFilesystemIdentity"
Type "azure-native:storagecache:CacheIdentity"
Type "azure-native:synapse:ManagedIdentity"
Type "azure-native:videoanalyzer:VideoAnalyzerIdentity"
Type "azure-native:videoindexer:ManagedServiceIdentity"
Type "azure-native:virtualmachineimages:ImageTemplateIdentity"
Type "azure-native:web:ManagedServiceIdentity"
Type "azure-native:webpubsub:ManagedIdentity"
Type "azure-native:workloads:UserAssignedServiceIdentity"
```

### DIFF
--- a/examples/azure-native-sdk/go-user-assigned-identity/main.go
+++ b/examples/azure-native-sdk/go-user-assigned-identity/main.go
@@ -25,7 +25,7 @@ func main() {
 		}
 
 		// Create an Azure resource (Storage Account)
-		account, err := storage.NewStorageAccount(ctx, "sa", &storage.StorageAccountArgs{
+		_, err = storage.NewStorageAccount(ctx, "sa", &storage.StorageAccountArgs{
 			ResourceGroupName: resourceGroup.Name,
 			Sku: &storage.SkuArgs{
 				Name: pulumi.String("Standard_LRS"),


kuznero:
Any progress on this? Anything I can help with to get it moving?

kuznero:
What is the accepted workaround before 2.0 is landed? Is it relying on KubernetesCluster of Azure Core package?

kuznero:
> @kuznero there's some discussion of the workarounds on the original issue: #812
> 
> Essentially, you just have a map where the key is the string value and the value is an empty map.

Unfortunately, I am yet to see it working in Go. All the variations mentioned in all the comments throughout internet do not work in Go unfortunately. If there is one I missed, please share the snippet.

provider/pkg/gen/schema.go
```diff
@@ -1402,6 +1379,47 @@ func (m *moduleGenerator) getOneOfValues(property *pschema.PropertySpec) (values
...
+				MinLength:            resolvedProperty.MinLength,
+				MaxLength:            resolvedProperty.MaxLength,
+				Pattern:              resolvedProperty.Pattern,
+				Items:                m.itemTypeToProperty(propertySpec.Items),
+				AdditionalProperties: m.itemTypeToProperty(propertySpec.AdditionalProperties),
+			}
+
+			// HACK: If there's no object value type, then it's just a set of strings which we'll represent as a string
+			// array in the SDK, but leave the metadata to indicate we need to convert it.
+			if propertySpec.Type == "object" && propertySpec.AdditionalProperties == nil && propertySpec.Ref == "" {
```
thomas11, line 1410: Modifying one of the method arguments is a bit yucky. Since this func is about API properties, should the metadata change rather be outside of it?

provider/pkg/provider/provider.go
```diff
@@ -633,6 +633,18 @@ func (k *azureNativeProvider) validateProperty(ctx string, prop *resources.Azure
 
 		failures = append(failures, k.validateType(ctx, &typ, value)...)
 	case []interface{}:
+		// If the prop should be an object with empty values, then validate the input as a string set.
+		if prop.Type == "object" && prop.AdditionalProperties == nil {
```
thomas11, line 637: Why do we check for `propertySpec.Ref == ""` in the conversion method but not here?

If the two checks were the same, we could extract them into a helpfully named helper.

provider/pkg/resources/convert.go
```diff
@@ -96,12 +96,30 @@ func (k *SdkShapeConverter) convertPropValue(prop *AzureAPIProperty, value inter
...
+				Type: "string",
+			}
+			mapResult := map[string]interface{}{}
+			emptyValue := struct{}{}
+			for i := 0; i < s.Len(); i++ {
+				item := k.convertPropValue(&stringProp, s.Index(i).Interface(), convertMap)
+				switch v := item.(type) {
+				case string:
+					mapResult[v] = emptyValue
+				default:
```
thomas11, line 113: It's ok to ignore non-strings here?

examples/go-user-assigned-identity/main.go
```diff
@@ -0,0 +1,62 @@
...
+			Identity: &storage.IdentityArgs{
+				Type:                   pulumi.String("UserAssigned"),
+				UserAssignedIdentities: pulumi.StringArray{userAssignedId.ID()},
+			},
+		})
+		if err != nil {
+			return err
+		}
+
+		// Export the primary key of the Storage Account
```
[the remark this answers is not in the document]
danielrbradley: Think it was just a copy/paste from a storage account example. Pretty sure we could just remove that bit.

provider/pkg/provider/provider.go
danielrbradley: Reworked this to use an explicit `IsStringSet` option instead so the type is actually array of string and we don't have to do anything special within the validate step.

provider/pkg/gen/schema.go
danielrbradley: Fixed!

examples/go-user-assigned-identity/main.go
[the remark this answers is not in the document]
danielrbradley: Nope - just copied from an existing example. Have slimmed this test down a little more now.

provider/pkg/resources/convert.go
danielrbradley: Yup - non-strings will be handled by validate. Have tweaked and added a comment to explain 👍

provider/pkg/provider/provider.go
danielrbradley: Reworked this to use an explicit `IsStringSet` option instead so the type is actually array of string and we don't have to do anything special within the validate step.

provider/pkg/gen/schema.go
danielrbradley: Fixed!

thomas11 (approved):

PR #1855, opened 2022-07-26
java auto publish

stack72:
- [internal] Adding the work to auto-publish the java sdk
- Adding Java SDK generated files

This PR brings in #1829 as part of it to ensure the generation is correct


sdk/java/src/main/java/com/pulumi/azurenative/Utilities.java
```diff
@@ -75,7 +75,7 @@ public static String getVersion() {
     }
 
     static {
-        var resourceName = "com/pulumi/azure-native/version.txt";
+        var resourceName = "com/pulumi/azurenative/version.txt";
```
pawelprazak, line 78: hmm, need to make sure this file gets created properly by the `build.gradle`:
`sdk/java/src/resources/com/pulumi/azurenative/version.txt`

Like here: https://github.com/pulumi/pulumi-kubernetes/blob/cbc8c830eda8bd0fea78919d0f53ca7275a0086f/sdk/java/build.gradle#L68

danielrbradley:
As discussed, let's build from the `schema-full.json` to get the API versions too. This is a build artifact and not committed, so we need to ensure it's there before the generate_java step.

This will massively increase the SDK size so we might want to not commit the Java SDK.

sdk/java/src/main/java/com/pulumi/azurenative/Utilities.java
pawelprazak: FYI: released in Java SDK `0.5.4`

t0yv0:
AFAIK Java SDK currently won't be able to build from the FULL schema in any reasonable time (CPU time + memory). Last time we tried builds were not finishing in 30 min. We can reopen a ticket in pulumi/pulumi-java.

If it is possible somehow to keep releasing azure-native from the partial schema for now we'd really appreciate that.

PR #1879, opened 2022-08-09
Spike - testing lazy TS module approaches

danielrbradley:
Two possible approaches in the two commits (have pretty similar outcomes for peformance):
1. Gnarly typescript - less hacky but produces slightly odd exports
2. SED replacement after TSC completes

Testing results using `examples/simple`:

Existing timings for `pulumi preview`
`import * as web from "@pulumi/azure-native/web"`  - 9 seconds
`import { web } from "@pulumi/azure-native"`  - 16 seconds
`import * as az from "@pulumi/azure-native"` - 36 seconds

New timings for `pulumi preview` ...
`import * as web from "@pulumi/azure-native/web"`  - 9 seconds
`import { web } from "@pulumi/azure-native"`  - 9 seconds
`import * as az from "@pulumi/azure-native"` - 9 seconds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,7 @@ build_nodejs::
 	cd ${PACKDIR}/nodejs/ && \
 	yarn install && \
 	NODE_OPTIONS=--max-old-space-size=8192 yarn run tsc --diagnostics && \
+	sed -i.bak '/^exports\..* = require("\..*");/d' bin/index.js && \
 	cp ../../README.md ../../LICENSE package.json yarn.lock ./bin/ && \
 	sed -i.bak -e "s/\$${VERSION}/$(VERSION)/g" ./bin/package.json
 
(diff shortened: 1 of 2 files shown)


t0yv0:
Hey @danielrbradley  sorry we somehow missed this, @RobbieMcKinstry is also currently working on the codegen change to  support lazy loading of modules. We don't have numbers yet but we have an approach that seems to work, coming shortly!

Just connecting in case we can benefit from any of your work here.

RobbieMcKinstry:
Apologies for not communicating this out earlier, but we're implementing this approach in codegen 🎉
If you don't mind, please keep this draft PR open for now, since there are a few differences between what I have in codegen and what's here that I plan to reference and ask you about :D

danielrbradley:
@t0yv0 @RobbieMcKinstry let's connect early next week. I was pretty close to having this shippable today but if we've got a working alternative in the next couple of weeks then hopefully we could switch over. The tricky part was more around how the module interdependence ties into the local build, test and publishing workflows.

t0yv0:
Would love to understand the extra constraints since we've been envisioning an approach where the rollout is just a completely transparent dependency update on Node codegen.

PR #1887, opened 2022-08-10
Use new versioner to calculate existing v1 lock file

danielrbradley:
There's probably more places we could simplify the v1 config to use tracking without affecting the v1.json lock file but need to experiment more here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,18 +75,18 @@ The default version is calculated and written to a 'lock' file which list every 
 
 To provide simpler discovery and usability, the Pulumi Azure Native provider adds a single 'default' version for each API which is the combination of one or more API versions. The aim is to include all resources, at the latest available version, where possible.
 
-#### Provider v1
+Originally the default version was recalculated every time a new API version was released. This led to frequent breaking changes which had to be fixed before release of an update.
 
-In version 1 of the provider, the default version was recalculated every time a new API version was released. This led to frequent breaking changes which had to be fixed before release of an update.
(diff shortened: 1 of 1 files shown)


provider/cmd/pulumi-versioner-azure-native/main.go
```diff
@@ -183,23 +183,23 @@ func pending(outputDir, versionFile string) error {
...
 
-	v1 := openapi.CalculateProviderDefaults(activePathVersions, providers)
+	v2, err := versioning.DefaultConfigToCuratedVersion(specVersions, v2Config)
 	if err != nil {
 		return err
 	}
 
 	return emitFiles(outputDir, map[Filename]Data{
-		"v1.json": v1,
+		"v1.json": v2,
```
stack72, line 202: Why are we writing v2 config to a v1 json file?

danielrbradley: Copy pasta error - didn't rename the variable .. it is v1 🤦

Makefile
```diff
@@ -86,7 +86,8 @@ versions/spec-resources.json: bin/pulumi-versioner-azure-native .git/modules/azu
 versions/active.json: bin/pulumi-versioner-azure-native azure-provider-versions/provider_list.json
 	bin/pulumi-versioner-azure-native active
 
-versions/v1.json: bin/pulumi-versioner-azure-native .git/modules/azure-rest-api-specs/HEAD azure-provider-versions/provider_list.json
+V1CONFIG := $(wildcard versions/v1-config.yaml)
```
viveklak, line 89: Do we need the wildcard expansion here or are we planning to add one in the future?

danielrbradley: Nope, removed!

viveklak (approved):

PR #1920, opened 2022-08-18
Adding Java builds to the repository

stack72:
We are not checking in the generated sources here but we do ensure
that the gradle files are available to check on the dependencies


t0yv0:
🤞  curious to see build times on Java build.

t0yv0 (approved):

myhau:
Are you sure [this schema](https://github.com/pulumi/pulumi-azure-native/pull/1920/files#diff-545ff4fc09480bc4d71134270a4ab338e66251db350f31defeea5ad12ebbf017) was generated by `pulumi-gen-azure-native` with `pulumi-java` version bumped to `v0.5.x` (not `v0.2.0`)? 

I'm asking, because an issue that I've been investigating (https://github.com/pulumi/pulumi-java/issues/808) seems to be fixed in `pulumi-java` `v0.5.x`, but `schema.json` in this PR still contains invalid Java examples (try  `gron schema.json | grep -i 'resources["azure-native:keyvault:Vault"].description'` and then `cmd+f` for `Map.of`). 

I also checked this locally by building this PR (cmd tools + doc schema) — examples are fine, issue is not there.

stack72:
> @stack72
> 
> Are you sure [the `schema.json` file](https://github.com/pulumi/pulumi-azure-native/pull/1920/files#diff-545ff4fc09480bc4d71134270a4ab338e66251db350f31defeea5ad12ebbf017) was generated by `pulumi-gen-azure-native` with `pulumi-java` version bumped to `v0.5.x` (not `v0.2.0`)?
> 
> I'm asking, because an issue that I've been investigating ([pulumi/pulumi-java#808](https://github.com/pulumi/pulumi-java/issues/808)) seems to be fixed in `pulumi-java` `v0.5.x`, but `schema.json` in this PR still contains invalid Java examples (try `gron schema.json | grep -i 'resources["azure-native:keyvault:Vault"].description'` and then `cmd+f` for `Map.of`).
> 
> I also checked this locally by building this PR (cmd tools + doc schema) — examples are fine, issue is not there.

Hi @myhau 

The schema is generated using v0.5.0 of pulumi-java dependency - should it be a different version?

t0yv0:
https://github.com/pulumi/pulumi-azure-native/runs/7908286955?check_suite_focus=true#step:18:39

Building java in 6s -too good to be true? Something is not building.

stack72:
> https://github.com/pulumi/pulumi-azure-native/runs/7908286955?check_suite_focus=true#step:18:39
> 
> Building java in 6s -too good to be true? Something is not building.

I can pass verbose flags to gradle if you want - this is the same behaviour I get locally 

the build is super fast for me but the generation is slower

stack72:
(But not 6 seconds!)

t0yv0:
I've tried on my machine. After `cd sdk/java` I run `gradle build` and it's running for 1m53s now and still going. With Java compiler hogging CPU. That's what I'd expect.

However when I run the `make build_java` it exists immediately.

Something is fishy.

t0yv0:
OK so `--version` argument makes Gradle print a version and exit, instead of building! That's the problem I think

t0yv0:
Perhaps

```
build_java:: $(WORKING_DIR)/sdk/java/go.mod .version
	cd ${PACKDIR}/java/ && \
	 	gradle --console=plain -Pversion=$(shell cat .version) build

```

t0yv0 (changes requested):
Would love to see Java builds actually building, see above. TYVM.

Makefile
```diff
@@ -172,6 +177,19 @@ build_dotnet::
...
 
+$(WORKING_DIR)/sdk/java/go.mod:
+	echo "$$FAKE_MODULE" | sed 's/fake_module/fake_java_module/g' > $@
+
+generate_java:: bin/pulumi-java-gen generate_schema
+	$(WORKING_DIR)/bin/$(JAVA_GEN) generate --schema $(WORKING_DIR)/provider/cmd/$(PROVIDER)/schema.json --out sdk/java --build gradle-nexus
+
+build_java:: $(WORKING_DIR)/sdk/java/go.mod .version
+	cd ${PACKDIR}/java/ && \
+		gradle --console=plain build --version $(shell cat .version)
```
t0yv0, line 188: Still passing --version incorrectly here, this will make `gradle` report gradle version and quit.

t0yv0 (approved):
Looks better thank you.

PR #1926, opened 2022-08-19
Obtain auth tokens via the new azidentity/MSAL method

thomas11:
This PR replaces the default implementation of `getOAuthToken` with one that uses the new azidentity package, part of azure-sdk-for-go, to obtain an Azure auth token. This is also known as MSAL, whereas the previous, deprecated method is ADAL. 

Multitenant OAuth [is not implemented yet in azidentity](https://github.com/Azure/azure-sdk-for-go/issues/17159), that's the only feature gap I'm aware of. For that, we'll continue to fall back on ADAL.

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -191,6 +191,10 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 		return nil, errors.Wrap(err, "building authorizer")
 	}
 
+	if k.getConfig("useLegacyADALAuth", "USE_LEGACY_ADAL_AUTH") == "true" {
+		k.host.Log(ctx, diag.Warning, "", "USE_LEGACY_ADAL_AUTH is deprecated and will be removed in a future release.")
+	}
+
 	k.subscriptionID = authConfig.SubscriptionID
 	k.client.Authorizer = tokenAuth
 	k.client.UserAgent = k.getUserAgent()
@@ -414,10 +418,6 @@ func (k *azureNativeProvider) Check(ctx context.Context, req *rpc.CheckRequest) 
 				Reason: fmt.Sprintf("missing required property '%s'", name),
 			})
 		}
-	}
-
-	if k.getConfig("useLegacyADALAuth", "USE_LEGACY_ADAL_AUTH") == "true" {
-		k.host.Log(ctx, diag.Warning, urn, "USE_LEGACY_ADAL_AUTH is deprecated and will be removed in a future release.")
 	}
 
 	resInputs, err := plugin.MarshalProperties(news, plugin.MarshalOptions{


examples/go-auth/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+func main() {
+	pulumi.Run(func(ctx *pulumi.Context) error {
+		tokenResult, err := authorization.GetClientToken(ctx, &authorization.GetClientTokenArgs{}, nil)
+		if err != nil {
+			return err
+		}
+
+		log.Println("Got OAuth token: ", tokenResult.Token[:20])
+		q.Q(tokenResult.Token)
```
viveklak, line 19: We typically don't use q outside of purely debug scenarios (i.e. don't check this in.)

provider/pkg/provider/provider.go
```diff
@@ -192,6 +197,18 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 	}, nil
 }
 
+// Translate between the autorest and the azure-sdk-for-go representation of Azure clouds.
+func (k *azureNativeProvider) cloud() cloud.Configuration {
+	thisCloud := cloud.AzurePublic
+	if k.environment.Name == azure.USGovernmentCloud.Name {
```
viveklak, line 203: Nit - would switch/case be cleaner?

provider/pkg/provider/provider.go
```diff
@@ -233,7 +250,7 @@ func (k *azureNativeProvider) Invoke(ctx context.Context, req *rpc.InvokeRequest
 		if endpointArg := args["endpoint"]; endpointArg.HasValue() && endpointArg.IsString() {
 			endpoint = endpointArg.StringValue()
 		}
-		token, err := k.getOAuthToken(ctx, auth, endpoint)
+		token, err := k.getOAuthTokenNew(ctx, auth, endpoint)
```
viveklak, line 253: If we have some trepidation on support, we can add the new API behind an experimental provider option - we do this often in other providers, e.g. kubernetes: https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/provider/provider.go#L436

viveklak:
> The main reason this PR is a draft that of the multiple authentication options I have only tested Azure CLI so far.

This looks great! Pardon my ignorance but what auth options are currently possible with the old ADAL? I think if we get parity there and can get test coverage for the scenarios, we should feel fairly confident about pushing this through.

provider/pkg/provider/provider.go
thomas11: That's the kind of pointer I was hoping for :) Is there any precedent for making instead the previous version opt-in? In this case, customers are already asking for the new version so we don't want to delay too much.

Note that - unless I misunderstand something? - here is no API change here, the change is internal.

examples/go-auth/main.go
thomas11: I added some simple validation of the token string. It's probably overkill since we just received this token from Azure - if it's not valid, it should be their problem...

provider/pkg/provider/provider.go
viveklak: > That's the kind of pointer I was hoping for :) Is there any precedent for making instead the previous version opt-in? In this case, customers are already asking for the new version so we don't want to delay too much.
> 

Yes we have done something similar elsewhere before, i.e. make reverting to old default opt-in.

> Note that - unless I misunderstand something? - here is no API change here, the change is internal.

Yup we are on the same page here. I honestly think we can probably be somewhat aggressive here and just go with it. If users are having issues, they can pin to an older version of the provider as a fallback.

viveklak (approved):

provider/pkg/provider/provider.go
stack72: If the new way of working is backward compatible then let's just push forward

If it's not, we may need to allow people to opt-in

thomas11:
> > The main reason this PR is a draft that of the multiple authentication options I have only tested Azure CLI so far.
> 
> This looks great! Pardon my ignorance but what auth options are currently possible with the old ADAL? I think if we get parity there and can get test coverage for the scenarios, we should feel fairly confident about pushing this through.

ADAL with the hashicorp package tries these auth options:
  1. servicePrincipalClientCertificateAuth,
  2. servicePrincipalClientSecretMultiTenantAuth,
  3. servicePrincipalClientSecretAuth,
  4. managedServiceIdentityAuth,
  5. azureCliTokenAuth

I have replicated this order in my PR, except (2) is not supported yet by azidentity. And I've tested only (5) so far.

danielrbradley (approved):
This looks great - really nice that we can re-use all the same configuration.

Btw, another alternative to using `q` is to attach to the provider and step through - I've just written the [instructions for debugging providers in our team docs](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/debug-provider.md).

thomas11:
> The main reason this PR is a draft that of the multiple authentication options I have only tested Azure CLI so far.

I have now tested service principal with cert and service principal with client secret. MSI (managed identity) is only available when running in certain Azure environments, but I have verified that setting ARM_USE_MSI attempts to use it as expected.

thomas11:
Implemented by #2320

Issue #1930, opened 2022-08-23
Split Go SDK to import per-module


danielrbradley:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

We've hit the file size limit of Go's packages a few times. The long term solution would be to ship the SDK for Go in smaller parts.

This would also allow us to re-introduce inline documentation for Go as well as enabling some other code features which have currently been disabled.

Splitting SDKs is not something currently supported by the core codegen so would have to be implemented especially for azure-native.

t0yv0:
> Splitting SDKs is not something currently supported by the core codegen 

It's probably generally applicable though right? Should we consider a feature req for core codegen here?

danielrbradley:
Yup, would be nice to back-port this - we'll just be doing a tactical split for now. Would also be good to then consider how to do the same with the other languages too.

t0yv0:
SGTM will be curious. Java needs this for sure to support azure-native full schema compilation.

danielrbradley:
Completed by #1934

PR #1934, opened 2022-08-24
Add split Go SDK

danielrbradley:
1. Generate secondary go SDK into `sdk/pulumi-azure-native-sdk` to match new [pulumi-azure-native-sdk repository](https://github.com/pulumi/pulumi-azure-native-sdk)
2. Insert root and leaf `go.mod` files as part of codegen (based on directories with `init.go` files).
3. Include replace directives so the modules can be used locally.
4. Incude current version number as part of go.mod references
    - Fall back to `latest` if not available.
    - Strip build number from codegen version to make compatible with go module versioning.
 5. Build each go module using find.
 6. Add `go_prepublish` target to strip `go.sum` files and `go.mod` replace directives
 7. Adapt GitHub workflows to generate and persist new go SDK between steps
 8. During publish, checkout new repository in sub-folder then tag and push

Outstanding issues:
- [Generated version utility assumes the SDK is in the same repository](https://github.com/pulumi/pulumi-azure-native-sdk/blob/v1.76.0-alpha.24984ad3/pulumiUtilities.go#L70) so won't get the exact correct version.

# Original investigation notes

Investigation into solving #1930 as a proof of concept and to decide a route forward.

Either approach below would be best combined with committing the new go SDK into a new repo such as `pulumi-azure-sdk` as ~200 tags will need to be added on each release, and it adds significantly more code.

## Add go.mod files to the original structure
Pros: The core generation is almost identical, just with extra files appended

Have to reference the root sdk module of the same version. This would require publishing the root SDK first with the new version number, then pulling in the new version number into the per-module go.mod then publishing each after.

We really need the provider to be included in each sub-module directly to remove any internal dependencies between modules that have to be published together.

## Split the schema and then generate independently

This results in every module having its own provider class (and config). This doesn't affect when using the default provider, but would look a little more messy with explicit providers.

Split by modifying discovery glob - makes discovery really quick for single modules.

Go package name is calculated from end of importBasePath in go language config.

`authorization`, `storate` and `synapse` seem to be mixins - not from specs. Could add options to skip adding these, have just filtered them out by token for now.

Generate the go.mod and add to list of files to emit.

### Optional: Remove Double Nesting

Go module is at something like `github.com/pulumi/pulumi-azure-native-sdk/compute`. Codegen wants to create a second `compute` folder within that, but that's not great to have to import `github.com/pulumi/pulumi-azure-native-sdk/compute/compute`.

How do we avoid the double nesting of module name in Go when there's only one module? Could manually change the paths for now. Should we just live with double nesting for now?

There's separate init files in the root and module folders. The root file has construct provider and registers resource package. The module file has the resource "construct" method and registers resource module. Seems like we can just rename the module init file on moving, then fix the import used for the PkgVersion.

## Opportunities to improve core codegen

- Add option to generate go.mod
- Allow option for a "default" module to be pulled to the top-level rather than every module having its own folder.

## Out of scope

- Integration of the new Go SDK for ProgramTest - could continue to generate the monolithic SDK temporarily for testing.
- Publishing workflow - would need to add steps to commit to a separate repo and add relevant tags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,9 +201,13 @@ build_go::
 	find sdk/pulumi-azure-native-sdk -type d -maxdepth 1 -exec sh -c "cd \"{}\" && go mod tidy && go build" \;
 
 prepublish_go:
-	@# must use `sed -i -e` to be portable - but leaves go.mod-e behind on macos
+	@# Remove go module replacements which are added for local testing
+	@# Note: must use `sed -i -e` to be portable - but leaves go.mod-e behind on macos
 	find sdk/pulumi-azure-native-sdk -maxdepth 2 -type f -name go.mod -exec sed -i -e '/replace github\.com\/pulumi\/pulumi-azure-native-sdk /d' {} \;
+	@# Remove sed backup files if using older sed versions
 	find sdk/pulumi-azure-native-sdk -maxdepth 2 -type f -name go.mod-e -delete
+	@# Delete go.sum files as these are not used at the point of publishing.
+	@# This is because we depend on the root package which will come from the same release commit, that doesn't yet exist.
 	find sdk/pulumi-azure-native-sdk -maxdepth 2 -type f -name go.sum -delete
 	cp README.md LICENSE sdk/pulumi-azure-native-sdk/
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,3 +109,9 @@ The versioner tool has been added to break apart the process of reporting and co
 Supporting every version of every API causes the SDK size to be very large. One method we've used to reduce the required size is to only have a single set of types for each API which are re-used across all versions. However, not all the types are compatible and therefore a type with the same name might be incompatible in the next version. This incompatibility is currently detected but ignored on a case-by-case basis for now.
 
 Ideally, at the point a new incompatible version of the type is added, it should then be created with a unique, stable name. Alternatively, we could create a union of the two possible types
+
+## New Go SDK
+
+As the size of the Go SDK is close to exceeding the limit of 512Mb, we've created a new SDK which we're publishing in parallel which defined a Go module per Azure namespace rather than a single root Go module. The additional go modules are auto-generated with the required dependencies in [provider/cmd/pulumi-gen-azure-native/main.go](./provider/cmd/pulumi-gen-azure-native/main.go#L312).
+
+This new SDK is published to its own repository at [github.com/pulumi/pulumi-azure-native-sdk](https://github.com/pulumi/pulumi-azure-native-sdk). This is separate partly due to the large number of tags which are created in this new repository per release, but also to remove the need to commit the SDK code into this repository.


Frassle:
> Add option to generate go.mod

I think we need that in platform for some things anyway. We want to generate some types into the core sdk, but just the code files none of the project related stuff (i.e. no .csproj, go.mod, etc). I think that option would work for you here as well.

provider/cmd/pulumi-gen-azure-native/main.go
```diff
@@ -271,10 +276,85 @@ func emitPackage(pkgSpec *schema.PackageSpec, language, outDir string) error {
 		return errors.Wrapf(err, "generating %s package", language)
 	}
 
+	// go-split ... add go.mods
```
danielrbradley, line 279: Remove this

provider/cmd/pulumi-gen-azure-native/main.go
```diff
@@ -271,10 +276,85 @@ func emitPackage(pkgSpec *schema.PackageSpec, language, outDir string) error {
...
+	files["pulumi-azure-native-sdk/go.mod"] = []byte(`
+module github.com/pulumi/pulumi-azure-native-sdk
+
+go 1.17
+
+require (
+	github.com/blang/semver v3.5.1+incompatible
+	github.com/pkg/errors v0.9.1
+	github.com/pulumi/pulumi/sdk/v3 v3.37.2
```
danielrbradley, line 312: Parameterise versions

provider/cmd/pulumi-gen-azure-native/main.go
```diff
@@ -271,10 +276,85 @@ func emitPackage(pkgSpec *schema.PackageSpec, language, outDir string) error {
...
+		version = "v" + buildVersion.String()
+	} else {
+		version = "latest"
+	}
 	for f, contents := range files {
 		if err := emitFile(outDir, f, contents); err != nil {
 			return errors.Wrapf(err, "emitting file %v", f)
 		}
+
+		if matches := re.FindStringSubmatch(f); len(matches) > 0 {
```
danielrbradley, line 335: Might be able to use https://pkg.go.dev/path/filepath

Makefile
```diff
@@ -173,13 +173,18 @@ build_dotnet::
...
 generate_go:: .version
-	$(WORKING_DIR)/bin/$(CODEGEN) go $(shell cat .version)
+	$(WORKING_DIR)/bin/$(CODEGEN) go,go-split $(shell cat .version)
 	cd ${PACKDIR}/go/ && find . -type f -exec sed -i '' -e '/^\/\/.*/g' {} \;
 
 build_go::
 	# Only building the top level packages and building 1 package at a time to avoid OOMing
 	cd sdk/ && \
 	GOGC=50 go list github.com/pulumi/pulumi-azure-native/sdk/go/azure/... | grep -v "latest\|\/v.*"$ | xargs -L 1 go build
+	find sdk/pulumi-azure-native-sdk -type d -maxdepth 1 -exec sh -c "cd \"{}\" && go mod tidy && go build" \;
```
danielrbradley, line 183: Need to address missing default versions

.github/workflows/build.yml
```diff
@@ -298,6 +307,15 @@ jobs:
...
         github.workspace}}/sdk/${{ matrix.language}}
+    - name: Tar split Go SDK folder
+      if: ${{ matrix.language == 'go' }}
+      run: tar -zcf sdk/pulumi-azure-native-sdk.tar.gz -C sdk/pulumi-azure-native-sdk .
+    - name: Upload split Go artifacts
+      uses: actions/upload-artifact@v2
+      if: ${{ matrix.language == 'go' }}
+      with:
+        name: pulumi-azure-native-sdk.tar.gz
+        path: ${{ github.workspace}}/sdk/pulumi-azure-native-sdk.tar.gz
```
danielrbradley, line 318: This should be untar

.github/workflows/build.yml
```diff
@@ -450,6 +468,14 @@ jobs:
     - name: Uncompress nodejs SDK
       run: tar -zxf ${{github.workspace}}/sdk/nodejs.tar.gz -C
         ${{github.workspace}}/sdk/nodejs
+    - name: Download go SDK
+      uses: actions/download-artifact@v2
+      with:
+        name: nodejs-sdk.tar.gz
+        path: ${{ github.workspace}}/sdk/
```
danielrbradley, line 475: Wrong asset

.github/workflows/release.yml
```diff
@@ -448,12 +470,30 @@ jobs:
...
       run: tar -zxf ${{github.workspace}}/sdk/nodejs.tar.gz -C
         ${{github.workspace}}/sdk/nodejs
+    - name: Download go SDK
+      uses: actions/download-artifact@v2
+      with:
+        name: nodejs-sdk.tar.gz
+        path: ${{ github.workspace}}/sdk/
+    - name: Uncompress go SDK
+      run: tar -zxf ${{github.workspace}}/sdk/nodejs.tar.gz -C
+        ${{github.workspace}}/sdk/nodejs
```
danielrbradley, line 480: Wrong SDK

danielrbradley (commented):
Pair review with @thomas11 

Document & changelog

Makefile
danielrbradley: Resolved via #1949

viveklak (commented):
Should/can we add an acceptance test for this?

provider/cmd/pulumi-gen-azure-native/main.go
```diff
@@ -249,6 +255,95 @@ func emitPackage(pkgSpec *schema.PackageSpec, language, outDir string) error {
...
+module github.com/pulumi/pulumi-azure-native-sdk
+{{ else }}
+module github.com/pulumi/pulumi-azure-native-sdk/{{ .SubmoduleName }}
+{{ end }}
+
+go 1.17
+
+require (
+	github.com/blang/semver v3.5.1+incompatible
+	github.com/pkg/errors v0.9.1
```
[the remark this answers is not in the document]
danielrbradley: No idea on what we actually use. This is taken from our currently manually committed go.mod in the `sdk` folder which we currently manually maintain. I can add something to the contributing docs around updating this. Long term, I think codegen should be generating these dependencies - so they can be upgraded based on what the generated code will need - similar in approach to [this issue around Python dependencies](https://github.com/pulumi/pulumi/issues/10565)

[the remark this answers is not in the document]
danielrbradley: Just checked and apparently we do actually use that package in the generated SDK.

danielrbradley:
> Should/can we add an acceptance test for this?

So, not entirely straightforward to figure out, but it seems like it should work!

danielrbradley:
@viveklak doesn't look like program tests are likely to work at the time. Worth just doing a quiet release alongside the old SDK so we can continue testing against an actual real release?

viveklak (approved):

Issue #1566, opened 2022-03-11
Migrate away from use of go-autorest for authentication


viveklak:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
As per https://github.com/Azure/go-autorest/issues/636 - with the push to migrate from ADAL to MSAL, the use of go-autorest for authentication is starting to cause issues. See https://github.com/pulumi/pulumi-azure-native/discussions/1565 for instance. It appears there is no plan to add MSAL authentication support to go-autorest and the proposal is to use the azidentity package in the AZ go sdk: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity. We should migrate to this model for the azure-native provider.

### Affected area/feature
* getClientConfig
* Authentication

stack72:
So @danielrbradley and I just had a small conversation about this and after some small thoughts we believe:

* We should support BOTH authentication styles for an undetermined period of time to allow transition
* the current authentication system will stay the default BUT we should be able to opt-in to the new authentication system
  * `pulumi config set azure-native:auth future true` or a determined config name
* This work will not happen as part of the 2.0 launch but will be the first thing we handle after the 2.0 work
* in 3.0, the `future` authentication will be made the default experience

PR #1942, opened 2022-08-29
Upgrade our authentication library to support MSAL

thomas11:
The latest version of `hashicorp/go-azure-helpers/authentication` supports the new MSAL authentication (#1566).

This PR also addresses #1565 by requiring the user to have a compatible version of the Azure CLI installed and printing an informative message if not. The corresponding documentation change is in pulumi/registry#1420.

Here's a test of the version handling, done by putting a shell script named `az` in my path that just prints some version.

```
➜  az version
{
  "azure-cli": "2.36.0"
}
➜  pulumi up
Previewing update (dev)
...
Diagnostics:
    error: checking az version: found incompatible az version 2.36.0. Please make sure the Azure CLI 2.37 or greater, but less than 3.x, is installed.
 

➜  az version
{
  "azure-cli": "2.3.7"
}
➜  pulumi up 
Previewing update (dev)
...
Diagnostics:
    error: checking az version: found incompatible az version 2.3.7. Please make sure the Azure CLI 2.37 or greater, but less than 3.x, is installed.
 

➜  pulumi up
Previewing update (dev)
...
Diagnostics:
    error: checking az version: found incompatible az version 2.0.0. Please make sure the Azure CLI 2.37 or greater, but less than 3.x, is installed.
 

➜  pulumi up
Previewing update (dev)
...
Diagnostics:
    error: checking az version: found incompatible az version 3.0.0. Please make sure the Azure CLI 2.37 or greater, but less than 3.x, is installed.
 

➜  mv ~/go/bin/az ~/go/bin/fake_az
➜  az version
{
  "azure-cli": "2.39.0",
  "azure-cli-core": "2.39.0",
  "azure-cli-telemetry": "1.0.6",
  "extensions": {}
}
➜  pulumi up                      
Previewing update (dev)
...
Resources:
    + 3 to create
Do you want to perform this update? no

➜  mv /opt/homebrew/bin/az /opt/homebrew/bin/aztemp
➜  az version
zsh: command not found: az
➜  pulumi up                                       
Previewing update (dev)
...
Diagnostics:
    error: checking az version: could not find `az`. Please make sure the Azure CLI 2.37 or greater, but less than 3.x, is installed.

➜  mv /opt/homebrew/bin/aztemp /opt/homebrew/bin/az
➜  az version
{
  "azure-cli": "2.39.0",
  "azure-cli-core": "2.39.0",
  "azure-cli-telemetry": "1.0.6",
  "extensions": {}
}
```


provider/pkg/provider/provider.go
```diff
@@ -211,16 +274,18 @@ func (k *azureNativeProvider) Invoke(ctx context.Context, req *rpc.InvokeRequest
...
 			return nil, fmt.Errorf("getting auth config: %w", err)
 		}
-		objectId := ""
+		var objectId *string
 		if auth.GetAuthenticatedObjectID != nil {
 			objectId, err = auth.GetAuthenticatedObjectID(ctx)
 			if err != nil {
 				return nil, fmt.Errorf("getting authenticated object ID: %w", err)
+			} else {
+				*objectId = ""
```
lblackstone, line 283: Doesn't this always set the value back to `""` on the happy path?

lblackstone (commented):
Changes LGTM overall. Is this already tested as part of CI?

provider/pkg/provider/provider.go
```diff
@@ -155,6 +161,11 @@ func (p *azureNativeProvider) Attach(context context.Context, req *rpc.PluginAtt
 // Configure configures the resource provider with "globals" that control its behavior.
 func (k *azureNativeProvider) Configure(ctx context.Context,
 	req *rpc.ConfigureRequest) (*rpc.ConfigureResponse, error) {
+
+	if err := assertAzVersion(); err != nil {
```
stack72, line 165: I worry about this - one of the supported ways of working is to set Env Vars - not using the CLI. I think this will break that way of working?

provider/pkg/provider/provider.go
thomas11: Oops, that's silly, will fix

provider/pkg/provider/provider.go
thomas11: Ooh good point! I tested the env vars method, but it only worked because `az` was installed, even if it wasn't used. I'll refactor this to only check the `az` version when we know we need it.

provider/pkg/provider/provider.go
```diff
@@ -176,7 +183,10 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 	}
 	k.environment = env
 
-	tokenAuth, bearerAuth, err := k.getAuthorizers(authConfig)
+	// The ctx Context given by gRPC is request-scoped and will be canceled after this request. We
```
thomas11, line 186: This is an interesting discovery I made by stepping through in the debugger. [This SO post](https://stackoverflow.com/questions/52829481/can-the-content-of-the-context-be-transmitted-in-grpc) seems to confirm it as well.

If true, this means that `k.setLoggingContext(ctx)` above is also incorrect. I verified in the debugger that the logging context is in canceled state after the first request. It simply doesn't seem to be used, however.

viveklak: > This is an interesting discovery I made by stepping through in the debugger. [This SO post](https://stackoverflow.com/questions/52829481/can-the-content-of-the-context-be-transmitted-in-grpc) seems to confirm it as well.
> 
> If true, this means that `k.setLoggingContext(ctx)` above is also incorrect. I verified in the debugger that the logging context is in canceled state after the first request. It simply doesn't seem to be used, however.

The request scoping of the `context` is not surprising - since we have had efforts in place to use that as a means for implementing cancellation in some providers. The scope of the ctx is subtle though as you pointed out and easily forgotten. The logging context is mostly used as a mechanism to seed logging values and not for cascading cancellations as other contexts - and perhaps its use for purely logging is still reasonable. But we should definitely make your discovery above abundantly clear as it pertains to the state of the contexts.

provider/pkg/provider/provider.go
```diff
@@ -211,7 +221,8 @@ func (k *azureNativeProvider) Invoke(ctx context.Context, req *rpc.InvokeRequest
 		if err != nil {
 			return nil, fmt.Errorf("getting auth config: %w", err)
 		}
-		objectId := ""
+		var objectId *string
+		*objectId = ""
```
viveklak, line 225: Isn't this guaranteed to panic?

lblackstone: Vivek is correct. Line 224 will create the pointer type with a default value of `nil`, which causes a panic when you try to assign it.

https://go.dev/play/p/XjBegZNhgYw

provider/pkg/provider/provider.go
```diff
@@ -1680,18 +1691,31 @@ func (k *azureNativeProvider) getAuthConfig() (*authentication.Config, error) {
...
+
+	useMsi := k.getConfig("useMsi", "ARM_USE_MSI") == "true"
+	clientSecret := k.getConfig("clientSecret", "ARM_CLIENT_SECRET")
+	clientCertPath := k.getConfig("clientCertificatePath", "ARM_CLIENT_CERTIFICATE_PATH")
+	// Without either of those, we need the `az` CLI to authenticate. Check that we have a good
+	// version (#1565). The check needs to happen before builder.Build(), or we return the less
+	// fitting error message from go-azure-helpers.
+	if !useMsi && clientSecret == "" && clientCertPath == "" {
+		err := assertAzVersion()
+		if err != nil {
```
viveklak, line 1707: Nit
```suggestion
		if err := assertAzVersion(); err != nil {
```

provider/pkg/provider/provider.go
```diff
@@ -1709,31 +1733,90 @@ func (k *azureNativeProvider) getAuthConfig() (*authentication.Config, error) {
...
+	var stderr bytes.Buffer
+	var stdout bytes.Buffer
+
+	cmd := exec.Command("az", arg...)
+
+	cmd.Stderr = &stderr
+	cmd.Stdout = &stdout
+
+	if err := cmd.Run(); err != nil {
+		err := fmt.Errorf("running az: %+v", err)
```
viveklak, line 1774:
```suggestion
		err := fmt.Errorf("running az: %w", err)
```
Were you trying to dump the stack there? In general better to wrap the error instead I find and let the caller decide what to do with the error.

provider/pkg/provider/provider.go
```diff
@@ -1709,31 +1733,90 @@ func (k *azureNativeProvider) getAuthConfig() (*authentication.Config, error) {
...
+	}
+
+	if actual.LessThan(minAzVersion) || actual.GreaterThanOrEqual(nextMajorAzVersion) {
+		return fmt.Errorf("found incompatible az version %s. %s", actual, versionHint)
+	}
+
+	return nil
+}
+
+func jsonUnmarshalAzCmd(i interface{}, arg ...string) error {
```
guineveresaenger, line 1764: Is there a specific reason we're using json? Asking because I've not seen this pattern to parse terminal commands and am curious. :)

provider/pkg/provider/provider.go
```diff
@@ -1709,31 +1733,90 @@ func (k *azureNativeProvider) getAuthConfig() (*authentication.Config, error) {
...
+
+	_, err := exec.LookPath("az")
+	if err != nil {
+		return fmt.Errorf("could not find `az`. %s", versionHint)
+	}
+
+	var azVersion struct {
+		Cli string `json:"azure-cli"`
+	}
+	err = jsonUnmarshalAzCmd(&azVersion, "version")
```
viveklak, line 1747: This seems to be the only call to jsonUnmarshalAzCmd. Should we instead move azVersion etc. into a specialized `getAzVersion` function instead? A few things about this function seem a little atypical:
1. Target struct to serialize to taken as first argument. Understand this is because of the variadic args
2. Assumption that arg is a pointer etc.

If you don't have obvious immediate uses for this I'd recommend the specialized version instead.

provider/pkg/provider/provider.go
```diff
@@ -1709,31 +1733,90 @@ func (k *azureNativeProvider) getAuthConfig() (*authentication.Config, error) {
...
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("could not determine az version. %s", versionHint))
```
viveklak, line 1749: Use `fmt.Errorf` with `%w` instead? Generally we should be preferring/replacing the wrap usage with the fmt.Errorf variant if possible.

provider/pkg/provider/provider.go
viveklak, line 1764: could you use a more descriptive argument name? e.g. `target`?

guineveresaenger:
I hooked this up to a local build for auth suggestions and it turns out my `az` CLI wasn't up to snuff, so the version verifier worked for me:

Screen Shot 2022-09-07 at 4 30 42 PM
Screen Shot 2022-09-07 at 4 31 35 PM

provider/pkg/provider/provider.go
viveklak: I believe its because `az version` actually dumps json 🤣 : 
```
az version
{
  "azure-cli": "2.39.0",
  "azure-cli-core": "2.39.0",
  "azure-cli-telemetry": "1.0.6",
  "extensions": {}
}
```

provider/pkg/provider/provider.go
thomas11: Oops, that was silly. I was under the impression that `getClientConfig` was part of the regular `pulumi up` code path and therefore was tested. Fixed, will add a test later.

provider/pkg/provider/provider.go
```diff
@@ -1709,31 +1733,90 @@ func (k *azureNativeProvider) getAuthConfig() (*authentication.Config, error) {
...
+	}
+
+	actual, err := goversion.NewVersion(azVersion.Cli)
+	if err != nil {
+		return fmt.Errorf("could not parse az version %q: %+v. %s", azVersion.Cli, err, versionHint)
+	}
+
+	if actual.LessThan(minAzVersion) || actual.GreaterThanOrEqual(nextMajorAzVersion) {
+		return fmt.Errorf("found incompatible az version %s. %s", actual, versionHint)
```
mikhailshilkov, line 1758: Will this break users on older CLI versions that currently work? Is this a breaking change that we can ship? Could we fall back to the old auth mechanism for older CLIs?

provider/pkg/provider/provider.go
danielrbradley: The loggingContext is also interesting because we can send multiple grpc commands in paralell at which point the loggingContext on the global provider object will be overwritten (and might even be able to cause race conditions). Would be interested to reason about this more deeply with you @thomas11

provider/pkg/provider/provider.go
```diff
@@ -1709,31 +1733,90 @@ func (k *azureNativeProvider) getAuthConfig() (*authentication.Config, error) {
 	return builder.Build()
 }
 
-func (k *azureNativeProvider) getAuthorizers(authConfig *authentication.Config) (tokenAuth autorest.Authorizer,
+func assertAzVersion() error {
+	const versionHint = "Please make sure the Azure CLI 2.37 or greater, but less than 3.x, is installed or configure another authentication method."
```
danielrbradley, line 1737: Might be helpful to have a concrete resource to point them at on _how_ to configure another authentication method. This URL is one possible option, though a little long.
```suggestion
	const versionHint = "Please make sure the Azure CLI 2.37 or greater, but less than 3.x, is installed or configure another authentication method. See https://www.pulumi.com/registry/packages/azure-native/installation-configuration/#credentials for more information"
```

provider/pkg/provider/provider.go
```diff
@@ -1709,31 +1733,90 @@ func (k *azureNativeProvider) getAuthConfig() (*authentication.Config, error) {
...
+	if err := cmd.Run(); err != nil {
+		err := fmt.Errorf("running az: %+v", err)
+		if stdErrStr := stderr.String(); stdErrStr != "" {
+			err = fmt.Errorf("%s: %s", err, strings.TrimSpace(stdErrStr))
+		}
+		return err
+	}
+
+	if err := json.Unmarshal(stdout.Bytes(), i); err != nil {
+		return fmt.Errorf("unmarshaling the result of Azure CLI: %v", err)
```
danielrbradley, line 1782: Perhaps include the actual CLI output string in the error if we can't parse it so we can then see what we might need to tweak

danielrbradley (commented):
I think this is basically ready to merge. I've left a couple of suggestions around messages back to the user if something goes wrong:
1. Aim to point them to where to help themselves.
2. Include information so we could reason about exactly why failures happened based on their local system.

Having said that, these are all things that could be added later, so let's not delay too much as this will already be a huge step forward for many users!

provider/pkg/provider/provider.go
danielrbradley: Don't know if there's a fallback option, but if the version is outside the expected range perhaps we could print a big warning but attempt to carry on?

danielrbradley (approved):

provider/pkg/provider/provider.go
thomas11: Chatting with Daniel 1:1 we opted to avoid further changes in this PR since logging is working.

provider/pkg/provider/provider.go
thomas11: CLI version 2.34 from March 1 2022 broke our users, like in #1565. Version 2.37 from May 24 2022 works again. The hashicorp/go-azure-helpers library only requires that the version is at least 2.0.81. So you're right @mikhailshilkov, we can actually allow versions between 2.0.81 and 2.33 and only exclude 2.34-2.36.

provider/pkg/provider/provider.go
thomas11: The target struct as argument is not that unusual IMO when a low-level function returns different kinds of JSON. The caller knows what shape to expect. Let's leave it for now so we can get this merged.

thomas11:
Implemented by #2320

PR #1944, opened 2022-08-30
Exploring lazy load modules

t0yv0:
This builds upon @danielrbradley prototype for lazy modules prototype. It appears that the approach in the prototype is technically a breaking change for certain TypeScript programs as illustrated by the example:

To repro:

```
make build_nodejs
make install_nodejs
test.lazy-imports-test
```

Note the example compiles on stock Pulumi.

```
index.ts:10:57 - error TS2694: Namespace '"/Users/anton/pulumi-azure-native/sdk/nodejs/index"' has no exported member 'storage'.

10 export function test1(resourceGroupName: string): azure.storage.StorageAccount {
                                                           ~~~~~~~

index.ts:11:23 - error TS2694: Namespace '"/Users/anton/pulumi-azure-native/sdk/nodejs/index"' has no exported member 'storage'.

11     const args: azure.storage.StorageAccountArgs = {
                         ~~~~~~~

index.ts:26:51 - error TS2503: Cannot find namespace 'storage2'.

26 export function test2(resourceGroupName: string): storage2.StorageAccount {
                                                     ~~~~~~~~

index.ts:27:17 - error TS2503: Cannot find namespace 'storage2'.

27     const args: storage2.StorageAccountArgs = {
                   ~~~~~~~~


Found 4 errors.
```

CC @RobbieMcKinstry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,7 +172,7 @@ build_dotnet::
 
 generate_go::
 	$(WORKING_DIR)/bin/$(CODEGEN) go ${VERSION}
-	
+
 build_go::
 	# Only building the top level packages and building 1 package at a time to avoid OOMing
 	cd sdk/ && \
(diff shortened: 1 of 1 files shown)


t0yv0:
Going with a different approach.

PR #1955, opened 2022-09-06
Upgrade Pulumi Go dependencies

thomas11:
Fixes dependabot gogo/protobuf security issues.

Upgraded via

```
go get -u github.com/pulumi/pulumi-azure-native/sdk@latest   
go mod tidy
```


examples/go-servicebus-recreate/step1/go.mod
```diff
@@ -3,6 +3,6 @@ module go-servicebus-recreate
 go 1.14
 
 require (
-	github.com/pulumi/pulumi-azure-native/sdk v1.35.0
-	github.com/pulumi/pulumi/sdk/v3 v3.13.2
+	github.com/pulumi/pulumi-azure-native/sdk v1.74.0
```
viveklak, line 6: I actually wonder if we even need the go.mod files here? Reading the integration testing framework code https://github.com/pulumi/pulumi/blob/490be1cc52673b7afe26cc4365e199005ed65d4c/pkg/testing/integration/program.go#L2071 it seems like we should be able to just use the "latest" as part of the `go mod init` then have the azure-native/sdk dependency be replaced with the local version as long as the dependency is provided - which it seems to do by default: https://github.com/pulumi/pulumi-azure-native/blob/master/examples/examples_go_test.go#L52

thomas11: I'm not sure if I understand "replaced with the local version as long as the dependency is provided" correctly. I assume it's the `go-mod-edit` command in your first link that makes the newly initialized go.mod use local dependencies instead of pulling them down? And that makes sure we're using the versions specified in the top-level go.mod?

thomas11: Simply removing `go.mod` and `go.sum` fails the `go mod edit -replace` step with errors like

```stackName-1662569089032762000-727887 imports
	github.com/pulumi/pulumi-azure-native/sdk/go/azure/resources: github.com/pulumi/pulumi-azure-native/sdk/go@v0.0.0-00010101000000-000000000000: replacement directory /Users/tkappler/go/src/github.com/pulumi/pulumi-azure-native/sdk/go does not exist```

The actual replacement dir would be `go/pkg/mod/github.com/pulumi/pulumi-azure-native/sdk@v1.74.0/go`.

thomas11: @viveklak should we get this merged and maybe file a simplification as a future work item?

PR #1967, opened 2022-09-13
Upgrade to 2022-04-01-preview for MobileNetwork

### LINKED ISSUE #1966: Our default version of MobileNetwork has been removed upstream
[Remove unused Microsoft.MobileNetwork 2022-01-01-preview API](https://github.com/Azure/azure-rest-api-specs/pull/20641#top)

This is currently set as our default version. Therefore, we should change our default version otherwise the default version will be missing. Flagged by the nightly upgrade highlighting breaking changes.

danielrbradley:
Fixes #1966


stack72 (approved):
We should note this in the changelog as this could be breaking for users

danielrbradley:
Yes, will changelog, but unlikely to have an impact because, according to Azure, this was removed because no-one was using it.

PR #1971, opened 2022-09-14
Improve local build process

danielrbradley:
- Model dependencies in make so the right things always build.
- Write all targets as file-based so we get up-to-date builds (including sentinel files).
- Download correct versions pulumictl and javagen binaries for this project (don't use global installs & lock versions).
- Rework gen program so we don't have to re-parse specs for codegen-only runs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,8 @@ VERSION_FLAGS   = -ldflags "-X github.com/pulumi/pulumi-azure-native/provider/pk
 
 default: init_submodules provider arm2pulumi local_generate
 
+ensure: .init_submodules.sentinel bin/pulumictl provider/.mod_download.sentinel
+init_submodules: .init_submodules.sentinel
 arm2pulumi: bin/arm2pulumi
 codegen: bin/$(CODEGEN)
 provider: bin/$(PROVIDER)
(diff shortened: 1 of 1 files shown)


stack72 (approved):

Makefile
```diff
@@ -1,92 +1,168 @@
...
 
 PROVIDER_PKGS   := $(shell cd ./provider && go list ./pkg/...)
 WORKING_DIR     := $(shell pwd)
 
-GO_SRC          := $(wildcard provider/go.*) $(wildcard provider/*/*/*.go)
+# azure-quickstart-templates is generated by arm2pulumi_coverage_report
+PROVIDER_PKG	= $(shell find provider/pkg -type f -not -path "*/azure-quickstart-templates/*")
+SPECS           = $(shell find azure-rest-api-specs/specification/*/resource-manager -type f -name "*.json" ! -path "**/examples/**")
 
 JAVA_GEN 		 := pulumi-java-gen
```
thomas11, line 16: [nit] indentation

Makefile
```diff
@@ -1,92 +1,168 @@
...
 JAVA_GEN_VERSION := v0.5.4
```
thomas11, line 17: unused now?

thomas11 (approved):

Makefile
danielrbradley: Ah the joys of tabs vs spaces!

PR #1975, opened 2022-09-16
Improve build process

danielrbradley:
Re-submission of #1971 

- Model dependencies in make so the right things always build.
- Write all targets as file-based so we get up-to-date builds (including sentinel files).
- Download correct versions pulumictl and javagen binaries for this project (don't use global installs & lock versions).
- Rework gen program so we don't have to re-parse specs for codegen-only runs.


danielrbradley:
[Test publish succeeded](https://github.com/pulumi/pulumi-azure-native/actions/runs/3067712794), re-merging

PR #1983, opened 2022-09-21
Fix nightly-sdk-generation

thomas11:
This nightly job started failing a few days ago, probably due to the merge of #1975.

The commits so far fix
  - The "clean" step of the workflow needs to remove more than `make clean` does.
  - `find bin ...` fails on a fresh clone with no `bin`.
  - `pulumictl get version` fails because the Git clone has no tags
     > Error: error calculating version: getting language versions: error determining base versionComponents: mostRecentTag: object not found

Still failing at this point: `local_generate`, with many errors like this:
> duplicate resource Network:CustomIPPrefix from 2021-08-01 and 2022-05-01

Unfortunately, adding the `fetch_depth: 0` option to the git checkout step adds about six minutes. But it's currently the only way to get tags, until the GH maintainers stop ignoring [this](https://github.com/actions/checkout/pull/579).

### DIFF
--- a/.github/workflows/nightly-sdk-generation.yml
+++ b/.github/workflows/nightly-sdk-generation.yml
@@ -37,8 +37,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         lfs: true
-    - name: Fetch tags # `pulumictl get version` needs them 
-      run: git fetch --tags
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):
Spot on!

PR #1988, opened 2022-09-26
Unbreak SDK cleanup of nightly SDK generation

thomas11:
Follow-up to #1983.

This should fix the release process which needs `sdk/go.mod`.


danielrbradley (approved):
Have re-added missing files too

PR #1998, opened 2022-09-28
Continue on error in Publish Java SDK

thomas11:
due to pulumi/pulumi-java#830, suggested [here](https://pulumi.slack.com/archives/C02TBBNKVBR/p1663622272520259)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -406,6 +406,7 @@ jobs:
         $GITHUB_ENV
     - name: Publish Java SDK
       uses: gradle/gradle-build-action@9b814496b50909128c6a52622b416c5ffa04db49
+      continue-on-error: true # https://github.com/pulumi/pulumi-java/issues/830
       with:
         arguments: publishToSonatype closeAndReleaseSonatypeStagingRepository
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):

viveklak:
@thomas11 please make sure you synchronize this against ci-mgmt which already has this: https://github.com/pulumi/ci-mgmt/blob/master/native-provider-ci/providers/azure-native/repo/.github/workflows/release.yml#L494

danielrbradley:
@viveklak I think we're actually quite a long way out of sync due to all the go-split SDK specializations and a few other hacks we've had to put in recently. I'd suggest we pause syncing this down from ci-mgmt until we can have a proper re-review of the situation.

viveklak:
> @viveklak I think we're actually quite a long way out of sync due to all the go-split SDK specializations and a few other hacks we've had to put in recently. I'd suggest we pause syncing this down from ci-mgmt until we can have a proper re-review of the situation.

Gotcha, missed that part. We have disabled automerge for all ci-mgmt PRs at the moment anyway so this should not be a big deal for now. Would be good to get it back on rails soon though.

PR #2000, opened 2022-09-28
Bump Pulumi deps to 3.40.2

thomas11:
This version includes pulumi/pulumi#10705 which fixes issues in our docs generation.


viveklak:
@thomas11 wouldn't you expect to see a diff in the schema.json as a result of this? Perhaps you need to run a `make local_generate`?

thomas11:
Yep, you're right. My mental model of the docs update process was off. Will
push regenerated docs.

danielrbradley (approved):

danielrbradley:
This looks good - FYI there is also a GitHub workflow which should be able generate this kinda PR automatically

PR #2056, opened 2022-10-27
Run ProgramTest acceptance tests on split Go SDKs

thomas11:
This PR adds a new test folder `examples/azure-native-sdk` which contains copies of the existing Go tests under `examples`, but updated to use the new split SDK of #1934.

The new driver `examples_go_sdk_test.go` runs them, passing the correct paths and the correct environment to ProgramTest. All imports of the Go SDK packages are re-targeted to the local `sdk/pulumi-azure-native-sdk` folder.

While I was at it, I also updated the tests to the latest dependencies.

Open design decisions:
  - The new tests still use the same `go` build tag, so one can only run old and new Go tests together. We might want to break this up.
  - The folder structure under `examples/` is now inconsistent, where the new Go tests are in their own folder but the others are in a flat hierarchy. Although, the old Go tests will disappear in the new future, making it less cluttered.

### DIFF
--- a/examples/examples_go_sdk_test.go
+++ b/examples/examples_go_sdk_test.go
@@ -73,12 +73,15 @@ func getGoBaseOptionsSdk(t *testing.T) integration.ProgramTestOptions {
 		t.Fatal(err)
 	}
 
+	rootSdkPath, err := filepath.Abs(sdkPath)
+	require.NoError(t, err)
+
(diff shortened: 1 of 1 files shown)


viveklak (approved):

danielrbradley (approved):
This looks great thank you! Doesn't add too much to the overall test time either 👍

PR #2068, opened 2022-11-04
Make id optional for network SubResource in 20220701

thomas11:
Fixes nightly-sdk-generation failure

```
incompatible type "azure-native:network/v20220701:SubResource" for resource "DnsForwardingRuleset" ("azure-native:network/v20220701:DnsForwardingRuleset"): required properties do not match: only required in A: id
```


provider/pkg/gen/schema.go
```diff
@@ -1538,6 +1538,13 @@ func (m *moduleGenerator) genTypeSpec(propertyName string, schema *spec.Schema,
 				return nil, nil
 			}
 
+			// https://github.com/Azure/azure-rest-api-specs/issues/21431
+			if tok == "azure-native:network/v20220701:SubResource" {
+				log.Printf("Removing required 'id' from %s of %s", propertyName, tok)
+				props.requiredProperties.Delete("id")
```
thomas11, line 1544: Not sure why there's `requiredProperties` and `requiredSpecs`? They have the same content in this case, and only `requiredSpecs` seems to be used downstream.

provider/pkg/gen/schema.go
```diff
@@ -1538,6 +1538,13 @@ func (m *moduleGenerator) genTypeSpec(propertyName string, schema *spec.Schema,
 				return nil, nil
 			}
 
+			// https://github.com/Azure/azure-rest-api-specs/issues/21431
+			if tok == "azure-native:network/v20220701:SubResource" {
```
thomas11, line 1542: Any thoughts on how to make this more future-proof for future API versions, in case they don't fix the underlying issue?

danielrbradley (commented):
I think it'd be worth also generating so we can see the effect of this override:

1. If it allows the generation to succeed.
2. If setting `requiredSpecs` has any undesired effects or is necessary.

Tip: I'd also commit the schema and SDKs separate if it's large so it's easier to see what's happening.

thomas11:
> We need to run and commit the generated files so we can see the effect of this override:
> 
> 1. If it allows the generation to succeed.
> 2. If setting `requiredSpecs` has any undesired effects or is necessary.
> 
> Tip: I'd also commit the schema and SDKs separate if it's large so it's easier to see what's happening.

@danielrbradley I tried it locally and generation succeeds (only with the change). Just pushed the schema and SDK.

danielrbradley:
Does it succeed or fail if you don't add the `requiredSpecs` changes (only leaving the `requiredProperties`)?

thomas11:
> Does it succeed or fail if you don't add the `requiredSpecs` changes (only leaving the `requiredProperties`)?

It fails. Actually, it succeeds if I omit the *`requiredProperties`* change, but i left it in for symmetry.

danielrbradley (approved):
Wel, I'm super-confused by the `requiredProperties` vs `requiredSpecs`. I feel like I've got a hole in my understanding here, but happy we've proved the approach.

PR #2071, opened 2022-11-08
Prepare warnings for Go SDK depracation

danielrbradley:
Add warnings for old Go SDK deprecation

TODO:
- [x] Update links to actual reference
- [x] Disable rebuilding the Go SDK (otherwise the build fails with dirty working tree)

Preview of output:
image


sdk/go/azure/init.go
```diff
@@ -34,4 +31,7 @@ func init() {
 		"azure-native",
 		&pkg{version},
 	)
+	fmt.Println("\033[33m" + "warning:" + "\033[0m" + " module github.com/pulumi/pulumi-azure-native/sdk is deprecated; " +
```
thomas11, line 34: I'm not very familiar with these color codes. Are they cross-platform?

thomas11 (approved):

sdk/go/azure/init.go
danielrbradley: Yup, colors work on Mac, Linux and Windows all fine.

Issue #2080, opened 2022-11-14
Create V2 version selection


danielrbradley:
Summary of sub-tasks:
- [ ] Re-compute a "latest" selection of versions ready for a V2 release
-  [ ] Manually audit selection
   -  [ ] Remove obsolete resources (e.g., after renames)
   -  [ ] Simplify where's there's lots of additional resources (as well as tracking versions)
    - [ ] If we're mainly using additional resources, consider removing "tracking" version
- [ ] Document manual changes and reasons
- [ ] Re-encorporate changes as logic in the config generator (ready for V3)

mikhailshilkov:
Isn't this done yet?

mikhailshilkov:
What's remaining here?

danielrbradley:
Closing as this is all complete and integrated into the v2.0-pre branch

PR #2140, opened 2022-12-18
Beginnings of explicit manual curation for API versioning

thomas11:
Getting this out as a draft PR because vacation is coming up.

This PR consists of two commits that I suggest reviewing separately:
  1. Remove the unused deprecation code
  2. Add a separate file for explicit manual curation

Currently, the curation code supports exclusion of resources. When running `make versions/v2.json`, the resulting `v2.json` does not have the resources excluded in `v2-curation.yaml`.

`v2-curation.yaml` also contains the actual manual curation of the resource providers starting with 'A' and 'B'.

Not implemented yet are the curations that affect the building of `vN-config.yaml`:
  - `preferPreview` (over stable)
  - `explicit` (no tracking version, list all resources)

Part of #2080.

### DIFF
--- a/provider/pkg/versioning/manualCuration_test.go
+++ b/provider/pkg/versioning/manualCuration_test.go
@@ -1,0 +1,22 @@
+package versioning
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsExcluded(t *testing.T) {
+	var curations = make(Curations)
+
+	assert.False(t, curations.IsExcluded("Compute", "someResource"))
+
+	curations["Compute"] = providerCuration{}
+	assert.False(t, curations.IsExcluded("Compute", "someResource"))
+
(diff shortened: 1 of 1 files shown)


danielrbradley:
First thoughts on the WIP:
1. The structure of the curations file looks good.
2. Are you manually finding the name for the `folders` properties? Is there a quick way of looking this up?
3. It looks like the process is something like: Generate `v2-config.yaml` (old code), hand-write `v2-curations.yaml`, generate `v2.json` from `v2-config.yaml` + `v2-curations.yaml`?
4. I wonder if the process should instead be something like: Handwrite `cutations.yaml`, generate `v2-config.yaml` from `cutations.yaml`, generate `v2.json` from `v2-config.yaml`. Making the process a little more linear, and the curations isn't per-version as such - but is used at specific points in time to generate a new vNext config.

danielrbradley (approved):
Yup, let's get this on trunk and pick it up more later.

PR #2143, opened 2022-12-20
Remove web.ContainerApp resource

mikhailshilkov:
web.ContainerApp has been deprecated by Azure a while ago: https://github.com/microsoft/azure-container-apps/issues/109

If I try using them, I get 

```
Creation or update of environments with type managed in Microsoft.Web namespace is no longer supported.
Please use Microsoft.App namespace instead. More information available via aka.ms/microsoftappmigration.
```

Therefore, I believe we should simply remove this resource and avoid confusing it with `app.ContainerApp`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ local_generate_code: generate_nodejs
 local_generate_code: generate_python
 local_generate_code: generate_dotnet
 local_generate_code: generate_go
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
Yup, agree this is a good removal to avoid confusion.

PR #2146, opened 2022-12-20
Fix makefile dependencies & automerge issues

danielrbradley:
- Centralise all sentinel targets in `.make/*` - committed the migration step by step.
- Add missing dependencies.
- Change all committed files to use sentinel files as checkouts will alter timestamps.
- Enable parallel builds.
- Build all SDKs in default target.

Unrelated housekeeping - use NopCloser from `io` instead of `ioutil` to fix warning.

Update workflows to remove language versions in matrix so we can require each matrix step passing to fix the current issue with failing PRs being auto-merged every night.
- Bump node and dotnet to latest LTS version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ local_generate_code: generate_nodejs
 local_generate_code: generate_python
 local_generate_code: generate_dotnet
 local_generate_code: generate_go
-local_generate: local_generate_code generate_docs generate_schema
+local_generate: generate_schema local_generate_code generate_docs
 
 build: init_submodules codegen local_generate provider build_sdks
 # Required for the codegen action that runs in pulumi/pulumi
--- a/provider/pkg/openapi/discover.go
+++ b/provider/pkg/openapi/discover.go
@@ -4,14 +4,16 @@ package openapi
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+
 	"github.com/go-openapi/spec"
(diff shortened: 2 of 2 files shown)


sdk/dotnet/CognitiveServices/V20221201/GetCommitmentPlan.cs
```diff
@@ -27,6 +27,12 @@ public static Output<GetCommitmentPlanResult> Invoke(GetCommitmentPlanInvokeArgs
 
     public sealed class GetCommitmentPlanArgs : global::Pulumi.InvokeArgs
     {
+        /// <summary>
+        /// The name of Cognitive Services account.
+        /// </summary>
+        [Input("accountName", required: true)]
+        public string AccountName { get; set; } = null!;
```
t0yv0, line 34: Looks like there're some small semantic changes in the PR together with make changes, that's ok but ideally those'd be separate :)

t0yv0 (approved):

sdk/dotnet/CognitiveServices/V20221201/GetCommitmentPlan.cs
danielrbradley: Ah yeah - this needs reversing out. This is resolved by this PR: https://github.com/pulumi/pulumi-azure-native/pull/2147

t0yv0: Oh that's from the random map traversal order! Gift that keeps on giving.

Issue #1602, opened 2022-03-25
ApiManagement: ListSubscription fails with status 400


JoschaMetze:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Invoking ListSubscription on the master subscription of an ApiManagement instance fails with error code 400 (Request payload must contain JSON object) .
Invoking the same call via pure REST (or the azure rest page) works just fine with the same URL parameters.

### Steps to reproduce

1. Call `Pulumi.AzureNative.ApiManagement.ListSubscriptionSecrets.Invoke(new Pulumi.AzureNative.ApiManagement.ListSubscriptionSecretsInvokeArgs() { ResourceGroupName = <rg>, ServiceName = <apimname>, Sid = "master" });`
2. pulumi up

Expected: Access the primaryKey of the result
Actual: 
`Grpc.Core.RpcException: Status(StatusCode="Unknown", Detail="invocation of azure-native:apimanagement:listSubscriptionSecrets returned an error: request failed /subscriptions/xxxxxxxxxxx/resourceGroups/yyyyyyyyyy/providers/Microsoft.ApiManagement/service/zzzzzzzzzz/subscriptions/master/listSecrets: autorest/azure: Service returned an error. Status=400 Code="ValidationError" Message="Request payload must contain JSON object."")
       at async Task<InvokeResponse> Pulumi.GrpcMonitor.InvokeAsync(InvokeRequest request)
       at async Task<SerializationResult> Pulumi.Deployment.InvokeRawAsync(string token, SerializationResult argsSerializationResult, InvokeOptions options)
       at async Task<OutputData<T>> Pulumi.Deployment.RawInvoke<T>(string token, InvokeArgs args, InvokeOptions options)`

Cheers
Joscha

PR #2155, opened 2022-12-22
Add a default POST body for invokes to fix 400 Request payload must contain JSON object

mikhailshilkov:
Fix https://github.com/pulumi/pulumi-azure-native/issues/1602

If the body is calculated as empty while making a POST invoke, we currently send a `nil` response body. This PR changes it to an empty map (`{}`) which satisfies the requirement of being a valid JSON.

I _think_ this should be safe to do universally and we already [do this](https://github.com/pulumi/pulumi-azure-native/blob/a9b14aabbbd946ce132efff82f81d5df8f19aec7/provider/pkg/provider/provider.go#L1010) for READ-as-POST. If it's not safe, I hope our tests will catch it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## HEAD (Unreleased)
 
 - Remove a deprecated resource `web.ContainerApp`. Please use `app.ContainerApp` instead.
+- Fix ApiManagement: ListSubscription fails with status 400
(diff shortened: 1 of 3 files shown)


danielrbradley (approved):
Agree, this makes sense to ensure we always send a valid JSON body for a POST.

PR #2163, opened 2023-01-06
Add new RP Voice Services and bump to latest spec

thomas11:


thomas11:
The removal of "createMode" is ok, coming from commit e0aca4c321, stating
> Fix s360 alerts for properties being returned in GET but not exising in PUT requests

danielrbradley (approved):

PR #2180, opened 2023-01-18
Rework build & CI

### LINKED ISSUE #2181: CI: Implement Schema Explosion solution
## Hello!

## Issue details

Currently, the entire pulumi schema for Azure Native is stored locally in a single file.  @danielrbradley has demonstrated a method for partitioning that schema into several much smaller files which can be reasonably viewed in IDEs, reviewed as a part of the PR review process, etc.  We should implement that here, where said schema file is largest so we can actually commit those files into source control (and gain all the performance benefits which come along with doing this).

danielrbradley:
Fixes #2181 

## Significant changes

1. Add missing dependency on `init_submodules` from `version_spec` target.
2. Upgrade to the latest version of pulumictl.
3.  Stop the makefile interacting with git to calculate the version. Make PROVIDER_VERSION an input then convert the format for specific languages.
4. Remove version calculations for acceptance tests
5. Restructure CI to reflect makefile structure.
    - Rrerequisites captures artifacts from calling
    - `make versioner versions codegen schema provider`
    - SDK build & test calls restores binaries then runs 
    - `make --touch versioner versions codegen schema provider`
6. Remove obsolete `metadata.json` and fix all outdated dependencies on it.
7. Stop generating and publishing the legacy Go SDK which has been unusabl. Add a note directing to migration instructions.
8. Generate `schema-full.json` and `metadata-compact.json` centrally to the `bin` directory then copy locally when needed.
9. Extract build & test as a common, reusable workflow which is called for PR, master and release runs.
10. Check out PR merge commits rather than the branch HEAD - so we're testing what _will_ happen when we merge.

## Impact

Total time reduced from **1h 48m** to **35m**.
No duplicated build and test steps repeated between workflows.
Better coverage having removed legacy non-updated files.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -17,19 +17,10 @@ env:
   PROVIDER_VERSION: ${{ inputs.version }}
   PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
   PULUMI_LOCAL_NUGET: ${{ github.workspace }}/nuget
-  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-  NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-  NUGET_PUBLISH_KEY: ${{ secrets.NUGET_PUBLISH_KEY }}
-  PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
   # TRAVIS_OS_NAME required by https://github.com/pulumi/scripts/blob/master/ci/publish-tfgen-package
   TRAVIS_OS_NAME: linux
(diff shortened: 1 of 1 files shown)


thomas11 (approved):

.github/workflows/pull-request.yml
```diff
@@ -1,4 +1,4 @@
-name: run-acceptance-tests
+name: pull-request
```
thomas11, line 1: I think we should allow this workflow to be triggered manually as well. It's useful for testing changes that are not ready for a PR. In that case, the previous name was better.

thomas11 (approved):

.github/workflows/pull-request.yml
danielrbradley: ok, changing back to just `acceptance-test` instead

thomas11 (approved):

PR #2202, opened 2023-01-26
Fix publishing workflows

danielrbradley:
Extract provider publish job to its own reusable workflow which can be optionally included in acceptance testing. This allows us to start testing the publishing process (for prerelease) before merging to master which fixes the need for us to test directly on the master branch.

Added a "short test" option to the acceptance test workflow which allows us to only run a smaller selection of end-to-end tests which reduces the CI run by around 8 minutes.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -56,7 +56,9 @@ jobs:
   publish:
     uses: ./.github/workflows/publish-provider.yml
     secrets: inherit
-    needs: build_test
+    needs:
+      - version
+      - build_test
     with:
       version: ${{ needs.version.outputs.version }}
       ref: ${{ github.ref }}


.github/workflows/build-test.yml
```diff
@@ -112,39 +116,56 @@ jobs:
...
           tag: v2.4.0
           repo: GoTestTools/gotestfmt
+
       - name: Run tests
+        if: "!inputs.short_test"
         run: |
           set -euo pipefail
           cd examples && go test -v -json -count=1 -cover -timeout 2h -tags=${{ matrix.language }} -parallel 4 . 2>&1 | tee /tmp/gotest.log | gotestfmt
+
+      - name: Run tests
```
thomas11, line 143: Updating this name would make it clearer which tests were run when looking at a past workflow

.github/workflows/publish-provider.yml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: Prerequisites artifact restore
+        uses: ./.github/actions/prerequisites-artifact-restore
+
+      - name: Mark prerequisites as up-to-date
+        # Don't include provider as that's the bit we're going to rebuild
+        run: make --touch versioner versions codegen schema
+
+      - name: Ensure provider build prerequisites
+        run: make provider_prebuild
```
thomas11, line 51: This is new. Is it because after extracting publish into its own workflow, we cannot rely on having the prebuilt steps already done on the same VM?

thomas11 (approved):

.github/workflows/publish-provider.yml
danielrbradley: Yup - because when we restore the schema and metadata, we have to copy it back into the `cmd/` folder ready to be used for the embed - which we do within the prebuild.

.github/workflows/build-test.yml
danielrbradley: Oh yeah, that'd make sense .. will update as a follow-up!

PR #2208, opened 2023-01-30
Fix arm2pulumi coverage report

danielrbradley:
Remove calls to `make init_submodules`
- We no longer handle submodule checkout via make.
- We do still have `make ensure` which installs other tooling prerequisites.

Additional tiding:
- Fix remaining uses of ::set-output
- Remove all uses of GO111MODULE=on
- Reformat scripts to remove YAML blank lines
- Remove superfluous env in workflows

### DIFF
--- a/.github/workflows/arm2pulumi-coverage-report.yml
+++ b/.github/workflows/arm2pulumi-coverage-report.yml
@@ -1,32 +1,13 @@
 name: arm2pulumi-coverage-report
+
 on:
   schedule:
     - cron: 35 17 * * *
   workflow_dispatch: {}
+
 env:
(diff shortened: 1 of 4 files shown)


squaremo (approved):
These commits all make sense :+1: Thanks Daniel.
For the last one, on the face of it one might expect the ARM-specific workflows to need ARM_* env entries -- it might be worth mentioning in the commit message that they are really only used for end to end testing (which isn't modified here).

danielrbradley:
Logged https://github.com/pulumi/pulumi-azure-native/issues/2209 as a follow-up issue

PR #2210, opened 2023-01-30
Extract SDK publishing steps into reusable workflow

danielrbradley:
- Split each language into own job to allow us to re-run on failure.
- Remove dependencies on pulumi/scripts for the ci/publish-tfgen-package script which couples all publishing together.
- Adds pushing of a branch for the Go SDK prerelease as well as tags so the code is easier to find.

Decouple publishing from the pulumi/scripts repo.
- Copy the NPM install script to bin as part of the build.
- Calculate tag in a simple bash script.
- Just call `npm publish` directly. Fail if this package already exists as we want to know about it!

### DIFF
--- a/.github/workflows/publish-sdks.yml
+++ b/.github/workflows/publish-sdks.yml
@@ -224,7 +224,8 @@ jobs:
 
       - name: Calculate branch name
         id: branch
-        run: echo "name=${{ inputs.prerelease == true && 'master' || inputs.version }}" >> $GITHUB_OUTPUT
+        # Only write to master on a release - write to a stub branch for prerelease
+        run: echo "name=${{ inputs.prerelease == false && 'master' || inputs.version }}" >> $GITHUB_OUTPUT
 
       - name: Checkout branch
         run: git checkout ${{ steps.branch.outputs.name }}


.github/workflows/publish-sdks.yml
```diff
@@ -118,8 +114,6 @@ jobs:
           ${{github.workspace}}/sdk/dotnet
 
       - name: Publish NuGet Package
-        # TODO: Enable after testing
```
squaremo: Re-enabling all the publish steps is really its own thing; I would put it in a commit shuffled to the end of the PR branch.

.github/scripts/publish-npm-sdk.sh
```diff
@@ -1,76 +0,0 @@
...
-# Extracted from pulumi/scripts/ci/publish-tfgen-package
-
-# Publish the NPM package.
-echo "Publishing NPM package to NPMjs.com:"
-
-# First, add an install script to our package.json
-mkdir -p "${SOURCE_ROOT}/sdk/nodejs/bin/scripts"
-cp "${SCRIPT_ROOT}/install-pulumi-plugin.js" \
-  "${SOURCE_ROOT}/sdk/nodejs/bin/scripts"
-node "${SCRIPT_ROOT}/add-plugin-installer-script.js" < \
```
squaremo: This bit doesn't seem to be recreated in either the `Makefile` or `publish-sdks.yml`. Is it unnecessary, or .. what am I missing ..?

squaremo (commented):
I like the vast simplification. There's one thing I'm not sure on, so a comment rather than approval: the script that was extracted from pulumi/scripts, then superceded, did a bit of extra work to rewrite the package.json (it called https://github.com/pulumi/scripts/blob/master/ci/add-plugin-installer-script.js). Is that no longer needed?

.github/scripts/publish-npm-sdk.sh
danielrbradley: It looks like this was written before codegen started generating the installer script so this was redundant.

.github/workflows/publish-sdks.yml
danielrbradley: I'm planning to just squash this PR as the interim steps are not super-helpful and contain various WIP commits

.github/workflows/publish-sdks.yml
```diff
@@ -0,0 +1,275 @@
...
+
+jobs:
+  publish-python-sdk:
+    runs-on: ubuntu-latest
+    name: publish-python-sdk
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+        with:
+          submodules: true
```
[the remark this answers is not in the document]
danielrbradley: Hmm, probably not - might save a few seconds each run.

thomas11:
> Just call npm publish directly. Fail if this package already exists as we want to know about it!

Might we ever want to overwrite a published package, e.g., when it erroneously contains secrets?

danielrbradley:
> Might we ever want to overwrite a published package, e.g., when it erroneously contains secrets?

Nope, it's impossible to republish. You can delete but you can't then re-upload.

danielrbradley:
@squaremo 
> Is that no longer needed?

Nope this is done by code-gen now and this step is obsolete.

thomas11 (approved):

danielrbradley:
Full release just succeeded from this branch: https://github.com/pulumi/pulumi-azure-native/actions/runs/4056964235/jobs/6982623299

.github/workflows/publish-sdks.yml
squaremo: May I suggest two commits -- one squashing all the WIP, and one re-enabling bits of workflow (which would amount to unrelated changes in the previous commit).

.github/scripts/publish-npm-sdk.sh
squaremo: Got it, thanks for explaining. I think it's worth including that detail in the commit message -- that it can just be `npm publish`ed because
 1. some of the messing around with package.json that used to be done in this script is now obviated by codegen; and,
 2. it's better to bail if the package already exists
(and that latter point might bear some explanation, too)

.github/workflows/publish-sdks.yml
```diff
@@ -224,7 +224,8 @@ jobs:
 
       - name: Calculate branch name
         id: branch
-        run: echo "name=${{ inputs.prerelease == true && 'master' || inputs.version }}" >> $GITHUB_OUTPUT
+        # Only write to master on a release - write to a stub branch for prerelease
+        run: echo "name=${{ inputs.prerelease == false && 'master' || inputs.version }}" >> $GITHUB_OUTPUT
```
squaremo, line 228: Does `! inputs.prerelease` work? I realise this is GHA expression rather than say, the JavaScript that it resembles.

PR #2211, opened 2023-01-30
Remove old go sdk

### LINKED ISSUE #2192: Remove old Go SDK
The Go SDK at `~/sdk/go` is no longer generated or published because it's too large to download via Go's package cache. The's a new SDK now available from https://github.com/pulumi/pulumi-azure-native-sdk

The old SDK is still remaining in the main repository for reference for now, but we should remove this and reduce the size of checkouts.

danielrbradley:
Fixes #2192 

- Avoid packaging Go SDK artifact during build (we do this separately for the new SDK).
- Leave the readme file in the `sdk/go` folder with pointers to the new SDK location and migration guide.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -138,12 +138,12 @@ jobs:
         run: |
           cd sdk/pulumi-azure-native-sdk
           find . -maxdepth 1 -not -name '.*' -exec rm -rf {} \;
-      - name: Download Go split SDK
+      - name: Download Go SDK
         uses: actions/download-artifact@v3
         with:
           name: pulumi-azure-native-sdk.tar.gz
           path: ${{ github.workspace}}/sdk/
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yml
```diff
@@ -146,9 +146,11 @@ jobs:
           cd examples && go test -v -json -count=1 -cover -timeout 15m -short -tags=${{ matrix.language }} -parallel 4 . 2>&1 | tee /tmp/gotest.log | gotestfmt
 
       - name: Tar SDK folder
+        if: ${{ matrix.language != 'go' }}
```
thomas11, line 149: at some point we should rename all "go-split" naming to just "go"

danielrbradley: Agreed. Pushed!

PR #2215, opened 2023-02-01
Refactor workflow triggers

danielrbradley:
1. Combine release and prerelease trigger workflows and just calculate `prerelease` input.
2. Combine `publish-provider` and `publish-sdk` workflows. These are always run together so don't need to be callable independently.
3. Add the `shorttest` option to the dispatch on the `master` workflow so we can use this for testing.
4. Remove publishing from the acceptance tests as this can be done exactly the same by triggering the `master` workflow manually.
5. Make the `arm2pulumi` release part of the main release workflow
    - Skip publishing arm2pulumi on prerelease as we've not done pre-release versions of this before. This could be added and tested later.
    - Release after the main provider release is complete as that's the primary asset.
    - Combines all feedback into a single workflow run making arm2pulumi more of a first-class citizen


.github/workflows/release.yml
```diff
@@ -46,7 +20,7 @@ jobs:
     with:
       ref: ${{ github.ref }}
       version: ${{ github.ref_name }}
-      prerelease: false
+      prerelease: ${{ contains(github.ref_name,'-') }}
```
kpitzen, line 23: Does this describe the totality of prerelease refs we might publish?  And does its negation fully describe full releases?

danielrbradley: Yup, this is identical to the existing filters:

```
  push:
    tags:
      - v*.*.*
      - "!v*.*.*-**"
```

Release is any version number which doesn't have "-" in the last section. Prerelease is any version that contains a hyphen in the last section.

kpitzen: Ack!  That makes sense - since we're already filtering tags to symbols which match the above, this switch makes more sense.  Thanks!

kpitzen (approved):

PR #2229, opened 2023-02-07
Refine automated PR workflow process

danielrbradley:
Stop using the update PR description action as is not reliable and this can't get pulled into release notes anyway.

Disable automerge if there's breaking changes in the schema. Started by doing this in the PR workflow, but now doing this before even enabling automerge from the nightly workflow. Currently, we rely on the changelog label being missing to avoid automerging before review which is *not* reliable and will break when we move away from the manual changelog process to automated release notes. 

Also fixing up a bunch of warnings from the nightly workflow around invalid arguments and outdated runtimes.

Example workflow with breaking changes: https://github.com/pulumi/pulumi-azure-native/actions/runs/4115176481/jobs/7103534228
- Automerge step was skipped
- Resulting PR: https://github.com/pulumi/pulumi-azure-native/pull/2232

Example run which should also have warnings fixed: https://github.com/pulumi/pulumi-azure-native/actions/runs/4116003769

### DIFF
--- a/.github/workflows/nightly-sdk-generation.yml
+++ b/.github/workflows/nightly-sdk-generation.yml
@@ -87,7 +87,7 @@ jobs:
 
       - name: Create PR
         id: create-pr
-        uses: repo-sync/pull-request@v2.6.2
+        uses: repo-sync/pull-request@v2.12.1
         with:
           destination_branch: master
           github_token: ${{ secrets.PULUMI_BOT_TOKEN }}


.github/workflows/nightly-sdk-generation.yml
```diff
@@ -32,66 +32,92 @@ jobs:
...
+        with:
+          repo: pulumi/schema-tools
+
+      - name: Check Schema is Valid
+        id: schema-tools
+        shell: bash
+        run: |
+          echo 'summary<<EOF' >> $GITHUB_OUTPUT
+          schema-tools compare --provider="azure-native" --new-commit="--local-path=provider/cmd/pulumi-resource-azure-native/schema.json" >> $GITHUB_OUTPUT
+          echo 'EOF' >> $GITHUB_OUTPUT
```
thomas11, line 110: where is this GITHUB_OUTPUT used?

danielrbradley: In the next step - as part of the conditional: `steps.schema-tools.outputs.summary`

danielrbradley:
This has already been merged accidentally as part of this test PR: https://github.com/pulumi/pulumi-azure-native/pull/2233

Issue #2026, opened 2022-10-13
Pass through resource replacement in the schema missing for azure-native resources


tusharshahrs:
### What happened?

All inputs across all resources in azure-native are missing whether `changes to this property will trigger replacement`

Azure Native:
For example: [resourcegroup](https://www.pulumi.com/registry/packages/azure-native/api-docs/resources/resourcegroup/#inputs) doesn't show when changes are made to inputs such as location or name that 
this `will trigger replacement`.  
 

Azure Classic: 
For example,  [resourcegroup](https://www.pulumi.com/registry/packages/azure/api-docs/core/resourcegroup/#inputs) inputs  (such as location or name) state `changes to this property will trigger replacement`

[Pass through resource replacement in the schema was approved and merged](https://github.com/pulumi/pulumi/pull/9603#issuecomment-1129186128) for azure-native but the changes have not shown up in this provider.  Comment in regards to when this will show up in other [providers](https://github.com/pulumi/registry/issues/518#issuecomment-1150378303) 

### Steps to reproduce

Select any resource in https://www.pulumi.com/registry/packages/azure-native/api-docs/ and look at any of the inputs

### Expected Behavior

Expected that at least some of the inputs in each resource would have:   `Changes to this property will trigger replacement`

### Actual Behavior

Not a single resource in azure-native has inputs that are labeled with the information above.

### Output of `pulumi about`

n/a

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

thomas11:
pulumi/pulumi#9603 added the foundation for this in the engine. Now the provider needs to annotate the schema with the required properties.

Also related: #1216

mikhailshilkov:
The enhancement to the provider has been merged. The docs should update after the next release of Azure Native.

PR #2234, opened 2023-02-07
Add WillReplaceOnChanges annotations

mikhailshilkov:
Fix https://github.com/pulumi/pulumi-azure-native/issues/2026

As far as I know, this one-liner is all that is needed to show "changes to this property will trigger replacement" in Azure Native docs. It has no impact on provider functionality.


thomas11 (approved):
Lol that's how simple it was??

No way to test before releasing, though, is there?

danielrbradley (approved):
Wow that's an awesome quick win 👍

mikhailshilkov:
> No way to test before releasing, though, is there?

I checked it's in the schema.json, I assume the docs pipeline will work from here, but I can be wrong

thomas11:
> > No way to test before releasing, though, is there?
> 
> I checked it's in the schema.json, I assume the docs pipeline will work from here, but I can be wrong

Good enough, let's merge and see

PR #2260, opened 2023-02-23
Add new Resource Providers NetworkCloud and ResourceHealth

thomas11:
ResourceHealth has actually been around since 2015 but only the latest preview version matches our criteria for inclusion.

Fixes the nightly SDK generation failures.


danielrbradley (approved):
This will need the schema and SDKs re-generating but otherwise all good 👍

Issue #2264, opened 2023-02-24
Docs for get... methods say "that you want to create or update"


thomas11:
[Example](https://www.pulumi.com/registry/packages/azure-native/api-docs/compute/getgalleryimageversion/):
> azure-native.compute.getGalleryImageVersion
> Specifies information about the gallery image version that you want to create or update

PR #2276, opened 2023-02-27
Change the way we populate function descriptions

mikhailshilkov:
This PR changes the way we build the description for functions (invokes). By default, it takes the description of the corresponding HTTP method (GET or POST). If that description is empty, it falls back to the old way (use the description of the schema type of the response object).

This method seems to yield better results on average. Some description won't be perfect, but they are often framed in terms of "getting things" rather than "things", which is close to what we need.

The example in #2264 now says "Retrieves information about a gallery image version." instead of "Specifies information about the gallery image version that you want to create or update"

Fix https://github.com/pulumi/pulumi-azure-native/issues/2264


danielrbradley (approved):
This looks great! Much better documentation 🎉

PR #2279, opened 2023-02-28
Fix go publishing

danielrbradley:
- Use version input rather than reading version text file to ENV - wasn't exported either so wasn't available within fine sub-commands.
- Fix find command to use current directory, scanning only first level of subfolders for files named go.mod

### DIFF
--- a/.github/scripts/tag-go-modules.sh
+++ b/.github/scripts/tag-go-modules.sh
@@ -8,7 +8,7 @@ VERSION=${1}
 git tag "v${VERSION}"
 
 # This script tags all go modules in the repository with the current version
-find . -name "go.mod" -type f -mindepth 2 -maxdepth 2 -exec sh -c '
+find . -name "go.mod" -type f -mindepth 2 -maxdepth 2 -exec sh -cx '
   DIR=$(dirname ${1})
   BASE=$(basename $DIR)
   TAG="${BASE}/v${VERSION}"


danielrbradley:
Test run: https://github.com/pulumi/pulumi-azure-native/actions/runs/4294292530

.github/workflows/publish.yml
```diff
@@ -345,14 +345,13 @@ jobs:
...
           git config user.email "bot@pulumi.com"
           git add .
-          git commit -m "$VERSION"
-          git tag "$VERSION"
-          # Find go.mod's, get parent dir name, concat with version, set as a tag
-          find "*/go.mod" -type f -exec sh -c 'dirname {$1} | xargs basename | xargs printf "%s/$VERSION" | xargs git tag' shell {} \;
+          git commit -m "v${{ inputs.version }}"
+          git tag "v${{ inputs.version }}"
+          # Find go.mod files in immediate subdirectories, get parent dir name, concat with version, set as a tag
+          find . -name "go.mod" -type f -mindepth 2 -maxdepth 2 -exec sh -c 'dirname {$1} | xargs basename | xargs printf "%s/v${{ inputs.version }}" | xargs git tag' shell {} \;
```
squaremo, line 354: This is pretty difficult to read, with all the quoting. Could you make it easier to grok by using `for .. in` and assigning into vars?

squaremo: I'm not even clear that GitHub Actions will expand `${{ inputs.version }}` when it's within double quotes, themselves within single quotes (single quotes suppresses _shell_ expansion).

danielrbradley: Hmm, shellcheck's not keen https://www.shellcheck.net/wiki/SC2044

squaremo: Well yeah, shellcheck has a point. But here it's pretty safe, I would think, to use

```
for f in */go.mod; do
  name=`dirname "$f" | xargs basename`
  tag="$name/v${{ inputs.version }}"
  git tag "$tag"
done

Comment (squaremo) [reply .github/workflows/publish.yml:354]: If you still don't like using `for` though, you could just use `name=...` etc., in the command provided to `sh -c`.
If interpolation of `${{ inputs.version }}` is a problem, you can pass it as another argument to `sh -c` (and referred to within as `$2`).

Comment (squaremo) [reply .github/workflows/publish.yml:354]: I don't grok the `{$1}` bit -- specifically, why the enclosing braces?

Comment (danielrbradley) [reply .github/workflows/publish.yml]: Ok, gone with just extracting to a external script.

**Comment by danielrbradley** [comment]
Re-run: https://github.com/pulumi/pulumi-azure-native/actions/runs/4295278078

File: .github/scripts/tag-go-modules.sh [hunk 2]
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+test "${1}" = "" && echo "Usage: $0 <version>" && exit 1
+
+export VERSION=${1}
```
squaremo, line 5: Does this need to be exported? I think you only need it for expansion.

squaremo (commented):
One thing to clear up: what's `{$1}` for?

.github/scripts/tag-go-modules.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Tag root module
+git tag "v${VERSION}"
+
+# This script tags all go modules in the repository with the current version
+find . -name "go.mod" -type f -mindepth 2 -maxdepth 2 -exec sh -c '
+  DIR=$(dirname {$1})
```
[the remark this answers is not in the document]
danielrbradley: Yes ... bit of a left over from refactoring - works but is not needed. This is replacing the `{}` syntax for injecting each filename from the find.

.github/scripts/tag-go-modules.sh
danielrbradley: Just retested and seems you're right - fixed an issue while this was inline, but seems to be fine again now.

danielrbradley:
Re-test https://github.com/pulumi/pulumi-azure-native/actions/runs/4296949373

.github/scripts/tag-go-modules.sh
```diff
@@ -0,0 +1,16 @@
...
+
+test "${1}" = "" && echo "Usage: $0 <version>" && exit 1
+
+VERSION=${1}
+
+# Tag root module
+git tag "v${VERSION}"
+
+# This script tags all go modules in the repository with the current version
+find . -name "go.mod" -type f -mindepth 2 -maxdepth 2 -exec sh -c '
```
thomas11, line 11: Maybe add `-x` (print commands) to the `sh` invocation to get clearer logs?

thomas11 (approved):

.github/scripts/tag-go-modules.sh
danielrbradley: Nice! Not seen that before

PR #2280, opened 2023-03-01
Support importing storage blobs

### LINKED ISSUE #2107: Import of Blob storage not working
### What happened?

A customer tries to import a Blob storage in Pulumi and the import fails with following error message:

```bash
pulumi import azure-native:storage:Blob myresource2 /subscriptions/<uuid>/resourceGroups/<name>/providers/Microsoft.Storage/storageAccounts/<name>/blobServices/default/containers/<name>/blobs/cover.png
Previewing import (dev)

View Live: https://app.pulumi.com/dirien/azure-import-blob/dev/previews/af9f7f50-a05a-4bb6-be90-fd0132754cdd

     Type                          Name                   Plan       Info
     pulumi:pulumi:Stack           azure-import-blob-dev             1 error
 =   └─ azure-native:storage:Blob  myresource2            import     1 error


Diagnostics:
  azure-native:storage:Blob (myresource2):
    error: Preview failed: "resourceGroupName" not found in resource state

  pulumi:pulumi:Stack (azure-import-blob-dev):
    error: preview failed
```

When I checked the code in the provider, I can see in line 

https://github.com/pulumi/pulumi-azure-native/blob/d5ad08b26bfb7336cfba0de2eff6b2020c507f1a/provider/pkg/provider/provider.go#L990

a check for so called custom resources:

https://github.com/pulumi/pulumi-azure-native/blob/d5ad08b26bfb7336cfba0de2eff6b2020c507f1a/provider/pkg/provider/provider.go#L191

And Blob is one of the custom resources:

https://github.com/pulumi/pulumi-azure-native/blob/d5ad08b26bfb7336cfba0de2eff6b2020c507f1a/provider/pkg/resources/customresources.go#L50-L56

This leads to the situation, that the fields for `resourceGroupName` are not populated and the error message gets created

https://github.com/pulumi/pulumi-azure-native/blob/d5ad08b26bfb7336cfba0de2eff6b2020c507f1a/provider/pkg/resources/custom_storage.go#L591-L594

### Steps to reproduce

Just run the pulumi import command

### Expected Behavior

Imports the blob storage

### Actual Behavior

Not import the blob storage

### Output of `pulumi about`

```bash
 pulumi about
CLI          
Version      3.48.0
Go Version   go1.19.2
Go Compiler  gc

Plugins
NAME          VERSION
azure-native  1.87.0
go            unknown

Host     
OS       darwin
Version  12.5
Arch     arm64

This project is written in go: executable='/opt/homebrew/Cellar/go/1.19.3/libexec/bin/go' version='go version go1.19.3 darwin/arm64'




Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/dirien


Dependencies:
NAME                                       VERSION
github.com/pulumi/pulumi-azure-native/sdk  1.87.0
github.com/pulumi/pulumi/sdk/v3            3.48.0


### Additional context

-

### Contributing

### PR DESCRIPTION
**Pull request by mikhailshilkov**
Storage Blob is a custom resource that we implemented manually. The previous implementation assumes that properties like `resourceGroupName` are available on inputs, which isn't true during import (the properties map is empty since there's no state).

This PR passes the `id` from the provider to custom resources. Storage Blob parses the ID in case old properties aren't available, and uses those parsed values to read the Blob properties. Tested manually that importing a Blob works now.

Fix #2107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 - Treat location as case- and spacing-insensitive while calculating resource diffs
   [#2281](https://github.com/pulumi/pulumi-azure-native/pull/2281)
 
+### Bug fixes
+
+- Support importing storage blobs
+  [#2280](https://github.com/pulumi/pulumi-azure-native/pull/2280)
+
 ## 1.96.0 (2023-02-24)
 
 - New resource providers `networkcloud` (API version _2022-12-12-preview_) and `resourcehealth` (API version _2022-10-01-preview_).

### CODE REVIEW

File: provider/pkg/resources/custom_storage.go [hunk 1]
```diff
@@ -587,6 +593,46 @@ func (r *blob) read(ctx context.Context, properties resource.PropertyMap) (map[s
...
+	if len(blobServicesParts) != 2 {
+		return nil, false
+	}
+	acc := blobServicesParts[0]
+	blobsParts := strings.Split(blobServicesParts[1], "/blobs/")
+	if len(blobsParts) != 2 {
+		return nil, false
+	}
+	container := blobsParts[0]
+	name := blobsParts[1]
```
mikhailshilkov, line 626: I'm accepting ideas on more elegant implementation. Something something regex, maybe?

thomas11: go-autorest which we already import has https://pkg.go.dev/github.com/Azure/go-autorest/autorest/azure#ParseResourceID. Of course, autorest is deprecated :(

mikhailshilkov: Oh, that would be neat! But I guess we don't want to expand dependency on deprecated stuff...

kpitzen: If it's deprecated, is there any reason we can't reimplement the same functionality here?

mikhailshilkov: Looking at the implementation [here](https://github.com/Azure/go-autorest/blob/ee71315119d4d7088d74ca9fcbc7301ce2ed2bc1/autorest/azure/azure.go#L215), it doesn't cover all our needs, but I can adapt it to parse all relevant parts of the ID. Regex it is...

danielrbradley (commented):
I think the general implementation makes makes sense. Probably a good opportunity to add some targeted unit tests for this parsing logic - even just as future reference for the variations we need to handle.

provider/pkg/resources/custom_storage.go
mikhailshilkov: Changed to regex and added a test

kpitzen (approved):
LGTM

danielrbradley (approved):
Nice change with the Regex 👍

PR #2281, opened 2023-03-02
Apply case-spacing-insensitive diffing to all location properties

### LINKED ISSUE #1005: Import servicebus.NameSpace contains location such as "Central US" instead of "centralus" and this causes issues when standardizing on location naming convention.
When importing an azure-native, the location property is usually of the form "centralus" or "eastus" 
However, when importing servicebus.NameSpace resource, the location is returned as "Central US" or "East US" in the scaffolding code and the state.
And, if you try to change the location to "centralus" from "Central US", Pulumi does a replace due to this change instead of just accepting it as a noop.

## Steps to reproduce

1. Create a resource group and ServiceBus namespace (Pulumi code to do this is below to make life easier).
2. Create a blank project and pulumi up the stack
3. In the blank project stack directory, run `pulumi import azure-native:servicebus:Namespace NAMESPACE_NAME NAMESPACE_ID`
4. Add generated scaffolding code to the project, remove the `protect=True` resource option.
5. run `pulumi up` to strip the protect option and prove that the stack is happy and quiescent.
6. Change the location of the namespace resource to "centralus" 
7. run `pulumi up` and see that this change requires a replacement action.
8. If bored, accept the pulumi up and watch it delete the old and create the new.

Expected: 
Changing location from "Central US" or similar to "centralus" or equivalent should not cause a replacement event.
Actual: 
Changing the location as such results in a replacement event.

Helper code for step 1:
```
import pulumi as pulumi
import pulumi_azure_native as azure_native
from pulumi_azure_native import resources

# Create an Azure Resource Group
resource_group = resources.ResourceGroup('createrg')

namespace = azure_native.servicebus.Namespace("create-namespace",
    location="centralus",
    resource_group_name=resource_group.name,
    sku=azure_native.servicebus.SBSkuArgs(
        name="Standard",
        tier="Standard",
    ))

pulumi.export("namespace_name", namespace.name)
pulumi.export("namespace_id", namespace.id)
```

mikhailshilkov:
All `location` properties in Azure are case- and space- insensitive, so both "West Europe" and "westeurope" are valid values representing the same region.

This PR adjusts our diff calculation logic with this special rule. We get a diff calculated by vanilla platform diffing, then we check if there is an Update to the property called `location`, and if so, it normalizes both values and compares those. Any non-significant changes are ignored.

Fix #1809
Fix #1005

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## HEAD (Unreleased)
 
+### Enhancements
+
+- Treat location as case- and spacing-insensitive while calculating resource diffs
+  [#2281](https://github.com/pulumi/pulumi-azure-native/pull/2281)
+
 ## 1.96.0 (2023-02-24)
 
 - New resource providers `networkcloud` (API version _2022-12-12-preview_) and `resourcehealth` (API version _2022-10-01-preview_).


thomas11 (approved):
Wow, this is great! Nice tests, too.

I always thought it was more tricky than case and spaces, but I just checked via `az account list-locations` and it seems not. It's just a bunch of teams inside Azure who use custom short names like "northeu".

PR #2288, opened 2023-03-08
Fix build issues and add new RPs

thomas11:
Fixes the nightly SDK generation "Commit empty SDK" failures. [A successful run on this branch](https://github.com/pulumi/pulumi-azure-native/actions/runs/4360962657).

### DIFF
--- a/versions/v1-config.yaml
+++ b/versions/v1-config.yaml
@@ -445,6 +445,8 @@ FluidRelay:
   tracking: 2021-03-12-preview
   additions:
     listFluidRelayServerKeys: "2022-04-21"
+GraphServices:
+  tracking: 2022-09-22-preview
 GuestConfiguration:
   tracking: "2020-06-25"
 HDInsight:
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):
Any idea how we fix this and satisfy the actionlint warning? Perhaps we could split this command into two separate lines and use an interim variable?

danielrbradley:
Fixed the schema-tools with a new version release and added the explicit version to unblock the nightly job.

Test nightly run: https://github.com/pulumi/pulumi-azure-native/actions/runs/4364700298

danielrbradley:
@thomas11 I've pushed an extra tentative commit to address the actionlint warning too if you can review that back

danielrbradley:
Looks like we're in a better place now - failing with:

```
	* no version specified for ProfessionalService, available versions: 2023-07-01-preview
	* no version specified for ManagedNetworkFabric, available versions: 2023-02-01-preview
	* no version specified for GraphServices, available versions: 2022-09-22-preview
```

danielrbradley:
Re-run with new service versions added: https://github.com/pulumi/pulumi-azure-native/actions/runs/4364845105

PR #2292, opened 2023-03-10
Fix breaking API changes in networkfabric and orbital

thomas11:
- Manually resolve conflicting `OptionBProperties` for networkfabric
- Accept sibing `allOf` since Orbital uses it
- Regenerate SDKs

I wasn't 100% sure about the `allOf` change but the resulting Orbital 202211 SDK looks good to me.


danielrbradley (approved):

danielrbradley:
Yeah, pretty tricky to reason about the the intention here, but the SDK looks reasonable, so let's get this unblocked and released 👍

PR #2299, opened 2023-03-13
Fix release version calculation

danielrbradley:
Most recent release failed due to not being able to invoke the version normalisation script.

- Avoid using external script because we're not checking out the repo in this step (would slow things down quite a bit).
- Use bash variable replacement instead of sed for simplicity.
- Remove explicit (and incorrect) job names - all other jobs just use their id as the name.

### DIFF
--- a/.github/workflows/acceptance-test.yml
+++ b/.github/workflows/acceptance-test.yml
@@ -23,7 +23,6 @@ env:
 jobs:
   comment-notification:
     runs-on: ubuntu-latest
-    name: comment-notification
     steps:
       - name: Create URL to the run output
         id: vars
@@ -54,7 +53,6 @@ jobs:
 
   lint_workflows:
     runs-on: ubuntu-latest
-    name: Test Provider
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3


.github/workflows/release.yml
```diff
@@ -8,11 +8,11 @@ on:
 jobs:
   version:
     runs-on: ubuntu-latest
-    name: Calculate Dev Version
     steps:
       - id: version
         name: Calculate build version
-        run: echo "version=$(.github/scripts/normalise-tag-version.sh ${{ github.ref_name }})" >> "$GITHUB_OUTPUT"
+        # Remove "v" if present
+        run: echo "version=${GITHUB_REF_NAME/v/}" >> "$GITHUB_OUTPUT"
```
thomas11, line 15: This will replace the first 'v' even if it's not at the beginning. Currently, that doesn't matter, but if we ever get a GITHUB_REF_NAME without a leading 'v' we might get a surprising substitution here.

thomas11 (approved):

.github/workflows/release.yml
danielrbradley: > This will replace the first 'v' even if it's not at the beginning. Currently, that doesn't matter, but if we ever get a GITHUB_REF_NAME without a leading 'v' we might get a surprising substitution here.

Hmm, actually this could cause an issue if anyone tried a prerelease tag e.g. v1.2.3-dev - would be 1.2.3-de 🤔 not often, but we should probably account for it

thomas11: So much for replacing sed for simplicity :)

squaremo: How will this be resolved? I think it's worth embellishing on the comment there, since it's (demonstrably) not obvious this has a weakness.

danielrbradley: > only the first occurrence of X is replaced

Ah, there's no issue then because this workflow only runs if there's a 'v' at the start of the tag.

PR #2301, opened 2023-03-14
Add pulumi-plugin to all go SDK sub-modules

### LINKED ISSUE #2300: Split go SDK doesn't download plugin correctly
The `pulumi-plugin.json` doesn't exist in sub-modules so the engine does not discover the plugin correctly.

`pulumi-plugin.json` needs to be copied into each module sub-folder.

danielrbradley:
Fixes #2300

No generated difference because the SDK is not committed here.

The pulumi-plugin.json doesn't exist in sub-modules so the engine does not discover the plugin correctly.

pulumi-plugin.json needs to be copied into each module sub-folder.

### DIFF
--- a/provider/cmd/pulumi-gen-azure-native/main.go
+++ b/provider/cmd/pulumi-gen-azure-native/main.go
@@ -5,12 +5,13 @@ package main
 import (
 	"bytes"
 	"fmt"
-	"github.com/pulumi/pulumi/pkg/v3/codegen"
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
A kind reminder to add a meaningful description to each PR with both motivation of a change and a short description of it. Thank you!

thomas11 (approved):

PR #2306, opened 2023-03-17
Remove arm2pulumi coverage workflow

squaremo:
The result is not consulted, and workflow failures are ignored. So: just don't do it.


danielrbradley (commented):
We should also remove:
- [ ] `provider/pkg/arm2pulumi/internal/test/s3-upload-script.sh`
- [x] `provider/pkg/arm2pulumi/internal/test/results.json`
- [ ] Maybe `provider/pkg/arm2pulumi/internal/test/quickstart_coverage_test.go`

squaremo:
> I think that other test file is only useful in the context of this coverage report so could be removed too, but this is a good first step to get the nightly failures fixed.

It's possible to run the tests with coverage locally (and it passes!) -- `make arm2pulumi_coverage_report` -- so I'm inclined to leave that test code in. But yeah, borderline case.

PR #2308, opened 2023-03-20
MSAL and OIDC support

thomas11:
## About

This PR contains two related pieces:
1. Switch the provider from the deprecated ADAL authentication method to the current MSAL method. This is discussed in detail in the document [MSAL auth in the Azure Native Provider](https://docs.google.com/document/d/1MvbucF5w2rwssTJKRQLut0wwm9OILk0PxLurWEEcy-g/edit?usp=sharing). However, this document outlines three options for using MSAL, each with drawbacks. Since writing this document I believe I have found a better option that this PR implements: detect the unsupported cli case dynamically and fall back to ADAL in that case.
2. Enable OIDC for authentication. The actual OIDC implementation is in our dependencies but enabling it was gated behind MSAL. This PR exposes the relevant configuration settings and adds a dedicated test.

## Testing

I picked an existing e2e test, `keyvault`. In general, any e2e test will do since we only want to see that authentication works. However, `keyvault` has more authentication test coverage than the other tests because it also creates a key vault secret, which is a data plane operation that requires a different token.

I used Go build tags to make this test as `oidc`. This required moving it into its own file.

In the OIDC test run, we run only this test, unsetting the ARM client secret to make sure we really test OIDC.

## Next steps

To fully release OIDC support we need to document it in our user-facing documentation.

I should also document the particular OIDC integration used in testing in our wiki (AD principal, roles, permissions etc.).

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -164,25 +164,25 @@ jobs:
           cd examples
           go test -v -json -count=1 -cover -timeout 15m -tags=${{ matrix.language }},oidc -parallel 4 . 2>&1 | tee /tmp/gotest.log | gotestfmt
 
-      - name: Log in via az CLI for OIDC tests
-        if: ${{ matrix.language == 'nodejs' }}
-        uses: azure/login@v1
-        with:
-          # tkappler oidc-test - temporary unless our main app is OIDC-enabled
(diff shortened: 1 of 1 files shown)


danielrbradley:
Replaced with #2320

PR #2310, opened 2023-03-21
Refactor version generation process

danielrbradley:
Generate versions as part of the main generation program

Aims:
- Remove build of versioner binary.
- Remove double load of specs (for versions then schema) - keep it all in memory.
- Simplify makefile to remove all version-specific file targets - all the time is spent reading the specs - so might as well generate all version files at the same time.
- Implement switch to allow the upgrading of the config during nightly upgrade jobs.

Asides:
- Fix validation of language inputs
- Renamed version files (`curations` -> `config` -> `vX` becomes `config` -> `spec` -> `lock`)

### DIFF
--- a/provider/pkg/openapi/discover.go
+++ b/provider/pkg/openapi/discover.go
@@ -117,7 +117,7 @@ func AddDefaultVersion(providers map[string]map[string]VersionResources, default
 	for providerName, versionMap := range providers {
 		// Add a default version for each resource and invoke.
 		defaultResourceVersions := defaultVersion[providerName]
-		versionMap[""] = buildCuratedVersion(versionMap, defaultResourceVersions)
+		versionMap[""] = buildDefaultVersion(versionMap, defaultResourceVersions)
 
 		// Set compatible versions to all other versions of the resource with the same normalized API path.
 		pathVersions := calculatePathVersions(versionMap)
@@ -155,10 +155,10 @@ func ApplyDeprecations(providers AzureProviders, deprecated ProviderVersionList)
 	return providers
 }
 
-func buildCuratedVersion(versionMap ProviderVersions, curatedResourceVersions map[ResourceName]ApiVersion) VersionResources {
+func buildDefaultVersion(versionMap ProviderVersions, defaultResourceVersions map[ResourceName]ApiVersion) VersionResources {
 	resources := map[string]*ResourceSpec{}
 	invokes := map[string]*ResourceSpec{}
-	for resourceName, apiVersion := range curatedResourceVersions {
+	for resourceName, apiVersion := range defaultResourceVersions {
 		if versionResources, ok := versionMap[ApiToSdkVersion(apiVersion)]; ok {
 			if resource, ok := versionResources.Resources[resourceName]; ok {
 				resourceCopy := *resource


provider/pkg/openapi/discover.go
```diff
@@ -76,28 +76,33 @@ type ResourceSpec struct {
...
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 
+	return ApplySpecModifications(providers, defaultVersion, deprecated, removed), nil
+}
+
+// ApplySpecModifications adds the default version for each provider and deprecates and removes specified API versions.
+func ApplySpecModifications(providers AzureProviders, defaultVersion CuratedVersion, deprecated, removed ProviderVersionList) AzureProviders {
```
thomas11: I know it's not from this PR but it's confusing that the type name is `CuratedVersion` and not `CuratedVersions`.

provider/pkg/versioning/gen.go
```diff
@@ -21,11 +32,11 @@ func GenerateVersionFiles(providers openapi.AzureProviders, refreshConfig bool)
 
 	v1Config, err := ReadDefaultConfig(path.Join("versions", "v1-config.yaml"))
 	if err != nil {
-		return nil, err
+		return VersionMetadata{}, err
```
thomas11, line 34: Isn't returning `nil` more in case of error more idiomatic? What's the rationale behind this change?

CONTRIBUTING.md
```diff
@@ -51,41 +51,31 @@ Azure's API specifications are organised into "Resource Providers" (or "Services
...
 
 - `tracking` specifies a single version of the API to fetch all resources from.
 - `additions` specified a map of specific resource (or invoke) name to API version to include in the default version.
 
 Both `tracking` and `additions` can be specified together but if the set of resources overlaps (the same resource exists) in both the tracking version and the additions, then the versioning program will fail with an error.
 
-From a config YAML file, we can generate a version 'lock' JSON file ([`v1.json`](./versions/v1.json) and [`v2.json`](./versions/v2.json)) which contain the fully expanded set of default resources. These are the files used at the point of generating the schema.
+From a config YAML file, we can generate a version 'lock' JSON file ([`v1-lock.json`](./versions/v1-lock.json) and [`v2-lock.json`](./versions/v2-lock.json)) which contain the fully expanded set of default resources. These are the files used at the point of generating the schema.
 
 To re-calculate the default version selection, run `make versions`.
```
thomas11, line 80: `make versions` is removed, no?

thomas11 (commented):
This great. Love the simplifications and clearer new names.

The only thing I'm not sure about is the rename `v2-curation` -> `config`. 
  1. I find "curation" is a bit clearer what this file is about than the generic "config".
  2. Shouldn't we keep the major version prefix? When v3 comes around, won't we want to preserve the v2 curation? Or is the idea to have Git history handle that?

provider/pkg/versioning/gen.go
danielrbradley: Previously we were returning a Map which could be nil, but now we're returning a struct. So we could change it to be a pointer to a struct, but opted for just using the struct directly as pointer syntax are kinda noisy in Go

provider/pkg/openapi/discover.go
danielrbradley: Renaming this completely to DefaultVersionLock to avoid ambiguity.

thomas11:
> @thomas11 I'm still not 100% settled on the naming. I wanted to move away from curations as that term was already overloaded internally.
> 
> I've now renamed "curations" internally to "default version" as that's much more clear and specific so that reason has now gone.
> 
> One other option is "constraints" rather than "config". I still find "curations" a little niché in language and constraints is much more of a well known word.

Alright, let's stick with "default version" then. It expresses nicely what the end goal is. "Constraints" sounds to me more like a version range like in a package manager.

PR #2313, opened 2023-03-22
Combined dependabot security upgrades

thomas11:


rquitales (approved):

sdk/go.mod
danielrbradley, line 1: Oh this removal is interesting - should we just remove this mod now that we publish this elsewhere?

thomas11: Oh, TBH I didn't even notice. I just ran `go mod tidy` and checked that there was no error. I guess we can remove it.

thomas11: Oh, TBH I hadn't even noticed. I just ran `go mod tidy`. I guess we can remove it.

danielrbradley: Don't know if it's worth leaving some kind of stub there so people will be directed to the new location ... but might not need the mod for that.

PR #2320, opened 2023-03-22
MSAL and OIDC support

thomas11:
## About

This PR contains two related pieces:
1. Switch the provider from the deprecated ADAL authentication method to the current MSAL method. Due to an issue in our dependencies, one of the authentication paths, the `az` cli, doesn't work properly with MSAL. We detect the unsupported cli case dynamically and fall back to ADAL in that case.
2. Enable OIDC for authentication. The actual OIDC implementation is in our dependencies but enabling it was gated behind MSAL. This PR exposes the relevant configuration settings and adds a dedicated test.

### What's MSAL?

The provider currently uses an authentication method called ADAL that’s out of support in Azure. ADAL, also called auth v1, is [replaced by MSAL](https://learn.microsoft.com/en-us/azure/active-directory/develop/msal-migration) (auth v2).

OIDC only works on MSAL.

These protocols are implemented in our dependencies _go-azure-helpers_ and _hamilton_ so we don't need to care about the details, except for two things that surface in our code:
  - MSAL has a new concept of _scopes_ for more targeted authentication scopes. You can't just get a token for everything in an Azure account or subscription, you can only get one for ResourceManager, or KeyVault, etc. Since Pulumi deals with infrastructure, we can get away with ResourceManager for most operations, but we do have some [custom resources](https://github.com/pulumi/pulumi-azure-native/pull/2320/files#diff-bd590eb4f290efeb8760fc0743f5af2db35ceb776e7ef64a1cb641303e878426) that touch the data plane and need a different scope.
  - The dependencies [don't implement MSAL correctly when the `az` cli is used](https://github.com/manicminer/hamilton/issues/189) for authentication. This PR falls back to ADAL when that case is detected.

### What's OIDC?

It stands for **OpenID Connect** and is an authentication protocol on top of OAuth. The idea is to allow trusted parties to verify a user's identity. For a practical example, let's look at how I set up OIDC authentication between Github Actions and Azure, which is used by the e2e test of this PR. We tell Azure that GitHub can be trusted:

1. On an Azure Active Directory (AD) App, add a new credential with issuer GitHub.
2. In GitHub, configure the Action with the id of the AD App.

Now, when the action runs, GitHub can use the Azure AD endpoint to exchange the GitHub token for an Azure token.

Note that there is no secret involved anywhere that would need to be safely stored, rotated, etc. This is one of the biggest advantages of OIDC.

Details for the interested [here (GH)](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers) and [here (Azure)](https://learn.microsoft.com/en-us/azure/active-directory/workload-identities/workload-identity-federation-create-trust?pivots=identity-wif-apps-methods-azp#configure-a-federated-identity-credential-on-an-app), but not required reading for reviewing this PR.

## Testing

I picked an existing e2e test, `keyvault`. In general, any e2e test will do since we only want to see that authentication works. However, `keyvault` has more authentication test coverage than the other tests because it also creates a key vault secret, which is a data plane operation that requires a different token.

I used Go build tags to make this test as `oidc`. This required moving it into its own file. In the OIDC test run, we run only this test, unsetting the ARM client secret to make sure we really test OIDC.

## Reviewing this PR

I added comments with leading numbers like **(1)**. I suggest following those in order. They show on the main page in the correct order, but without a link to the actual file, so that's kinda annoying.

## Next steps

To fully release OIDC support we need to document it in our user-facing documentation.

I should also document the particular OIDC integration used in testing in our wiki (AD principal, roles, permissions etc.).

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -138,7 +138,8 @@ jobs:
       - name: Run tests
         if: "!inputs.short_test"
         env:
-          RUN_OIDC_TESTS: true
+          # specifying this id will cause the OIDC test(s) to run against this AD application
+          OIDC_ARM_CLIENT_ID: "89380e12-5be6-486a-89ef-eea107af2f47" # AD app 'oidc-test'
         run: |
           set -euo pipefail
           cd examples && go test -v -json -count=1 -cover -timeout 2h -tags=${{ matrix.language }} -parallel 4 . 2>&1 | tee /tmp/gotest.log | gotestfmt
--- a/examples/examples_nodejs_keyvault_test.go
+++ b/examples/examples_nodejs_keyvault_test.go
@@ -22,18 +22,19 @@ func TestAccKeyVaultTs(t *testing.T) {
 
 func TestAccKeyVaultTs_OICD(t *testing.T) {
 	skipIfShort(t)
-	if os.Getenv("RUN_OIDC_TESTS") != "true" {
-		t.Skip("Skipping OIDC test outside of CI")
+
+	oidcClientId := os.Getenv("OIDC_ARM_CLIENT_ID")
+	if oidcClientId == "" {
+		t.Skip("Skipping OIDC test without OIDC_ARM_CLIENT_ID")
 	}
 	test := getJSBaseOptions(t).
 		With(integration.ProgramTestOptions{
 			Dir: filepath.Join(getCwd(t), "keyvault"),
 			Env: []string{
 				"ARM_USE_OIDC=true",
+				"ARM_CLIENT_ID=" + oidcClientId,
 				// not strictly necessary but making sure we test the OIDC path
 				"ARM_CLIENT_SECRET=",
-				// AD app 'oidc-test', owned by tkappler - temporary until our main app is OIDC-enabled
-				"ARM_CLIENT_ID=89380e12-5be6-486a-89ef-eea107af2f47",
 			},
 		})
 


provider/pkg/provider/auth.go
```diff
@@ -0,0 +1,293 @@
+// Copyright 2016-2022, Pulumi Corporation.
```
thomas11, line 1: Since the authentication code grew to about 300 lines, I moved it to its own file. Much of it is unchanged.

provider/pkg/provider/auth.go
```diff
@@ -0,0 +1,293 @@
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to unmarshal '%s' as Auxiliary Tenants", auxTenantsString)
+		}
+	}
+	envName := k.getConfig("environment", "ARM_ENVIRONMENT")
+	if envName == "" {
+		envName = "public"
+	}
+
+	useMsi := k.getConfig("useMsi", "ARM_USE_MSI") == "true"
```
thomas11, line 43: Read the OIDC config from the environment. Lines 47-65, 80-85 are new. It's standard configuration handling, except that there are two environment variables that we check for token and URL. The ones starting with `ARM_` are ours for the user to set, the ones with `ACTIONS_` are pre-set when running in GHA, so we check them directly instead of making users get the values and put them into the `ARM_` variables.

provider/pkg/provider/auth.go
```diff
@@ -0,0 +1,293 @@
...
+		SupportsClientCertAuth:         true,
+		SupportsClientSecretAuth:       true,
+		SupportsOIDCAuth:               useOIDC,
+		UseMicrosoftGraph:              useOIDC,
+		SupportsManagedServiceIdentity: useMsi,
+		SupportsAzureCliToken:          !useOIDC,
+		SupportsAuxiliaryTenants:       len(auxTenants) > 0,
+	}
+
+	needCli := needAzCli(builder)
```
thomas11, line 97: Check whether we'll use the `az` cli for authentication. That isn't new but I extracted it into the `needAzCli` function and added the OIDC case. Both OIDC and MSI auth are not supported via cli.

provider/pkg/provider/auth.go
```diff
@@ -0,0 +1,293 @@
...
+
+// Without any of the other methods above, we fall back to the `az` CLI to authenticate.
+func needAzCli(builder *authentication.Builder) bool {
+	return !builder.SupportsManagedServiceIdentity &&
+		!builder.SupportsOIDCAuth &&
+		builder.ClientSecret == "" &&
+		builder.ClientCertPath == ""
+}
+
+func getAzVersion() (*goversion.Version, error) {
```
thomas11, line 126: The following code is all about cli versions and is just moved over from provider.go. It gets interesting again in line 190.

provider/pkg/provider/auth.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	if err := json.Unmarshal(stdout.Bytes(), target); err != nil {
+		return fmt.Errorf("unmarshaling the result of Azure CLI: %w", err)
+	}
+
+	return nil
+}
+
+type AuthorizerFactory func(api environments.Api) (autorest.Authorizer, error)
```
thomas11, line 190: Autorest, the library we use to talk to the Azure REST API, takes an `autorest.Authorizer` interface to get auth tokens from. With ADAL, we could just make an Authorizer and use it everywhere. With MSAL, due to the added scopes, we need a factory to create Authorizers on demand for different scopes. Note how the token factory is a `func(api environments.Api)` but in the ADAL case, the `api` parameter is unused.

provider/pkg/provider/auth.go
```diff
@@ -0,0 +1,293 @@
...
+	buildSender := sender.BuildSender("AzureNative")
+
+	oauthConfig, err := k.buildOAuthConfig(authConfig.Config)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	endpoint := k.environment.TokenAudience
+
+	if authConfig.useCli {
```
thomas11, line 204: As mentioned above, due to [this issue](https://github.com/manicminer/hamilton/issues/189) we fall back to ADAL if we're using the cli.

The rest of auth.go is largely unchanged.

provider/pkg/provider/provider.go
```diff
@@ -181,16 +177,39 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
...
 	k.environment = env
 
-	tokenAuth, bearerAuth, err := k.getAuthorizers(authConfig)
+	hamiltonEnv := k.autorestEnvToHamiltonEnv()
+
+	// The ctx Context given by gRPC is request-scoped and will be canceled after this request. We
+	// need the authorizers to function across requests.
+	authCtx := context.Background()
+
+	tokenAuth, bearerAuth, err := k.makeAuthorizerFactories(authCtx, authConfig)
```
thomas11, line 186: In the provider's `Configure`, we use the new autorizer factories to make authorizers for `ResourceManager` and for `KeyVault`. The RM one is our standard authorizer because we generally only do resource management: `k.client.Authorizer = resourceManagerAuth`. The Key Vault one is passed to the custom resources builder so we can handle secrets there.

The rest of provider.go is just code moved to auth.go.

examples/examples_nodejs_keyvault_test.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2016-2020, Pulumi Corporation.  All rights reserved.
+//go:build nodejs || (nodejs && oidc) || all
```
thomas11, line 2: We use the existing test `TestAccKeyVaultTs` because it creates both resources and secrets, so both MSAL scopes are needed.

I moved it to its own file, unchanged, so I could add go build tags. It's run whenever tag `nodejs` is given, with or without `oidc`, whereas the other tests only run when `!oidc`.

The boolean expression `nodejs || (nodejs && oidc)` could be simplified to just `nodejs` but I wanted to show the duality with the other file's `nodejs && !oidc` - not sure if worth it.

.github/workflows/build-test.yml
```diff
@@ -139,12 +141,45 @@ jobs:
...
 
-      - name: Run tests
+      - name: Run short tests
         if: inputs.short_test
         run: |
           set -euo pipefail
           cd examples && go test -v -json -count=1 -cover -timeout 15m -short -tags=${{ matrix.language }} -parallel 4 . 2>&1 | tee /tmp/gotest.log | gotestfmt
 
+      - name: Run OIDC tests
+        if: ${{ matrix.language == 'nodejs' }}
```
thomas11, line 151: The `matrix.language == 'nodejs'` condition is a bit ugly because we could have OIDC tests in other languages, too. I added it because otherwise I would have had to add a `!oidc` tag to all other test files, or they would run twice, with and without `oidc`.

.github/workflows/build-test.yml
```diff
@@ -139,12 +141,45 @@ jobs:
...
+        env:
+          ARM_USE_OIDC: true
+          ARM_CLIENT_SECRET: ""
```
thomas11, line 154: We unset `ARM_CLIENT_SECRET` that the other tests use for authentication to disable that path so we know we're testing what we think we're testing.

.github/workflows/build-test.yml
```diff
@@ -139,12 +141,45 @@ jobs:
...
+          # AD app 'oidc-test', owned by tkappler - temporary until our main app is OIDC-enabled
```
thomas11, line 155: I lack the permissions to add the GitHub credential to our main App, need to follow up on that.

.github/workflows/build-test.yml
```diff
@@ -139,12 +141,45 @@ jobs:
...
+          ARM_CLIENT_ID: 89380e12-5be6-486a-89ef-eea107af2f47
+          TMPDIR: ${{ runner.temp }}
+        run: |
+          set -euo pipefail
+          cd examples
+          go test -v -json -count=1 -cover -timeout 15m -tags=${{ matrix.language }},oidc -parallel 4 . 2>&1 | tee /tmp/gotest.log | gotestfmt
+
+      ### tkappler: OIDC via the 'az' cli is generally possible, but the current implementation in
```
thomas11, line 163: Should probably be taken out. I figured out how to test the path where the `az` cli is used to do the OIDC handshake with Azure, only to realize it doesn't work in our case because [`go-azure-helpers` doesn't support it](https://github.com/hashicorp/go-azure-helpers/blob/v0.51.0/authentication/auth_method_azure_cli_token.go#L75).

examples/examples_nodejs_keyvault_test.go
danielrbradley: Another approach on testing for specific environments is to detct if the required OIDC configuration is available and either use `t.Skip()` or fail with an assert if not.

Then in the test we can manipulate the ENV of the program under test by setting the `Env` option in the `With` function:

```go
		With(integration.ProgramTestOptions{
			Dir: filepath.Join(getCwd(t), "keyvault"),
			Env: []string{"ARM_CLIENT_SECRET="},
		})
```

Docs: https://github.com/pulumi/pulumi/blob/master/pkg/testing/integration/program.go#L290

I think this would allow us to revert the modifications to the workflows for this specific case.

thomas11: Using `ProgramTestOptions` seems promising. Not sure I see the full picture, though. We want to run the "keyvault" test twice - authenticating once via secret, once via OIDC. Therefore, we need two different `go test` runs from CI with the different auth config. If we use `ARM_USE_OIDC` to detect OIDC, wouldn't every _other_ test need to check for it and skip if set, so we don't run everything twice?

Another option is to just always run the "keyvault" OIDC test by making a copy of it that sets all the OIDC config via `ProgramTestOptions`. Downside is that OIDC will necessarily fail when run from a developer's machine, only GitHub has the credentials.

danielrbradley: I'm possibly not seeing exactly what we need to set where. Essentially my thought is that we set the ENV required for both in CI. Then each test can set or unset the parts of the ENV it needs for the test.

I imagine you would have 2 copies of the Go test - which only vary by the ENV modifications.

Default ENV can also be set in the `examples_test.go` `getBaseOptions` method so you shouldn't have to modify each test individually.

thomas11: How would we **not** run the OIDC test locally, where it cannot succeed?

danielrbradley: Check the env during the setup of the test and skip the test if OIDC not se in the ENV.

danielrbradley: Here's an example of how the new test could be written in the existing file:

```go
func TestAccKeyVaultOidcTs(t *testing.T) {
	skipIfShort(t)
	if os.Getenv("CI") != "true" {
		t.Skip("Skipping OIDC test outside of CI")
	}
	test := getJSBaseOptions(t).
		With(integration.ProgramTestOptions{
			Dir: filepath.Join(getCwd(t), "keyvault"),
			Env: []string{"ARM_USE_OIDC=true", "ARM_CLIENT_SECRET="},
		})

	integration.ProgramTest(t, &test)
}
```

thomas11: Pushed. Although I introduced a dedicated env variable for this, I found it cleaner.

examples/examples_nodejs_test.go
```diff
@@ -43,16 +42,6 @@ func TestAccCosmosDBTs(t *testing.T) {
...
-func TestAccKeyVaultTs(t *testing.T) {
-	skipIfShort(t)
-	test := getJSBaseOptions(t).
-		With(integration.ProgramTestOptions{
-			Dir: filepath.Join(getCwd(t), "keyvault"),
-		})
-
-	integration.ProgramTest(t, &test)
-}
-
```
danielrbradley, line 55: Is it worth leaving this in copy of the test so it'll run against both the old and new auth method?

examples/examples_nodejs_keyvault_test.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: filepath.Join(getCwd(t), "keyvault"),
+			Env: []string{
+				"ARM_USE_OIDC=true",
+				// not strictly necessary but making sure we test the OIDC path
+				"ARM_CLIENT_SECRET=",
+				// AD app 'oidc-test', owned by tkappler - temporary until our main app is OIDC-enabled
+				"ARM_CLIENT_ID=89380e12-5be6-486a-89ef-eea107af2f47",
```
danielrbradley, line 36: Rather than hardcoding this here, should we expect it to be set in a different ENV variable for the test (e.g. `OIDC_ARM_CLIENT_ID`) then this test will set `ARM_CLIENT_ID` from `OIDC_ARM_CLIENT_ID` value - essentially `ARM_CLIENT_ID=${OIDC_ARM_CLIENT_ID}`, but written in Go?

Also, we could then just check if `OIDC_ARM_CLIENT_ID` is set to determine if we should skip the test - maybe nicer than the extra `RUN_OIDC_TESTS` env var.

danielrbradley (approved):
This looks loads better - much happier it's not complicating the build process!

Good to go as-is but have added a comment for moving where the hard-coded value sits.

examples/examples_nodejs_test.go
thomas11: It already does, the key vault file has two copies of the test.

kpitzen (approved):

PR #2321, opened 2023-03-22
Remove newly added preview versions to reduce SDK size

thomas11:
This will hopefully allow us to publish the NodeJS SDK which currently fails due to too many files.

I looked at `git diff --no-ext-diff -U400 5500e55190c -- spec.json` and removed the preview versions that showed as added in the diff.


danielrbradley (approved):
The following new preview services were removed before release in https://github.com/pulumi/pulumi-azure-native/issues/2295:

- managednetworkfabric
- graphservices
- professionalservice

Therefore we accidentally published docs for them, but they weren't present in the SDKs. This accounts for all listed breaking changes.

PR #2339, opened 2023-03-29
Skip OIDC tests in release workflow

thomas11:


danielrbradley (approved):
Looks good to me - can test via a pre-release build

thomas11:
> Looks good to me - can test via a pre-release build
> 
> Test run here: https://github.com/pulumi/pulumi-azure-native/actions/runs/4552467167
> 
> Run failed 🛑

The OIDC did run when it shouldn't have. The failure of the test itself is then expected. Will look into it.

PR #2346, opened 2023-03-30
Generate initial v2 schema with previous API versions documented

danielrbradley:
- Add ability to generate a draft v2 schema (in `bin/v2/schema-full.json`).
- For resources that existed in v1, include the previous version number in the new resource description.
- Unblock schema generation by making more types mergable and adding a couple more exceptions.
- Make schema generation testable - needed to diagnose memory issues introduced by intermediate bug.
  - Pass in all paths to load files from rather than assuming our working directory is always the repo root.
  - Work around issue in open api schema loading which truncated leading parent navigation in the root path due to being treated as a URL rather than only a file path.

Talking points:
- Are these exceptions ok or should we address them in a different way?

Preview of resource description including previous version:

```
Domain service.
API Version: 2022-12-01.
Previous API Version: 2021-03-01. See https://github.com/pulumi/pulumi-azure-native/discussions/TODO for information on migrating from v1 to v2 of the provider.
```

### Generated SDK changes

In the Network namespace there's multiple definitions of the `Sku` type. They all have a `Name` property which can be a string, but they each also union with an enum of their own valid values. 

Token `azure-native:network:Sku`
- "BastionHost": `azure-rest-api-specs/specification/network/resource-manager/Microsoft.Network/stable/2022-09-01/bastionHost.json`
- "WebApplicationFirewallPolicy": `"azure-rest-api-specs/specification/frontdoor/resource-manager/Microsoft.Network/stable/2022-05-01/webapplicationfirewall.json`

Token `azure-native:network/v20220501:Sku`:
- `azure-rest-api-specs/specification/frontdoor/resource-manager/Microsoft.Network/stable/2022-05-01/webapplicationfirewall.json` (WebApplicationFirewallPolicy)
- `azure-rest-api-specs/specification/network/resource-manager/Microsoft.Network/stable/2020-05-01/bastionHost.json` (BastionHost)

Token `azure-native:eventgrid:EventSubscriptionFilterResponse`: this is due to the v1 config using multiple versions of the API so the input set is cut back to the minimum common set.

### DIFF
--- a/sdk/dotnet/Network/V20220501/Inputs/SkuArgs.cs
+++ b/sdk/dotnet/Network/V20220501/Inputs/SkuArgs.cs
@@ -19,7 +19,7 @@ namespace Pulumi.AzureNative.Network.V20220501.Inputs
         /// Name of the pricing tier.
         /// </summary>
         [Input("name")]
-        public InputUnion<string, Pulumi.AzureNative.Network.V20220501.SkuName>? Name { get; set; }
+        public object? Name { get; set; }
 
         public SkuArgs()
         {
--- a/sdk/python/pulumi_azure_native/network/v20220501/_inputs.py
+++ b/sdk/python/pulumi_azure_native/network/v20220501/_inputs.py
@@ -16400,24 +16400,24 @@ class ServiceEndpointPropertiesFormatArgs:
 @pulumi.input_type
 class SkuArgs:
     def __init__(__self__, *,
-                 name: Optional[pulumi.Input[Union[str, 'SkuName']]] = None):
+                 name: Optional[pulumi.Input[Union[str, 'SkuName', 'BastionHostSkuName']]] = None):
         """
         The pricing tier of the web application firewall policy.
-        :param pulumi.Input[Union[str, 'SkuName']] name: Name of the pricing tier.
+        :param pulumi.Input[Union[str, 'SkuName', 'BastionHostSkuName']] name: Name of the pricing tier.
         """
         if name is not None:
             pulumi.set(__self__, "name", name)
 
     @property
     @pulumi.getter
-    def name(self) -> Optional[pulumi.Input[Union[str, 'SkuName']]]:
+    def name(self) -> Optional[pulumi.Input[Union[str, 'SkuName', 'BastionHostSkuName']]]:
         """
         Name of the pricing tier.
         """
         return pulumi.get(self, "name")
 
     @name.setter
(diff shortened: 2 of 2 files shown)


provider/pkg/gen/schema.go
```diff
@@ -827,14 +827,18 @@ func (g *packageGenerator) formatFunctionDescription(op *spec.Operation, respons
...
-func (g *packageGenerator) formatDescription(desc string, info *spec.Info) string {
+func (g *packageGenerator) formatDescription(desc string, info *spec.Info, resourceSpec *openapi.ResourceSpec) string {
+	description := desc
 	if g.apiVersion == "" {
-		return fmt.Sprintf("%s\nAPI Version: %s.", desc, info.Version)
+		description = fmt.Sprintf("%s\nAPI Version: %s.", description, info.Version)
 	}
-	return desc
+	if resourceSpec != nil && resourceSpec.PreviousVersion != "" {
+		description = fmt.Sprintf("%s\nPrevious API Version: %s. See https://github.com/pulumi/pulumi-azure-native/discussions/TODO for information on migrating from v1 to v2 of the provider.", description, resourceSpec.PreviousVersion)
```
thomas11, line 839: Looking ahead to v3 and beyond, it would be good to say which major version we mean by previous.

provider/pkg/gen/merging_test.go
```diff
@@ -0,0 +1,323 @@
...
+				}),
+				t2: objectComplexType(map[string]schema.PropertySpec{
+					"foo": {
+						TypeSpec: oneOfType(refType("#/definitions/FooB"), stringType()),
+					},
+				}),
+				isOutput: false,
+			},
+			want: objectComplexType(map[string]schema.PropertySpec{
+				"foo": stringPropertySpec(),
```
thomas11, line 196: While I understand the idea behind this collapsing, doesn't this break user programs that use FooA or FooB?

thomas11:
> TODO: even though the oneOf is intersected, the discriminator is not ... does this matter?

Not sure what you mean by that, can you elaborate?

danielrbradley:
Notes from review call:
- While intersection of input types creates a safer set it could make some funcitonality unreachable. We should opt for union everywhere.
- We need to also change the discriminator when changing the oneOf definition.
- If this does break the existing v1 schema, we'll introduce a temprary feature switch for the type merge function via an environment variable which can be removed after the v2 release.

thomas11 (approved):

PR #2348, opened 2023-03-31
Remove versions removed in most recent release

danielrbradley:
Adding these causes nodejs to be unreleasable so we have to remove these. This is unlikely to cause any user impact as they were only released for users of dotnet, python, java and go - if they updated within the last 24 hours and only if they start using the specified new explicit, undocumented versions.

Keeping signification updates from:
- AppConfiguration: 2023-03-01
- Communication: 2023-03-31
- Consumption: 2023-03-01


danielrbradley:
Deploy test to see if we're under the limit for nodejs: https://github.com/pulumi/pulumi-azure-native/actions/runs/4577367870

kpitzen (approved):
Pending linked deploy test

thomas11 (approved):

PR #2358, opened 2023-04-06
Refactor generation for testing

danielrbradley:
Make schema generation testable - needed to diagnose memory issues introduced by intermediate bug.
- Pass in all paths to load files from rather than assuming our working directory is always the repo root.
- Work around issue in open api schema loading which truncated leading parent navigation in the root path due to being treated as a URL rather than only a file path.

### DIFF
--- a/provider/pkg/versioning/gen_bench_test.go
+++ b/provider/pkg/versioning/gen_bench_test.go
@@ -8,34 +8,32 @@ import (
 	"github.com/pulumi/pulumi-azure-native/provider/pkg/openapi"
 )
 
+// BenchmarkGen benchmarks the generation of the Pulumi schema.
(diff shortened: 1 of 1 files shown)


provider/pkg/gen/utilities.go
```diff
@@ -7,10 +7,14 @@ import (
 	"unicode"
 )
 
+var legalIdentifierReplacer *strings.Replacer
+
 // MakeLegalIdentifier removes characters that are not allowed in identifiers.
 func MakeLegalIdentifier(s string) string {
-	replacer := strings.NewReplacer("-", "", "[", "", "]", "")
-	return replacer.Replace(s)
+	if legalIdentifierReplacer == nil {
```
thomas11, line 14: why not just `var legalIdentifierReplacer = strings.NewReplacer("-", "", "[", "", "]", "")` outside of the func?

thomas11 (approved):

PR #2361, opened 2023-04-08
Init static regex only once in openapi resolver

thomas11:
Removes the "regex compile" allocations in the lower right.

Screenshot 2023-04-07 at 21 16 18


rquitales (approved):
LGTM with a minor non-blocking suggestion.

PR #2369, opened 2023-04-11
Remove long-deprecated base variants

thomas11:
See the discussion on the changes below.


danielrbradley (commented):
This looks great - saves ~424 files in the nodejs SDK by my count.

My only question on reviewing the SDK changes is the Network changes - it seems to just remove deprecation notices but not remove any resources. It also adds some types. I haven't quite worked out how this relates back to the change in the schema.go.

Once we can reason about that I'm happy for this to go

thomas11:
> This looks great - saves ~424 files in the nodejs SDK by my count.
> 
> My only question on reviewing the SDK changes is the Network changes - it seems to just remove deprecation notices but not remove any resources. It also adds some types. I haven't quite worked out how this relates back to the change in the schema.go.

I'm not quite sure about it yet. Fortunately, it seems to concern only the `UserRule` and `AdminRule` resources. Before this PR, there were only `DefaultUserRule` and `DefaultAdminRule`. Now there are also general `UserRule` and `AdminRule` resources. So it would make sense that they need input args, represented by the new type `AddressPrefixItemArgs`. I don't understand yet why these general rule types didn't exist before.

mikhailshilkov:
I checked all removed resources in Metabase and none of them are used, which confirms they are actually impossible to use and can be safely removed.

It sounds like we need to figure out what to do with UserRule and AdminRule and then we can ship.

thomas11:
> I checked all removed resources in Metabase and none of them are used, which confirms they are actually impossible to use and can be safely removed.
> 
> It sounds like we need to figure out what to do with UserRule and AdminRule and then we can ship.

@mikhailshilkov I stepped through this in the debugger and see what happened. On the master branch, 'genResources' correctly generates resource `AdminRule`. But then, at the end of the method, it generates the main/base resource, which has the same `typeName` ([code](https://github.com/pulumi/pulumi-azure-native/blob/master/provider/pkg/gen/schema.go#L438)).
```go
	mainResource := &resourceVariant{
		ResourceSpec: resource,
		typeName:     typeName,
	}
```
Therefore, the correct `AdminRule` generated just previously is overwritten with a version that is incorrect (lacks body properties, among other things).

This can easily be verified by changing the above snippet to say `typeName:     typeName + "2"`. The generated metadata will then have the correct `AdminRule` and `UserRule`, plus the incorrect `AdminRule2` and `UserRule2`.

This PR removes the `mainResource`, so the correct resources are not overwritten anymore.

I'm not quite seeing what's special about these resources that makes them vulnerable to this issue. `AdminRule` and `DefaultAdminRule` are both instances of `BaseAdminRule` (via `allOf: [{ $ref: ... }]`). They have the usual and correct-looking `x-ms-discriminator-value`.

mikhailshilkov (approved):
LGTM :+1:

PR #2376, opened 2023-04-12
Add schema tools to makefile

danielrbradley:
- Add `make schema_squeeze` so we can start experimenting with schema-tools for removing old versions.
- Separate removals between v1 and v2 (solves at least part of https://github.com/pulumi/pulumi-azure-native/issues/2359)
- Add `make explode_schema_v2` to be able to navigate the new schema much easier for review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,6 +153,12 @@ test: build install_sdks
 schema_squeeze: bin/schema-tools bin/schema-full.json
 	./bin/schema-tools squeeze -s bin/schema-full.json --out bin/v1-squeeze.json
 
+.PHONY: explode_schema_v2
+explode_schema_v2: bin/v2/schema-full.json
+	rm -rf bin/schema_v2
+	mkdir -p bin/schema_v2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -149,6 +149,10 @@ test: export PULUMI_LOCAL_NUGET=$(WORKING_DIR)/nuget
 test: build install_sdks
 	cd examples && go test -v -tags=all -timeout 2h
 
+.PHONY: schema_squeeze
+schema_squeeze: bin/schema-tools
+	./bin/schema-tools squeeze -s bin/v2/schema-full.json --out versions/squeeze.json
```
thomas11, line 154: I think we should version all files, so something like `v2-squeeze.json`.

thomas11 (approved):
Left a change suggestion but it's not a dealbreaker, approving so I don't block.

Makefile
danielrbradley: Yes, that makes sense ... though I'm not actually sure if we should be running this from the v1 or v2 schema - guess it should be whichever schema has all explicit versions in it?

thomas11 (approved):

PR #2378, opened 2023-04-13
Improve Refresh accuracy for dictionary and array values

### LINKED ISSUE #2356: Extraneous diff for refreshed complex properties
### What happened?

After a `refresh`, a lot of (default?) values are read and stored in the state for Azure resources. When running `pulumi up` afterwards, a big diff shows up. For example, this coming from a `WindowsVirtualMachine` on the attached data disks:

```
        ~ azure-native:compute:VirtualMachine: (update)
            [id=/subscriptions/<redacted>/resourceGroups/ringo-import-test/providers/Microsoft.Compute/virtualMachines/vmtest]
            [urn=urn:pulumi:dev::pulumiRefreshIssue::<redacted>:component:virtualMachineComponent$azure-native:compute:VirtualMachine::smc1]
            [provider=urn:pulumi:dev::pulumiRefreshIssue::pulumi:providers:azure-native::default_1_98_1::243e4e43-712b-433f-bd27-a8b60cf2bf2e]
          ~ storageProfile: {
              ~ dataDisks: [
                  ~ [0]: {
                          - caching     : "None"
                          - deleteOption: "Detach"
                          - diskSizeGB  : 1024
                          ~ managedDisk : {
                              - storageAccountType: "Premium_LRS"
                            }
                          - name        : "dataDisk1"
                          - toBeDetached: false
                        }
                ]
              ~ osDisk   : {
                  ~ caching     : "ReadOnly" => "ReadWrite"
                  - deleteOption: "Detach"
                  ~ managedDisk : {
                      - id: "/subscriptions/<redacted>/resourceGroups/ringo-import-test/providers/Microsoft.Compute/disks/vmtest01_disk1_<redacted>"
                    }
                }
            }
```

Properties not managed by the Pulumi program were stored in the state during `refresh`, and are now removed again during `up`.

### Expected Behavior

From @danielrbradley:

> Refresh should never cause diffs. Looks like there’s a possible bug for diffs of nested properties (maybe only inside arrays).

### Steps to reproduce

* Create a `WindowsVirtualMachine` resource and attach a managed disk.
* Run `pulumi refresh`. You see additonal details being imported besides the one managed in the program
* Run `pulumi up`. You see the same details now proposed as the properties being deleted again from the state.

### Output of `pulumi about`

CLI
Version 3.59.1
Go Version go1.20.2
Go Compiler gc

Plugins
NAME VERSION
azure 5.38.0
azure-native 1.98.1
nodejs unknown

Host
OS darwin
Version 13.3
Arch arm64

This project is written in nodejs: executable='/Users/ringods/.volta/bin/node' version='v16.19.1'

Current Stack: team-ce/pulumiRefreshIssue/dev

TYPE URN
removed the resource list as it came from a customer project

Found no pending operations associated with team-ce/dev

Backend
Name pulumi.com
URL https://app.pulumi.com/ringods
User ringods
Organizations ringods, team-ce

Dependencies:
NAME VERSION
@types/node 16.18.21
@pulumi/azure-native 1.98.1
@pulumi/azure 5.38.0
@pulumi/pulumi 3.59.1

### Additional context

_No response_

### Contributing

mikhailshilkov:
Fix #2356

Azure Native has a peculiar way to update inputs when a `pulumi refresh` calls the provider's Read method. It's impossible to reads "inputs" from Azure, so we use the following algorithm to refresh existing inputs:
1. Read the current properties of a resource, it's "new outputs".
2. Diff "old outputs" from the state with these new outputs from Azure.
3. See which properties changed/added/removed.
4. Apply only those changed properties to existing inputs from state to produce new inputs.

This is intended to minimize the noise and diffs in `refresh` and subsequent `up` commands, which #2356 illustrates.

The problem is that the step (4) is currently operating on top-level properties only. For map- or array-properties that means that any change inside them will lead to us copying the entire content of that value to new inputs. This is the cause for #2356.

This PR changes the step (4) "apply diff" to be a recursive comparison across maps and arrays. We compare property-by-property and element-by-element and only copy the granular changes that we identify.

I extended our existing test to cover those scenarios. I also tested https://github.com/pulumi/pulumi-azure-native/issues/2356#issuecomment-1503672473 and got a nice diff on the last step of Ringo's repro:

image

cc @ringods if you want to take a look or give this branch a try locally (`make provider` then copy `./bin/pulumi-resource-azure-native` somewhere on your path).

### DIFF
--- a/provider/pkg/provider/diff.go
+++ b/provider/pkg/provider/diff.go
@@ -145,6 +145,12 @@ func (d *differ) calculateValueDiff(v *resource.ValueDiff, diffBase, replaceBase
 }
 
 // applyDiff produces a new map as a merge of a calculated diff into an existing map of values.
+// Note that the use case for it is rather specific to the particular Read implementation:
+// - values are inputs currently stored in state before Read was called
+// - diff is a calculated difference between old and new _outputs_
+// So, diff values are independent of the values map. Practically, diff values may contain
+// more properties in object maps, and that doesn't necessarily mean we need to copy all of
+// them. Instead, we want to copy only the changed properties, and do so recursively.
 func applyDiff(values resource.PropertyMap, diff *resource.ObjectDiff) resource.PropertyMap {
 	if diff == nil {
 		return values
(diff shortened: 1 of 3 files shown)


justinvp:
I'm curious (having not looked closely at this): do the other native providers need a similar treatment, or is this specific to Azure Native.

provider/pkg/provider/diff.go
```diff
@@ -159,12 +165,50 @@ func applyDiff(values resource.PropertyMap, diff *resource.ObjectDiff) resource.
...
+	if baseValue.IsArray() && oldValue.IsArray() && newValue.IsArray() {
+		baseValueArray := baseValue.ArrayValue()
+		oldValueArray := oldValue.ArrayValue()
+		newValueArray := newValue.ArrayValue()
+		length := len(newValueArray)
+		result := make([]resource.PropertyValue, length)
+
+		for i, el := range newValueArray {
+			if i >= len(baseValueArray) || i >= len(oldValueArray) {
+				// A new element was added, copy as-is.
```
thomas11, line 198: Can new values only be added at the end, not elsewhere, shifting the other values?

thomas11 (approved):

provider/pkg/provider/diff.go
mikhailshilkov: Good question - I was thinking about this as well.
I guess they could be added anywhere, but I don't know how to handle this situation other than diffing with the assumption that a new element got added to the end. We'll show large differences in this case, which I kind of fine, I hope.

mikhailshilkov:
> I'm curious (having not looked closely at this): do the other native providers need a similar treatment, or is this specific to Azure Native.

@justinvp Great question. Yeah, they do. It looks like we should refactor this code to a common reusable place. Any ideas what would be a natural package to host this? I assume - somewhere in pu/pu?

provider/pkg/provider/diff.go
thomas11: Sounds good. If we ever run into this problem in the real world, we can study fancier diff algorithms.

PR #2391, opened 2023-04-20
Squeeze the V2 schema

thomas11:
- Renaming a bunch of things in the schema code for consistent naming
  - Config: manually written curation and generation configuration
  - Spec: mapping of providers, optionally single resources, to api versions; generated from API and config
  - Lock: as before, the final result; a mapping of each resource to an API version
- Allow common make targets like `provider`, `schema` to produce and use V1 or V2 of the schema, via `make ... VERSION_DIR=v2/`.
  - `VERSION=v2` would be prettier but make can't deal with the extra slash in case the version is empty


danielrbradley (commented):
Looks like good progress. On-board with all your naming improvements 👍 

Will work through the flow of creating the various files in a little more detail next to make sure I'm following precisely.

PR #2405, opened 2023-04-25
Squeeze the V2 schema and allow building V2 SDKs

thomas11:
- Rename files and code for clarity, remove some unused code
- Move squeezing to just before schema generation
- Preserve resources in v1-lock when squeezing
- Groundwork for generating the v2 schema, but not plumbed all the way through yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -342,7 +342,7 @@ export FAKE_MODULE
 .make/build_nodejs: VERSION_JS = $(shell bin/pulumictl convert-version -l javascript -v "$(VERSION_GENERIC)")
 .make/build_nodejs: bin/pulumictl .make/nodejs_yarn_install
 	cd sdk/nodejs/ && \
-		NODE_OPTIONS=--max-old-space-size=8192 yarn run tsc --diagnostics --incremental && \
+		NODE_OPTIONS=--max-old-space-size=12288 yarn run tsc --diagnostics --incremental && \
 		cp ../../README.md ../../LICENSE package.json yarn.lock ./bin/ && \
 		mkdir -p bin/scripts && cp scripts/install-pulumi-plugin.js bin/scripts && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,6 +24,8 @@ GOEXE ?= $(shell go env GOEXE)
 LOCAL_PROVIDER_FILENAME := $(PROVIDER)$(GOEXE)
 LOCAL_ARM2PULUMI_FILENAME := arm2pulumi$(GOEXE)
 
+VERSION_DIR ?= 
+
```
danielrbradley, line 28: I think this is the only part of this PR I'm not sure about yet. It feels like it's adding a bit of complexity - but is only needed for a temporary transition period. 

Do we need the ability to generate the SDKs on the same branch as v1?

Could we just do the v2 prerelease from a branch which only contains a minimal change to switch which schema we use? Then can rebase/re-merge the v2 prerelease branch as we continue making changes on master?

danielrbradley (approved):
Approving for the schema work to go in. We'll pull out the 2 commits for the SDK building into a separate follow-on PR.

provider/pkg/versioning/gen.go
```diff
@@ -98,12 +133,17 @@ func ReadVersionSources(rootDir string) (VersionSources, error) {
 		return VersionSources{}, err
 	}
 
-	v1Spec, err := ReadDefaultConfig(path.Join(rootDir, "versions", "v1-spec.yaml"))
+	knownExplicitResources, err := ReadRequiredExplicitResources(path.Join(rootDir, "versions", "required-explicit-resources.txt"))
```
thomas11, line 136: Where does this file come from @danielrbradley?

thomas11: Ah, Metabase. Not too obvious to a reader not familiar with this area.

PR #2410, opened 2023-04-27
Squeeze the V2 schema and allow building V2 SDKs (#2405)

danielrbradley:
- Rename files and code for clarity, remove some unused code
- Move squeezing to just before schema generation
- Preserve resources in v1-lock when squeezing
- Groundwork for generating the v2 schema, but not plumbed all the way through yet.

---------

Cherry picked squashed merge of #2405 from v2.0-pre branch

### DIFF
--- a/provider/pkg/versioning/gen.go
+++ b/provider/pkg/versioning/gen.go
@@ -27,22 +27,15 @@ type VersionMetadata struct {
 	V2ResourcesToRemove           Squeeze
 }
 
-type MajorVersion int
-
-const (
-	V1 MajorVersion = 1
-	V2 MajorVersion = 2
-)
-
-func GenerateVersionMetadata(rootDir string, majorVersion MajorVersion, providers openapi.AzureProviders) (VersionMetadata, error) {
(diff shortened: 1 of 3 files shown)


provider/pkg/versioning/gen.go
```diff
@@ -1,95 +1,130 @@
...
 	versionSources, err := ReadVersionSources(rootDir)
 	if err != nil {
 		return VersionMetadata{}, err
 	}
-	return calculateVersionMetadata(versionSources, providers)
+	return calculateVersionMetadata(majorVersion, versionSources, providers)
 }
 
-func calculateVersionMetadata(versionSources VersionSources, providers map[string]map[string]openapi.VersionResources) (VersionMetadata, error) {
+func calculateVersionMetadata(majorVersion MajorVersion, versionSources VersionSources, providers map[string]map[string]openapi.VersionResources) (VersionMetadata, error) {
```
thomas11, line 45: I just realized that the new parameter `majorVersion` is actually unused now. We might want to remove it.

danielrbradley: Done

thomas11 (approved):

PR #2418, opened 2023-04-28
Generate V2 SDKs

danielrbradley:
Attempted to generate Java from same full schema but still fails:
1. Runs out of memory using the default 4gb limit. This can be temporaily worked around locally.
2. `Task :jar FAILED ` ... `archive contains more than 65535 entries.` - this does not seem to be resolvable.

Attempted to add examples back in to the full schema - so we can use the same schema for SDKs and documentation. This currently fails due to Go Program Generation getting stuck in an infinite loop: https://github.com/pulumi/pulumi/issues/12786

Skipping over the bugs, we can generate all examples into the SDKs - which approximately doubles the schema size: https://github.com/pulumi/pulumi-azure-native/compare/v2-sdk-examples
However, the example generation is still not stable so changes the output every time.


thomas11 (approved):
LGTM except for Makefile cleanups

provider/pkg/openapi/defaults/defaultResourcesState.go
```diff
@@ -185,23 +204,27 @@ func containsNonEmptyCollections(value map[string]interface{}) bool {
...
+	// State is the default state of a resource.
+	State map[string]interface{}
+	// SkipDelete is true if the PUT operation should be skipped during the delete step.
+	SkipDelete bool
+	// HasNonEmptyCollections is true if any property of the default state is a non-empty collection.
 	HasNonEmptyCollections bool
 }
 
-func GetDefaultResourceState(path string) DefaultResourceState {
+func GetDefaultResourceState(path string) *DefaultResourceState {
```
thomas11, line 216: [nit] for additional safety, consider documenting that return value can be nil or using a `value, ok` tuple as return value

danielrbradley: Yup, good thought ... this was from the merge with master so not directly in this branch. Thanks for feedback though

PR #2421, opened 2023-04-28
Cherry-pick Remove unused version parameter

thomas11:
Cherry-pick of b68d1c50 from #2410 to get `master` and `v2.0-pre` in sync


danielrbradley (approved):
I think the schema change warnings are just because we're comparing to master maybe?

PR #2428, opened 2023-05-03
Simplify schema gen process

danielrbradley:
Stacked on top of:
- #2438 
- #2439 

Highlights:
- Use single method with arguments to differentiate between v1/v2 and docs/full.
- Simplify gen program to remove option to generate multiple targets at once.


provider/pkg/versioning/build_schema.go
```diff
@@ -0,0 +1,101 @@
...
+	Version                 string
+}
+
+type BuildSchemaResult struct {
+	PackageSpec schema.PackageSpec
+	Metadata    resources.AzureAPIMetadata
+	Version     VersionMetadata
+}
+
+func BuildSchema(args BuildSchemaArgs) (*BuildSchemaResult, error) {
```
thomas11, line 36: There are three branches based on version in this method and all of them are about calling an `openapi` method. This could probably be refactored a bit further. Totally optional, though, it's already much better than it was!

provider/cmd/pulumi-gen-azure-native/main.go
```diff
@@ -210,23 +172,20 @@ func main() {
 		if err != nil {
 			panic(err)
 		}
-	}
-
-	for _, language := range basicLanguages {
```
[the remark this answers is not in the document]
danielrbradley: It's now just the makefile calling `pulumi package gen-sdk ...`

provider/pkg/versioning/build_schema.go
danielrbradley: Yeah, definitely further we can take this and tidy up more ... keen to get this integrated as it unblocks the v2 schema & SDK generation parts.

thomas11 (approved):

thomas11:
This PR is quite large now. Can we split off the p/p upgrade which should be independent of schema changes?

PR #2430, opened 2023-05-04
[arm2pulumi] - Adds input validation for expected languages to render

kpitzen:
- Adds a set of supported languages in pkg/arm2pulumi/renderer.go
- Validates all provided languages in cmd/arm2pulumi/main.go
- Provides dynamic list of possible languages in the primary prompt help

Closes https://github.com/pulumi/arm2pulumi/issues/32

### DIFF
--- a/provider/pkg/arm2pulumi/renderer.go
+++ b/provider/pkg/arm2pulumi/renderer.go
@@ -30,6 +30,20 @@ import (
 	"github.com/sourcegraph/jsonx"
 )
 
+var SupportedLanguages = map[string]interface{}{
+	"nodejs": true,
+	"python": true,
+	"dotnet": true,
+	"go":     true,
+	"java":   true,
+	"yaml":   true,
+}
+
+func IsSupportedLanguage(lang string) bool {
+	_, ok := SupportedLanguages[lang]
+	return ok
+}
+
 type Renderer struct {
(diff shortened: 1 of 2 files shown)


kpitzen:
Example output:
image

provider/pkg/arm2pulumi/renderer.go
```diff
@@ -30,6 +30,22 @@ import (
...
+	"python": true,
+	"dotnet": true,
+	"go":     true,
+	"java":   true,
+	"yaml":   true,
+}
+
+func IsSupportedLanguage(lang string) bool {
+	if _, ok := SupportedLanguages[lang]; ok {
+		return true
```
thomas11, line 44: [nit] can just return `ok`

provider/pkg/arm2pulumi/renderer.go
```diff
@@ -30,6 +30,22 @@ import (
 	"github.com/sourcegraph/jsonx"
 )
 
+var SupportedLanguages = map[string]interface{}{
```
thomas11, line 33: Just a note: a bit unfortunate that we have to repeat (and later update) the list here when we already import p/p. I just looked around briefly but it doesn't seem we have any exported, official set of languages there.

thomas11 (approved):

PR #2439, opened 2023-05-08
Use Pulumi CLI for SDK generation

danielrbradley:
Stacked on top of:
- #2438 

Allows us to:
- Upgrade SDKs independent to pulumi/pkg Go module.
- Remove need to maintain custom gen program features.


danielrbradley:
Merged as part of https://github.com/pulumi/pulumi-azure-native/commit/f59975b59215770819e3834960dc2fe6477463df

PR #2441, opened 2023-05-09
Remove some unused versioning code

thomas11:
Noticed by chance, probably a follow-up to #2428


danielrbradley (approved):
Looks good, but think this can go straight into `master` instead of `v2.0-pre`.

PR #2443, opened 2023-05-10
Rework git uncommitted files check

danielrbradley:
- Remove dependency on ci-scripts.
- Use `git status -s` for check instead of `git update-index -q --refresh && git diff-files --quiet`.

### DIFF
--- a/.github/actions/check-worktree-clean/action.yml
+++ b/.github/actions/check-worktree-clean/action.yml
@@ -4,15 +4,17 @@ description: Fail if there's uncommitted changes in git worktree.
 runs:
   using: "composite"
   steps:
-      - name: Checkout Scripts Repo
-        uses: actions/checkout@v3
-        with:
-          path: ci-scripts
-          repository: pulumi/scripts
-
-      - name: Check worktree clean
+      - run: git status
         shell: bash
-        run: ./ci-scripts/ci/check-worktree-is-clean
 
-      - run: git status --porcelain
+      - name: Check git status
+        id: git-status
         shell: bash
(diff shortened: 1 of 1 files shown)


.github/actions/check-worktree-clean/action.yml
```diff
@@ -4,15 +4,17 @@ description: Fail if there's uncommitted changes in git worktree.
...
-          repository: pulumi/scripts
+      - run: git status
+        shell: bash
 
-      - name: Check worktree clean
+      - name: Check git status
+        id: git-status
         shell: bash
-        run: ./ci-scripts/ci/check-worktree-is-clean
+        run: echo "STATUS=$(git status -s)" >> $GITHUB_OUTPUT
```
thomas11, line 13: This seems to have been the only usage. Can we remove the script in this PR, then?

danielrbradley: This script doesn't actually exist here - it's checked out from a different repo (`pulumi/scripts`) in the step before 🙃

danielrbradley: Here's what the existing script contained: 

```bash
set -o nounset -o errexit -o pipefail

# ensure the index is up to date, in CI we were seeing some cases
# where git diff-files would show the entire tree was unmerged
# and this addresses that.
git update-index -q --refresh

if ! git diff-files --quiet; then
    >&2 echo "error: working tree is not clean, aborting!"
    git status
    git diff
    exit 1
fi
```

https://github.com/pulumi/scripts/blob/b6ee1692954d2b39e7a22b0db80b8cc4b735c101/ci/check-worktree-is-clean

thomas11 (approved):

PR #2456, opened 2023-05-16
Skip default delete

danielrbradley:
For resources without DELETE operations defined (only PUT) and no obvious "default" state to reset to, add the option for delete to be a no-op. 

This is required for Sql database TransparentDataEncryption which requires different properties to be set for different API versions. This could be solved by introducing version awareness, but that would be a larger piece of work.

We model skipping the delete by setting the "default body" to nil.

This includes a refactor to disentwine path normalisation and default resource state from the main discovery process. These are moved into their own modules so only public members can be interacted with - to avoid interacting with the wrong internal parts.

### DIFF
--- a/provider/pkg/openapi/defaults/defaultResourcesState.go
+++ b/provider/pkg/openapi/defaults/defaultResourcesState.go
@@ -3,6 +3,7 @@
 package defaults
 
 import (
+	"fmt"
 	"reflect"
 
 	"github.com/pulumi/pulumi-azure-native/provider/pkg/openapi/paths"
@@ -81,9 +82,6 @@ var defaultResourcesStateRaw = map[string]map[string]interface{}{
 	"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Sql/servers/{serverName}/databases/{databaseName}/securityAlertPolicies/{securityAlertPolicyName}": {
 		"state": "Disabled",
 	},
-	"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Sql/servers/{serverName}/databases/{databaseName}/transparentDataEncryption/{transparentDataEncryptionName}": {
-		"status": "Disabled",
-	},
 	"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Sql/servers/{serverName}/auditingSettings/{blobAuditingPolicyName}": {
 		"state": "Disabled",
 	},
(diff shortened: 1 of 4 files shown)


provider/pkg/provider/provider.go
```diff
@@ -1260,8 +1260,8 @@ func (k *azureNativeProvider) Delete(ctx context.Context, req *rpc.DeleteRequest
 	case res.Singleton:
 		// Singleton resources can't be deleted (or created), set them to the default state.
 		for _, param := range res.PutParameters {
-			if param.Location == "body" && res.DefaultBody != nil {
-				requestBody := k.converter.SdkPropertiesToRequestBody(param.Body.Properties, *res.DefaultBody)
+			if param.Location == "body" && res.SkipDelete {
```
thomas11, line 1263: The comment for `skipDelete` says 
> true for resources that don't have a delete operation and _no sensible default_ to set its state back to

but here we set the default if the property is true?

danielrbradley: Oops yup that's missing a `!`

provider/pkg/openapi/defaults/defaultResourcesState.go
```diff
@@ -179,3 +204,15 @@ func containsNonEmptyCollections(value map[string]interface{}) bool {
 	}
 	return false
 }
+
+type DefaultResourceState struct {
+	HasDefault             bool
```
thomas11, line 209: A default state that has no default (`!HasDefault`) reads a bit odd. Maybe `HasDefaultBody` or `HasDefaultProperties`?

thomas11 (approved):
LGTM, although I'd be more confident with some test coverage.

provider/pkg/openapi/defaults/defaultResourcesState.go
danielrbradley: Hmm, this was to indicate if we actually have any form of delete behaviour specified in the map. Not sure on the most "ideomatic go" here ... a second return for "ok" or rename this to `Found` perhaps ... or just make the whole return a pointer and return nil if not found 🤔

thomas11: I think the structure is ok, I'd just consider a different name for `HasDefault`.

provider/pkg/openapi/defaults/defaultResourcesState_test.go
```diff
@@ -0,0 +1,50 @@
...
+func TestNonEmptyCollection(t *testing.T) {
+	actual := GetDefaultResourceState("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Insights/components/{resourceName}/currentbillingfeatures")
+	expected := DefaultResourceState{
+		State: map[string]interface{}{
+			"currentBillingFeatures": []string{"Basic"},
+			"dataVolumeCap":          map[string]string{},
+		},
+		HasNonEmptyCollections: true,
+	}
+	assert.Equal(t, &expected, actual)
```
mikhailshilkov, line 49: Could you please explain what we are testing here? `GetDefaultResourceState` does two things: 

- Calls `NormalizePath`
- Calls a map lookup and returns the value or default

What do we get by having 5 tests for it? It feels like we are testing the contents of `defaultResourcesStateNormalized`?

PR #2460, opened 2023-05-18
Fix Go module versions

danielrbradley:
Roughly following the [provider marjor version update guide](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/tf-provider-major-version-update.md) but adjusting for the fact we're not a bridged provider.

- When generating Go SDK modules, update major version within their module path.
- Build locally as v2 via the default makefile version.
- Bump provider module version to v2 (is this actually required given we don't push any `provider/vX` tags?)

Not bumping examples module version as has no benefit.


thomas11 (approved):
LGTM!

PR #2466, opened 2023-05-23
Refactor genProperties

danielrbradley:
Bring generation of schema and metadata for properties into the same place so we can modify both easily in sync e.g. for modifying external representations or calculating defaults internally rather than through the schema.

1. Re-organise the code to not live in a single 2k+ LoC file.
2. Group property specific functions together.
3. Group type specific functions together.
4. Extract `genApiProperty` from plural `genProperties` .
5. Refactor signature of `genProperty` & update call sites.
6. Simplify api property construction.
7. Combine `genApiProperty` into `genProperty`.

### DIFF
--- a/provider/pkg/gen/properties.go
+++ b/provider/pkg/gen/properties.go
@@ -249,41 +249,35 @@ func (m *moduleGenerator) genProperty(name string, schema *spec.Schema, context 
 		TypeSpec:    *typeSpec,
 	}
 
-	apiProperty := m.genApiProperty(name, &propertySpec, resolvedProperty, isOutput, isType)
-
-	return &propertySpec, &apiProperty, nil
-}
-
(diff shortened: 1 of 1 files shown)


thomas11 (approved):
This is great, thank you!

Going forward beyond this PR, IMO some of the naming could use some love, too. Mainly to reflect the difference between schema and metadata more clearly.

kpitzen:
> This is great, thank you!
> 
> Going forward beyond this PR, IMO some of the naming could use some love, too. Mainly to reflect the difference between schema and metadata more clearly.

Part of me agrees, but another, different, part of me wishes we had no distinction between schema and metadata at all. To me, schemata _are_ metadata, and it would be nice if instead of two huge files with a separation of these concepts, we instead had something along the lines of:
- Schema manifest (list of expected files/resources/services etc which comprises the full schema)
- Exploded schema which partitions the provider by service/resource, and which includes the relevant metadata for each resource contained within, each partitioned section being referenced by the above manifest

I understand this would require some pretty major changes to p/p, but it feels like a much cleaner way to handle this whole thing.

PR #2473, opened 2023-05-25
Prepare v2 pre-release 2

danielrbradley:
Fix SDK generation:
- Set root package name - use pre-release of pulumi/pkg to fix bug
  - https://github.com/pulumi/pulumi/pull/13039
- Fix module import aliases to include v2 suffix.
- Remove the version from the module replacement as it's not required.
- Fix the stripping of the replacements in the makefile to work with v2.
- Fix the import base path in the original schema and remove override during the split package emit.

Fix Go tests:
- Fix up go tests to be able to run locally - allows us to run tests by just clicking in a UI.
- Use LocalProviders in tests to point to the provider binary instead of requiring it to be on the path.
- Generate replacements based on existing go mod requirements.
- Set PULUMI_GO_DEP_ROOT automatically if not already set.
- Set Dir in base options.
- Add more targets for running specific tests through the makefile.

We can't set plugins/providers in examples because ProgramTest LocalProviders doesn't remove conflicting entries so then there's duplicates and the existing entry no longer resolves because the relative path is incorrect after being copied to a temp directory.

Disable go work when running through make for consistency with CI.

Override dev version for v2:
- https://github.com/pulumi/pulumi-azure-native/issues/2482

Write version files before schema and metadata, otherwise make marks the schema as out of date on every build.

Skip failing arm2pulumi test
- https://github.com/pulumi/arm2pulumi/issues/33


Makefile
```diff
@@ -23,6 +23,7 @@ GOARCH ?= $(shell go env GOARCH)
 GOEXE ?= $(shell go env GOEXE)
 LOCAL_PROVIDER_FILENAME := $(PROVIDER)$(GOEXE)
 LOCAL_ARM2PULUMI_FILENAME := arm2pulumi$(GOEXE)
+export GOWORK := off
```
kpitzen, line 26: SHOULD be a noop in ci, but is this intentional for everyone else?

thomas11: IMO, when in doubt, consistency wins.

thomas11 (approved):

Makefile
thomas11: Although, it just occurred to me that a remark in CONTRIBUTING could be helpful to prevent confusion for people who have workspaces set up.

mikhailshilkov (changes requested):
Please add a PR description according to our [standards](https://docs.google.com/document/d/1QG5f95vWtBNI7EuQ3kgBlyFtMJKSCeN0AGqYPosV9lg/edit). This helps others review the PR now and later down the line. Thank you!

.github/workflows/dev-version.yml
```diff
@@ -24,6 +24,8 @@ jobs:
           LAST_VERSION=$(echo "${GH_VERSION}" | cut -f1)
           # Use semver NPM CLI tool to find next minimal version
           NEXT_VERSION=$(npx -y semver --increment patch "${LAST_VERSION}")
+          # TODO: Remove after v2 release
```
mikhailshilkov, line 27: Let's make sure we open an issue for every TODO in code, and reference that issue from the TODO

> Added TODOs must include links to issue numbers in TODO[pulumi/pulumi#7973] format.

PR #2478, opened 2023-05-29
Allow relative IDs in self-referencing resources

### LINKED ISSUE #631: Add shorthand reference field when self-references are required
## Affected feature
Quite a lot of azure.networking defenitions require subresources to reference other subresources of the resource one is trying to create. In my instance, ApplicationGateway resource requires the defenition of `requestRoutingRules`, which contain multiple references to other subresources of the ApplicationGateway, among which is the `httpListener`, which in turn requires references to `frontendIPConfigurations` and `frontendPorts`. 

Simply put - for resource A, subresource A.x references A.y which references A.v and A.w.  

While this is an issue with the underlying ARM templates themselves, the feature I would request to resolve it would be the introduction of a `name` (or `nameRef` or `subresourceName`) field which is outside the specification, xor with the `id` field, and would internally generate the correct `id:` for use in the arm template.

Take the following example specification of an ApplicationGateway: (apologies for it being huge- but that's part of the point I'm trying to make as well..)
```ts
const appGW = new azure-native.network.ApplicationGateway("myGW", {
    applicationGatewayName: config.require("appGWName"),
    autoscaleConfiguration: {
        maxCapacity: 20,
        minCapacity: 0
    },
    backendAddressPools: [{
        backendAddresses: [],
        name: "defaultbackendpool" // referenced by requestRoutingRules
    }],
    backendHttpSettingsCollection: [{
        name: "defaulthttpsetting", // referenced by requestRoutingRules
        requestTimeout: 60,
        port: 80,
    }],
    enableHttp2: config.requireBoolean("http2binary"),
    frontendIPConfigurations: [{
        name: "publicIp", // referenced by httpListeners
        publicIPAddress: {
            id: pubIp.id // reference to previously created azure-native.network.PublicIPAdress resource
        }
    }],
    frontendPorts: [{
        name: "httpport", // referenced by httpListeners
        port: 80
    }],
    gatewayIPConfigurations: [{}], // omitted for brevity,
    httpListeners: [{
        frontendIPConfiguration: {
            // This is where the problem lies - the id field is resolvable to the id of the resource we are currently generating, but can be "easily" calculated from resources we are already aware of.
            // the improvement would be to introduce a new field, as shown in the example below
            id: myResourceGroup.id.apply(id => `${id}/providers/Microsoft.Network/applicationGateways/${config.require("appGWName")}/frontendIPConfigurations/publicIp`),
        },
        frontendPort: {
            id: myResourceGroup.id.apply(id => `${id}/providers/Microsoft.Network/applicationGateways/${config.require("appGWName")}/frontendPorts/httpport`),
            
        },
        name: "defaulthttplistener", // referenced by requestRoutingRules
        protocol: "Http",
    }],
    location: config.require("location"),
    probes: [],
    requestRoutingRules: [{
        backendAddressPool: {
            id: myResourceGroup.id.apply(id => `${id}/providers/Microsoft.Network/applicationGateways/${config.require("appGWName")}/backendAddressPools/defaultbackendpool`),
        },
        backendHttpSettings: {
            id: myResourceGroup.id.apply(id => `${id}/providers/Microsoft.Network/applicationGateways/${config.require("appGWName")}/backendHttpSettingsCollection/defaulthttpsetting`),
        },
        httpListener: {
            id: myResourceGroup.id.apply(id => `${id}/providers/Microsoft.Network/applicationGateways/${config.require("appGWName")}/httpListeners/defaulthttplistener`),
        },
        name: "appgwrule"
    }],
    resourceGroupName: config.require("rgName"),
    sku: {
        name: "WAF_v2"
        tier: "WAF_v2"
    },
    sslCertificates: [],
    tags: config.requireObject("standardTags"),
    webApplicationFirewallConfiguration: {
        enabled: true,
        firewallMode: "Prevention",
        ruleSetType: "OWASP",
        ruleSetVersion: "3.0"
    },
```

An improved version of this where the subresource references do not need to manually caluculate IDs of subresources that do not yet exist would be like such:

```ts
const appGW = new azure-native.network.ApplicationGateway("myGW", {
    applicationGatewayName: config.require("appGWName"),
    autoscaleConfiguration: {
        maxCapacity: 20,
        minCapacity: 0
    },
    backendAddressPools: [{
        backendAddresses: [],
        name: "defaultbackendpool" // refrenced by requestRoutingRules
    }],
    backendHttpSettingsCollection: [{
        name: "defaulthttpsetting", // referenced by requestRoutingRules
        requestTimeout: 60,
        port: 80,
    }],
    enableHttp2: config.requireBoolean("http2binary"),
    frontendIPConfigurations: [{
        name: "publicIp", // referenced by httpListeners
        publicIPAddress: {
            id: pubIp.id // reference to previously created azure-native.network.PublicIPAdress resource
        }
    }],
    frontendPorts: [{
        name: "httpport", // referenced by httpListeners
        port: 80
    }],
    gatewayIPConfigurations: [{}], // omitted for brevity,
    httpListeners: [{
        frontendIPConfiguration: {
            // This is much better. We could simplify more by not even requireing the "#¤%"# field to be an object, but perhaps note it simply as
            // frontendIPConfigurationName: "publicIp"
            subResourceName: "publicIp"
        },
        frontendPort: {
            subResourceName: "httpport",
        },
        name: "defaulthttplistener", // referenced by requestRoutingRules
        protocol: "Http",
    }],
    location: config.require("location"),
    probes: [],
    requestRoutingRules: [{
        backendAddressPool: {
            subResourceName: "defaultbackendpool"
        },
        backendHttpSettings: {
            subResourceName: "defaulthttpsetting"
        },
        httpListener: {
            subResourceName: "defaulthttplistener"
        },
        name: "appgwrule"
    }],
    resourceGroupName: config.require("rgName"),
    sku: {
        name: "WAF_v2"
        tier: "WAF_v2"
    },
    sslCertificates: [],
    tags: config.requireObject("standardTags"),
    webApplicationFirewallConfiguration: {
        enabled: true,
        firewallMode: "Prevention",
        ruleSetType: "OWASP",
        ruleSetVersion: "3.0"
    },
```

Now I don't personally mind how you do it, but the less I have to calculate by hand (enumerate with string concatenation et cetera) the better. 

**TL:DR;** - Whenever there is an Input requiring an `id` of a SubResource type, create a "helper" field `subResourceName` that allows us to reference these with reusable strings.  Bonus points to eliminate these objects entirely and operate purely on subresource name references (as the `id` resolves to internally anyway).

mikhailshilkov:
Resolves #631

A bunch of Network resources have an unfortunate API design, where a single resource is in fact a whole set of resources that are created as a single operation. Those resources also need to reference each other (e.g. load balancing rule reference front end configurations) and they reference those by absolute resource IDs (/subscription/...). This makes Pulumi programs limited and ugly, with resource names hard-coded and `apply` blocks needed like

```
frontendIPConfiguration: {
    // This is where the problem lies - the id field is resolvable to the id of the resource we are currently generating, but can be "easily" calculated from resources we are already aware of.
    // the improvement would be to introduce a new field, as shown in the example below
    id: myResourceGroup.id.apply(id => `${id}/providers/Microsoft.Network/applicationGateways/${config.require("appGWName")}/frontendIPConfigurations/publicIp`),
},
```

This PR simplifies the UX here without any changes to resource schemas. The `id` property remains and can still accept the same values as currently. In addition, the subresource's `id` property will accept a partial ID that is relative to the parent resource. For the example above, it would be

```
frontendIPConfiguration: {
    // Note that there is no apply here. '$self/' in the beginning means that ID is relative.
    id: "$self/frontendIPConfigurations/publicIp",
}
```

Users still need to get both parts right, but at least those are both strings that can be known statically at the time of program creation.

The implementation injects this resolution into SDK-to-JSON conversion. If it finds a type of `SubResource` with a property `id`, and the value doesn't start with a `/` (as all absolute IDs do), it will prepend it with the parent resource ID. There is a bit of gymnastics of passing the resource ID to the correct place and making sure all the callbacks are wired correctly, but. otherwise the change is fairly straightforward.

Side note: I explored automatic parent-id resolution (e.g. infer `frontendIPConfigurations` above) but it looks tricky. The type “SubResource” is defined once and then referenced by numerous properties from many resources, so it’s impossible to add a single piece of metadata to it that would hold the parent ID. I guess we’d have to store this metadata on the property referencing SubResource, but they are also types that in theory could be reused too… Anyway, to be continued later.

### DIFF
--- a/sdk/dotnet/Compute/Inputs/SubResourceArgs.cs
+++ b/sdk/dotnet/Compute/Inputs/SubResourceArgs.cs
@@ -13,7 +13,10 @@ namespace Pulumi.AzureNative.Compute.Inputs
     public sealed class SubResourceArgs : global::Pulumi.ResourceArgs
     {
         /// <summary>
-        /// Resource Id
+        /// Sub-resource ID. Both absolute resource ID and a relative resource ID are accepted.
+        /// An absolute ID starts with /subscriptions/ and contains the entire ID of the parent resource and the ID of the sub-resource in the end.
+        /// A relative ID replaces the ID of the parent resource with a token '$self', followed by the sub-resource ID itself.
+        /// Example of a relative ID: $self/frontEndConfigurations/my-frontend.
         /// </summary>
         [Input("id")]
         public Input<string>? Id { get; set; }
--- a/sdk/dotnet/Compute/V20150615/Inputs/SubResourceArgs.cs
+++ b/sdk/dotnet/Compute/V20150615/Inputs/SubResourceArgs.cs
@@ -13,7 +13,10 @@ namespace Pulumi.AzureNative.Compute.V20150615.Inputs
     public sealed class SubResourceArgs : global::Pulumi.ResourceArgs
     {
         /// <summary>
-        /// Resource Id
+        /// Sub-resource ID. Both absolute resource ID and a relative resource ID are accepted.
+        /// An absolute ID starts with /subscriptions/ and contains the entire ID of the parent resource and the ID of the sub-resource in the end.
+        /// A relative ID replaces the ID of the parent resource with a token '$self', followed by the sub-resource ID itself.
+        /// Example of a relative ID: $self/frontEndConfigurations/my-frontend.
         /// </summary>
         [Input("id")]
         public Input<string>? Id { get; set; }
--- a/sdk/dotnet/Compute/V20160330/Inputs/SubResourceArgs.cs
+++ b/sdk/dotnet/Compute/V20160330/Inputs/SubResourceArgs.cs
@@ -13,7 +13,10 @@ namespace Pulumi.AzureNative.Compute.V20160330.Inputs
     public sealed class SubResourceArgs : global::Pulumi.ResourceArgs
     {
         /// <summary>
-        /// Resource Id
+        /// Sub-resource ID. Both absolute resource ID and a relative resource ID are accepted.
+        /// An absolute ID starts with /subscriptions/ and contains the entire ID of the parent resource and the ID of the sub-resource in the end.
+        /// A relative ID replaces the ID of the parent resource with a token '$self', followed by the sub-resource ID itself.
+        /// Example of a relative ID: $self/frontEndConfigurations/my-frontend.
         /// </summary>
         [Input("id")]
         public Input<string>? Id { get; set; }
--- a/sdk/dotnet/Compute/V20160430Preview/Inputs/SubResourceArgs.cs
+++ b/sdk/dotnet/Compute/V20160430Preview/Inputs/SubResourceArgs.cs
@@ -13,7 +13,10 @@ namespace Pulumi.AzureNative.Compute.V20160430Preview.Inputs
     public sealed class SubResourceArgs : global::Pulumi.ResourceArgs
     {
         /// <summary>
-        /// Resource Id
+        /// Sub-resource ID. Both absolute resource ID and a relative resource ID are accepted.
+        /// An absolute ID starts with /subscriptions/ and contains the entire ID of the parent resource and the ID of the sub-resource in the end.
+        /// A relative ID replaces the ID of the parent resource with a token '$self', followed by the sub-resource ID itself.
+        /// Example of a relative ID: $self/frontEndConfigurations/my-frontend.
         /// </summary>
         [Input("id")]
         public Input<string>? Id { get; set; }
--- a/sdk/dotnet/Compute/V20170330/Inputs/SubResourceArgs.cs
+++ b/sdk/dotnet/Compute/V20170330/Inputs/SubResourceArgs.cs
@@ -13,7 +13,10 @@ namespace Pulumi.AzureNative.Compute.V20170330.Inputs
     public sealed class SubResourceArgs : global::Pulumi.ResourceArgs
     {
         /// <summary>
-        /// Resource Id
+        /// Sub-resource ID. Both absolute resource ID and a relative resource ID are accepted.
+        /// An absolute ID starts with /subscriptions/ and contains the entire ID of the parent resource and the ID of the sub-resource in the end.
+        /// A relative ID replaces the ID of the parent resource with a token '$self', followed by the sub-resource ID itself.
+        /// Example of a relative ID: $self/frontEndConfigurations/my-frontend.
         /// </summary>
         [Input("id")]
         public Input<string>? Id { get; set; }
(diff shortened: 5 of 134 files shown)


joeduffy:
This looks like an incredible improvement. Should eliminate tons of confusing boilerplate.

For what it's worth, upon my first review, I expected to see the reference to "self" be explicit. Sort of like

```typescript
id: "$self/frontendIPConfigurations/publicIp",
```

It took a couple reads for me to understand the model model of "absolute" versus "relative", which I personally think is way cleaner. We'll just want to be extra sure to call this out in the docs since it's a little subtle.

mikhailshilkov:
Thinking more about `$self` or implicit, another consideration is satefy. A `SubResource` type exists in `Compute`, `HybridNetwork`, `MobileNetwork`, `Network`, `Peering`, and `ServiceFabric` resource providers. While this change is driven by `Network` resources, there is no guarantee that other resource providers have the same semantics for this type. Could it be that some of them already allow relative IDs?

Virtual Machine types suggests some shape of relative ID:

> [sourceVault](https://www.pulumi.com/registry/packages/azure-native/api-docs/compute/virtualmachine/#sourcevault_nodejs) This property is required. [SubResource](https://www.pulumi.com/registry/packages/azure-native/api-docs/compute/virtualmachine/#subresource)
>    The relative URL of the Key Vault containing the key.

Is there a chance the explicit converter breaks some existing valid formats there? It's hard to find real-world examples of this property...

I'd say we may want to err on the safe side and either a) Limit the change to Network for now b) Introduce an explicit `$self` modified.

lukehoban:
> We'll just want to be extra sure to call this out in the docs since it's a little subtle.

Another related question - will we be able to codegen these in examples generation?  It will be easier to use these by default and to discover them if they are used when legal in all our examples in the registry.

danielrbradley:
Happy with the general approach. Would much prefer the explicit indicator at the start of the path. `$self` is good or I'd initially thought `~/` might be nice and neat.

Not had time to read through the full implementation yet though.

provider/pkg/resources/convert.go
```diff
@@ -112,8 +112,27 @@ func (k *SdkShapeConverter) convertPropValue(prop *AzureAPIProperty, value inter
...
+		// specifying the full Azure resource ID explicitly.
+		// The block below takes care of resolving those relative IDs to absolute IDs.
+		typeNameParts := strings.Split(typeName, ":")
+		if len(typeNameParts) == 3 && typeNameParts[2] == "SubResource" {
+			if subId, ok := values["id"].(string); ok && !strings.HasPrefix(subId, "/") {
+				values["id"] = fmt.Sprintf("%s/%s", id, subId)
+			}
+		}
+
+		// Otherwise, delegate to the normal map convertion flow.
```
thomas11, line 132: [nit] "Otherwise" is a little surprising here since there's no branching.

thomas11 (approved):
This is great. Extra kudos for including a runnable example.

kpitzen:
I think I agree with the general feedback that something like `$self` would be almost entirely upside here. Otherwise, the rest of the approach seems very sound.

mikhailshilkov:
@pulumi/providers I implemented the `$self` syntax, this is ready for the final review.

thomas11 (approved):

mikhailshilkov:
Opened #2527 to follow up with examples in API docs

PR #2480, opened 2023-05-30
Adds methods for inserting removal deprecation warnings for v2

kpitzen:
This PR adds a function which parses the existing `v2-removed-resources.yaml` file, which describes resources being removed from V2 and their candidate resources to upgrade to. It then maps those tokens to `deprecationMessage` inside `schema-full.json` which is not committed here. This is already then plumbed through into codegen for each language SDK.


provider/pkg/gen/schema.go
```diff
@@ -208,10 +213,11 @@ func PulumiSchema(providerMap openapi.AzureProviders) (*pschema.PackageSpec, *re
...
 			gen := packageGenerator{
-				pkg:        &pkg,
-				metadata:   &metadata,
-				apiVersion: version,
-				examples:   exampleMap,
+				pkg:             &pkg,
+				metadata:        &metadata,
+				apiVersion:      version,
+				examples:        exampleMap,
+				upgradeWarnings: upgradePathMap,
```
kpitzen, line 220: Wasn't _entire_ sure where to plumb this through, but this seemed like the lowest-impact place to do so.

provider/pkg/gen/upgradePath.go
```diff
@@ -0,0 +1,42 @@
...
+		res := strings.Split(line, ": ")
+		// This is a best-effort reading of a mostly-structured file.
+		// We'll skip things like preambles etc
+		if len(res) != 2 {
+			continue
+		}
+		outputMap[res[0]] = fmt.Sprintf(
+			"%s is being removed in the next major version of this provider."+
+				"\n Please upgrade to %s to guarantee forwards compatibility.", res[0], res[1],
+		)
```
kpitzen, line 39: The output of this is still pulumi tokens at this point- should we parse them down and make them a bit more user-friendly?

provider/pkg/gen/upgradePath.go
```diff
@@ -0,0 +1,42 @@
...
+	// none of the affected resources will be in the v2 schema, so any
+	// generated deprectation warnings will be ignored.
+	outputMap := make(map[string]string)
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+	// TODO: This version is hardcoded - we should find a way to inspect at least what major
+	// version we're building and read the proper file here.
+	upgradePathFile := filepath.Join(wd, "versions", "v2-removed-resources.yaml")
```
kpitzen, line 21: I know this is a structured document, and it would be reasonable to use a yaml library here, but it seems kind of unnecessary? Not sure what the idiomatic golang way would be.

provider/pkg/gen/upgradePath.go
mikhailshilkov: The message reads fine to me. Converting token to 100% correct language-specific types would be hard, so I'd leave it as-is.

provider/pkg/gen/schema.go
```diff
@@ -593,6 +600,14 @@ func (g *packageGenerator) genResourceVariant(prov string, apiSpec *openapi.Reso
 		aliases = append(aliases, pschema.AliasSpec{Type: &alias})
 	}
 
+	if upgradeMessage, ok := g.upgradeWarnings[resourceTok]; ok {
+		if resource.deprecationMessage == "" {
+			resource.deprecationMessage = upgradeMessage
```
mikhailshilkov, line 605: Do you know what these are at this point? Should we remove those resources in 2.0?

provider/pkg/gen/upgradePath.go
```diff
@@ -0,0 +1,42 @@
...
+	for _, line := range textLines {
+		res := strings.Split(line, ": ")
+		// This is a best-effort reading of a mostly-structured file.
+		// We'll skip things like preambles etc
+		if len(res) != 2 {
+			continue
+		}
+		outputMap[res[0]] = fmt.Sprintf(
+			"%s is being removed in the next major version of this provider."+
+				"\n Please upgrade to %s to guarantee forwards compatibility.", res[0], res[1],
```
mikhailshilkov, line 38: I suspect a multi-line warning to break some of the language SDK compilation (C#?)... You may have to remove `\n` if so.

mikhailshilkov, line 38: Maybe "upgrade to %s or later"?

provider/pkg/gen/schema.go
kpitzen: The ones with no current deprecation are the ones we've decided can be removed via our schema-tools analysis. The ones with existing deprecation messages are listed in `versions/deprecated.json` - we've been warning about these resources already, and they will also be removed in 2.0. I've just messaged Daniel about this as well.

kpitzen: Of all of the resources which were previously marked as deprecated, there are only ~300 which we didn't identify via schema-tools analysis
[deprecated-resources.txt](https://github.com/pulumi/pulumi-azure-native/files/11637317/deprecated-resources.txt)

kpitzen: Also note- none of these resources are targeted by upgrades via our suggested migration paths, so it should be safe to remove them:
- We've already been notifying users of their removal for some time
- They cannot be upgraded to from existing removed resources

danielrbradley (approved):
This looks good 👍

PR #2487, opened 2023-06-02
Fix Go programgen SDK references

danielrbradley:
Fixes https://github.com/pulumi/arm2pulumi/issues/33

- Create GeneratePatchedGoProgram to wrap gogen.GenerateProgram with post-processing step.
- Detect and replace imports where the `/vX` is in the wrong place.
- Allow for any version not just v2.
- Doesn't fix issue with version specific type import.
- Rename genFn so it doesn't hide the `gen` import.
- Re-generate docs schema to update paths in examples.


provider/pkg/gen/examples.go
```diff
@@ -428,3 +429,24 @@ $ pulumi import {{ .Token }} {{ .SampleResName }} {{ .SampleResID }}
 	pkgSpec.Resources[resourceName] = res
 	return nil
 }
+
+// GeneratePatchedGoProgram generates a Go program from the given HCL2 program, but patches the
+// generated import paths to adjust for the submodules being their own modules. For example:
+// "github.com/pulumi/pulumi-azure-native-sdk/v2/resources" -> "github.com/pulumi/pulumi-azure-native-sdk/resources/v2"
+// "github.com/pulumi/pulumi-azure-native-sdk/v2/resources/v20210203" -> "github.com/pulumi/pulumi-azure-native-sdk/resources/v2/v20210203"
+func GeneratePatchedGoProgram(program *hcl2.Program) (map[string][]byte, hcl.Diagnostics, error) {
```
thomas11, line 437: Does this need to be public? Doesn't really matter except if we expect users outside of example generation, I'd say some tests would be warranted.

thomas11 (approved):

provider/pkg/gen/examples.go
danielrbradley: It's also called from the arm2pulumi module which has it's own process for calling into example generation. In the long run it would make sense for the `gen` module to expose a method that `arm2pulumi` can also use for the generation of programs and reduce the duplicated code between the modules.

PR #2488, opened 2023-06-02
Add combined docs schema to the dist

danielrbradley:
Enable breaking the dependency of the docs publishing process on checking out the schema that's being committed in the repository. There is a filesize limit of 100mb which the full schema current exceeds and therefore can't be committed which is why we currently have a second smaller "docs" schema which is comitted but is missing all explicit versions.

- All files in `dist/` are picked up in the GH release as an asset.
- The explode + impode process includes the whole of the current committed docs schema plus all explicit versions (without examples).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ _ := $(shell mkdir -p .make)
 .PHONY: default ensure dist
 default: provider arm2pulumi build_sdks
 ensure: bin/pulumictl .make/provider_mod_download
-dist: dist/pulumi-azure-native_$(VERSION_GENERIC)_checksums.txt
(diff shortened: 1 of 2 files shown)


thomas11:
PR description is missing the "why". Although I know it'd be useful for building the hosted arm2pulumi.

danielrbradley:
@thomas11 good call. Updated ✅

danielrbradley:
cc: @sean1588 this should provide context for the request I made about the docs publishing process

thomas11 (approved):
This is great. Maybe we can tie this into https://github.com/danielrbradley/pulumi-azure-native-schema such that the exploded schema is published somewhere browsable?

PR #2489, opened 2023-06-02
Fix weekly upgrades workflow

danielrbradley:
- Set PROVIDER_VERSION to a value as the makefile requires a value in CI. Use the same value as in make.
- Add call to upgrade_tools to also upgrade all other tool versions used in make.


danielrbradley:
Test run: https://github.com/pulumi/pulumi-azure-native/actions/runs/5157225949

danielrbradley:
So, the upgrade process works, but upgrading to the latest pu/pu CLI currently breaks our workflow.

danielrbradley:
Upgrading to the latest version of the CLI is blocked by
- https://github.com/pulumi/pulumi/issues/13087

rquitales (approved):

Issue #2491, opened 2023-06-02
web.ContainerApp shouldn't light up in 2.0


mikhailshilkov:
### What happened?

We generate the `web.ContainerApp` https://github.com/pulumi/pulumi-azure-native/blob/v2.0-pre/sdk/dotnet/Web/ContainerApp.cs

### Expected Behavior

It's a legacy resource that doesn't work and was excluded from 1.x, e.g. https://github.com/pulumi/pulumi-azure-native/blob/master/sdk/dotnet/Web/ContainerApp.cs does not exist

### Steps to reproduce

See the v2.0-pre branch

### Output of `pulumi about`

irrelevant

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

danielrbradley:
This was solved in https://github.com/pulumi/pulumi-azure-native/pull/2494

PR #2493, opened 2023-06-02
Drop squeezed resources from Pulumi schema so we drop generated resources, too

thomas11:
Fixes #2486 

```
2023/06/02 15:15:50 Warning: removable resource azure-native:securityandcompliance/v20210111:privateLinkServicesForM365ComplianceCenter not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:azurestackhci/v20210701preview:virtualmachineRetrieve not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:azurestackhci/v20210701preview:galleryimageRetrieve not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:securityandcompliance/v20210111:privateLinkServicesForEDMUpload not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:workloads/v20221101preview:monitor not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:databricks/v20210401preview:vNetPeering not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:azurestackhci/v20210701preview:networkinterfaceRetrieve not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:securityandcompliance/v20210111:privateLinkServicesForM365SecurityCenter not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:azurestackhci/v20210701preview:virtualnetworkRetrieve not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:azurestackhci/v20210701preview:virtualharddiskRetrieve not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:securityandcompliance/v20210111:privateLinkServicesForSCCPowershell not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:databricks/v20220401preview:vNetPeering not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:healthbot/v20201020:getbot not found in schema
2023/06/02 15:15:50 Warning: removable resource azure-native:securityandcompliance/v20210111:privateLinkServicesForO365ManagementActivityAPI not found in schema

2023/06/02 15:15:50 Removed 8433 out of 8447 resources from schema
```

PR #2494, opened 2023-06-05
Remove invokes corresponding to removed resources

danielrbradley:
1. Identify invokes to remove by tracking the path of any resources we've removed then deleting any invokes which have paths directly beneath the removed resource.
   - This removes 1768 invokes. This removes 9356 files from the comitted SDKs.
2. Exclude ContainerApp completely from Web:
   - Remove all versions of ContainerApp.
   - Resource removals are done *after* calculating the default version which is why we have to also exclude in the config.
   - Use "explicit" setting to ensure all resources are enumerated and ContainerApp is not included.
3. Update all specs to latest versions and update default versions. To do this without requiring human intervention, this removes the re-generation of v1 default versions which are no longer being updated from now on.
4. Re-remove resources after schema generation:
   - Reworked from https://github.com/pulumi/pulumi-azure-native/pull/2493
   - Some resources are added manually during generation which won't therefore be matched during the first removal, and we want to exclude these too.

Fixes https://github.com/pulumi/pulumi-azure-native/issues/2491


provider/pkg/openapi/versioner.go
```diff
@@ -101,25 +102,45 @@ func (s RemovableResources) canBeRemoved(azureProvider, resource, version string
...
 			for resourceName, resource := range resources.Resources {
 				if removable.canBeRemoved(provider, resourceName, version) {
-					removedCount++
+					removedResourceCount++
 					continue
 				}
 				filteredResources.Resources[resourceName] = resource
 			}
+			if len(resources.Resources) > 0 && len(filteredResources.Resources) == 0 {
+				// If there are no resources left, we can remove the version entirely.
```
thomas11, line 124: The PR description says
>There's also 20 more versions which only include invokes which we leave for now.

but this seems to remove these cases?

provider/pkg/openapi/versioner.go
```diff
@@ -101,25 +102,45 @@ func (s RemovableResources) canBeRemoved(azureProvider, resource, version string
...
+				removedInvokeCount += len(resources.Invokes)
+				continue
+			}
+			for invokeName, invoke := range resources.Invokes {
+				matches := trimInvokePrefix.FindStringSubmatch(invokeName)
+				if len(matches) != 2 {
+					continue
```
thomas11: When would this be the case, an invoke not starting with a lowercase letter? Are we at risk of accidentally dropping invokes here?

danielrbradley: Reworked this to use paths now

versions/v2-removed-resources.yaml
```diff
@@ -244,7 +244,7 @@ azure-native:apimanagement/v20200601preview:Gateway: azure-native:apimanagement/
 azure-native:apimanagement/v20200601preview:GatewayApiEntityTag: azure-native:apimanagement/v20220801:GatewayApiEntityTag
 azure-native:apimanagement/v20200601preview:GatewayCertificateAuthority: azure-native:apimanagement/v20220801:GatewayCertificateAuthority
 azure-native:apimanagement/v20200601preview:GatewayHostnameConfiguration: azure-native:apimanagement/v20220801:GatewayHostnameConfiguration
-azure-native:apimanagement/v20200601preview:getApiManagementServiceDomainOwnershipIdentifier
+azure-native:apimanagement/v20200601preview:getApiManagementServiceDomainOwnershipIdentifier:
```
thomas11: Without comments, we're losing track here of why they were removed. That might make future investigations harder.

versions/v2-config.yaml
```diff
@@ -537,7 +537,9 @@ VisualStudio:
 VoiceServices:
 Web:
   preview: exclude
-  expectAdditions: true
+  explicit: true
```
thomas11: This is still weird to me. Do `exclusions` only work with `explicit`?

provider/pkg/versioning/gen.go
```diff
@@ -100,7 +83,8 @@ func (v VersionMetadata) WriteTo(outputDir string) error {
 type VersionSources struct {
 	activePathVersions        providerlist.ProviderPathVersions
 	requiredExplicitResources []string
-	v1Spec                    Spec
+	V1Lock                    openapi.DefaultVersionLock
```
thomas11, line 86: does this need to be public/uppercase, different from the other members?

provider/pkg/versioning/build_schema.go
```diff
@@ -82,6 +82,8 @@ func BuildSchema(args BuildSchemaArgs) (*BuildSchemaResult, error) {
 		return nil, err
 	}
 
+	dropFromSchema(pkgSpec, versionSources.V2ResourcesToRemove)
```
thomas11, line 85: I think this warrants a comment for future readers, here or above, about why we're removing resources twice.

thomas11 (approved):
LGTM overall! Just a few minor things.

versions/v2-config.yaml
danielrbradley: Yeah kinda. We can't currently remove items from the tracking version. Explicit forces the tracking version to be replaced with additions - which then works with exclusions. The ideal fix is to actually add support for "exclusions" to the "spec" file so we just copy the exclutions through to there from the config.

provider/pkg/versioning/gen.go
danielrbradley: Ah that's because I moved it from the VersionMetadata - in which everything is public. I'm not sure why everything's private - there's no need for them to be hidden.

thomas11: Hidden makes sense as default IMO to keep dependencies from spreading all over.

versions/v2-removed-resources.yaml
danielrbradley: I fear that ship has already sailed for this file. Essentially anything in here is just marked as no longer required in the future - we don't know how it was determined. I think for the next major version we'll have to think a little more about this removal process.

Issue #2477, opened 2023-05-25
Missing insights/diagnosticsetting in v2 alpha


mnlumi:
### What happened?

When trying out v2 alpha, a user reported that [insights/diagnosticsetting](https://www.pulumi.com/registry/packages/azure-native/api-docs/insights/diagnosticsetting/) is missing. I can also confirm that the resource is not shown in the temporary schema export [here](https://github.com/danielrbradley/pulumi-azure-native-schema/tree/main/insights/resources.)

`The type or namespace name 'DiagnosticSetting' does not exist in the namespace 'Pulumi.AzureNative.Insights'`

### Expected Behavior

I expect the resource to be available in v2.

### Steps to reproduce

Using v2 alpha (2.0.0-alpha.1), attempt to create a 'insights/diagnosticsetting' resource.

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

thomas11:
This seems to be a bug in [filterCandidateVersions](https://github.com/pulumi/pulumi-azure-native/blob/ab5615a2d2d191958e951acb175ef4da270bca17/provider/pkg/versioning/defaultVersion.go#L285). The resource `DiagnosticSetting` is only found in preview API versions, 2017-05-01-preview and 2021-05-01-preview. In `v2-config.yaml`, Insights is configured as `explicit: true` but not `preview: prefer`. That causes resources in stable API versions to be included and preview versions to be excluded (L285 linked above).

We might need to either change this logic, or add another option `preview: allow` or similar that includes resources from preview versions when set.

@danielrbradley, what are your thoughts?

danielrbradley:
I think your assessment is correct @thomas11 - this is due to in being only in preview versions, but there's also more recent stable versions - but these don't include this specific resource. 

This is a rather odd pattern of releases for the Insights service. 

In `2021-05-01-preview` there were 4 resources released:
- AutoscaleSetting
- DiagnosticSetting
- ManagementGroupDiagnosticSetting
- SubscriptionDiagnosticSetting

`AutoscaleSetting` was then released in the stable version `2022-10-01` but the other resouces were not. The other three resources have all had previous preview releases but no stable releases. 

There are frequent stable releases following previews for this service so I don't think our heuristic to preview stable versions only is wrong here. 

[...]

PR #2503, opened 2023-06-08
Include old preview versions

danielrbradley:
- Include latest preview version if no new stable versions after 1 year (adds 27 resources).
- Consider all version when calculating additions. Only ignore private preview versions. This takes us to up 242 additional resources in the default version.
- Fix sorting of versions so that previews are considered older than stable versions. 
- Upgrade CI to Go 1.20 so we can use `time.Compare`.

Fixes https://github.com/pulumi/pulumi-azure-native/issues/2359
Fixes https://github.com/pulumi/pulumi-azure-native/issues/2477


thomas11 (approved):
LGTM although I'm not 100% sure I totally understand the third commit.

provider/pkg/versioning/defaultVersion.go
```diff
@@ -294,12 +303,33 @@ func filterCandidateVersions(versions VersionResources, previewPolicy string) []
 		// If no more to iterate through, just add
 		if i == 0 || !containsRecentStable(orderedVersions[:i], version) {
 			candidateVersions.Add(version)
+			continue
+		}
+
+		isOld, err := isMoreThanOneYearOld(version)
```
mikhailshilkov, line 309: Could you please explain the genesis of this 1-year rule?

provider/pkg/versioning/defaultVersion.go
```diff
@@ -181,8 +181,16 @@ func buildSpec(providerName string, versions VersionResources, curations Curatio
 	if existing.Additions != nil {
 		existingAdditions = *existing.Additions
 	}
+	sortedVersions := []string{}
+	for version := range versions {
+		sortedVersions = append(sortedVersions, version)
+	}
+	sort.Strings(sortedVersions)
```
mikhailshilkov, line 188: Note that alphabetical sorting will put stable versions before preview ones (v20230101 < v20230101preview).

mikhailshilkov (commented):
Is schema-tools compare not wired correctly? It shows a bunch of breaking changes and also only 31 new resources...

provider/pkg/versioning/defaultVersion.go
danielrbradley: It's a somewhat arbitary amount of time, but there's a pattern of a preview version being replaced with a stable version after a few months - at which point we'd prefer to be pulling from stable versions if possible.

provider/pkg/versioning/defaultVersion.go
danielrbradley: Yes, good catch. I'll fix this up

PR #2512, opened 2023-06-13
Silence conflicting paths warning for ManagementGroupNetworkManagerConnection

thomas11:
Comparing network/resource-manager/Microsoft.Network/preview/2021-05-01-preview/networkManagerConnection.json with network/resource-manager/Microsoft.Network/stable/2022-11-01/networkManagerConnection.json, as well as their respective examples, indicates that it's the same resource.

We exclude the old path to silence the path conflict warning.


versions/v2-config.yaml
```diff
@@ -414,6 +414,8 @@ Network:
   explicit: true
   notes: Large surface area that's likely to conflict as versions evolve
   exclusions:
+    # The path of this resource changed since v1 but it's otherwise the same resource.
+    ManagementGroupNetworkManagerConnection: "2022-09-01"
```
mikhailshilkov, line 418: Does it mean we exclude the new stable API?

danielrbradley: If excluded, I'd also expect to see a change to the spec and lock file also.

thomas11: No, this excluded version is the one _before_ the latest stable that we include in v2. The only thing this exclusion does is silence the conflicting paths warning.

danielrbradley (approved):

PR #2515, opened 2023-06-13
Re-add old SQL resources

danielrbradley:
These are still listed in the documentation so should be made available in the top-level modules


danielrbradley:
Looks like the docs schema wasn't re-generated on the last PR so slightly bigger diff than expected.

thomas11 (approved):

PR #2521, opened 2023-06-13
Disambiguate conflicting paths between Service Fabric clusters

thomas11:


danielrbradley:
Any idea why it causes all the version modules to be added?

Can we exclude any based on previous deprecations or squashing?

danielrbradley (approved):
I'm happy with where this is at. I rebased again and also added:

- Removal of the private preview version.
- Ran the squeeze process to cut down the number of new versions getting added.

If you're happy too, go ahead and merge!

PR #2524, opened 2023-06-14
Handle string sets during preview

danielrbradley:
Closes https://github.com/pulumi/pulumi-azure-native/issues/2525

At preview time, we generally just return the inputs back as outputs - as a best-guess as to what might be available after deployment. However, for string sets the input and output SDK shapes are different. Therefore we need to handle this scenario specifically in the "input to output converter". For now, we return a top-level computed object - which is unknown at preview. 

It might be possible to make this known if all elements in the input array are also known but this would be significantly more complex with likely no real-world benefit as the properties within the values still won't be known and they could just extract a variable with their input ids as an array which would be easier to loop over if required rather than extracting the keys from the map.

- Fix preview conversion of StringSet input just gets an unknown output at preview-time rather than returning the input list.
- Add special test case for this type.
- Fix metadata so we have it as the object type in metadata, but an array in the schema.

Metadata diff (when converted to yaml using `cat bin/metadata-compact.json | yj -jy > bin/metadata-compact.yaml`):

```diff
2623,2625c2623
<         type: array
<         items:
<           type: string
---
>         type: object
2657,2659c2655
<         type: array
<         items:
<           type: string
---
>         type: object
7959,7961c7955
<         type: array
<         items:
<           type: string
---
>         type: object
9195,9197c9189
<         type: array
<         items:
<           type: string
---
>         type: object
10645,10647c10637
<         type: array
<         items:
<           type: string
---
>         type: object
12272,12274c12262
<         type: array
<         items:
<           type: string
---
>         type: object
14000,14002c13988
<         type: array
<         items:
<           type: string
---
... snipped ...
```

### DIFF
--- a/provider/pkg/resources/convert.go
+++ b/provider/pkg/resources/convert.go
@@ -402,6 +402,7 @@ func (k *SdkShapeConverter) previewOutputValue(inputValue resource.PropertyValue
 		// 		return resource.NewObjectProperty(v)
 		// 	}
 		// }
+
 		// Fallback to legacy behaviour - assuming the input is same as output
 		return inputValue
 
@@ -419,30 +420,11 @@ func (k *SdkShapeConverter) previewOutputValue(inputValue resource.PropertyValue
 }
 
 func (k *SdkShapeConverter) makeComputedValue(prop *AzureAPIProperty) resource.PropertyValue {
-	if prop == nil {
-		return resource.NewStringProperty("")
-	}
-	if prop.Const != nil {
+	if prop != nil && prop.Const != nil {
 		return resource.NewStringProperty(prop.Const.(string))
 	}
-	switch {
-	case prop.Type == "boolean":
-		return resource.NewBoolProperty(false)
-	case prop.Type == "integer" || prop.Type == "number":
-		return resource.NewNumberProperty(0)
-	case prop.Type == "array" || prop.Items != nil:
-		if prop != nil && prop.Items != nil {
-			return resource.NewArrayProperty([]resource.PropertyValue{})
-		}
-	case prop.Type == "object" || prop.AdditionalProperties != nil:
-		return resource.NewObjectProperty(resource.PropertyMap{})
-	case prop.Type == "object" || strings.HasPrefix(prop.Ref, "#/types/"):
-		// TODO: Expand internal type
-		return resource.NewObjectProperty(resource.NewPropertyMap(nil))
-	case prop.Ref != "":
-		return resource.NewObjectProperty(resource.NewPropertyMap(nil))
-	}
-	// Default to string
+	// To mark something as computed, we always use a string property with an
+	// empty string, regardless of the type.
 	return resource.NewStringProperty("")
 }
 
--- a/provider/pkg/resources/convert_test.go
+++ b/provider/pkg/resources/convert_test.go
@@ -621,7 +621,7 @@ func TestPreviewOutputs(t *testing.T) {
 				metadata: metadata,
 			})
 			expected := resource.PropertyMap{
-				"threshold": resource.MakeComputed(resource.NewNumberProperty(0)),
+				"threshold": resource.MakeComputed(resource.NewStringProperty("")),
 			}
 			assert.Equal(t, expected, actual)
 		})
@@ -633,7 +633,7 @@ func TestPreviewOutputs(t *testing.T) {
 				metadata: metadata,
 			})
 			expected := resource.PropertyMap{
-				"threshold": resource.MakeComputed(resource.NewNumberProperty(0)),
+				"threshold": resource.MakeComputed(resource.NewStringProperty("")),
 			}
 			assert.Equal(t, expected, actual)
 		})
@@ -662,7 +662,7 @@ func TestPreviewOutputs(t *testing.T) {
 				metadata: metadata,
 			})
 			expected := resource.PropertyMap{
-				"enabled": resource.MakeComputed(resource.NewBoolProperty(false)),
+				"enabled": resource.MakeComputed(resource.NewStringProperty("")),
 			}
 			assert.Equal(t, expected, actual)
 		})
@@ -674,7 +674,7 @@ func TestPreviewOutputs(t *testing.T) {
 				metadata: metadata,
 			})
 			expected := resource.PropertyMap{
-				"enabled": resource.MakeComputed(resource.NewBoolProperty(false)),
+				"enabled": resource.MakeComputed(resource.NewStringProperty("")),
 			}
 			assert.Equal(t, expected, actual)
 		})
@@ -784,7 +784,7 @@ func TestPreviewOutputs(t *testing.T) {
 				metadata: metadata,
 			})
 			expected := resource.PropertyMap{
-				"array": resource.MakeComputed(resource.NewArrayProperty([]resource.PropertyValue{})),
+				"array": resource.MakeComputed(resource.NewStringProperty("")),
 			}
 			assert.Equal(t, expected, actual)
 		})
@@ -796,7 +796,7 @@ func TestPreviewOutputs(t *testing.T) {
 				metadata: metadata,
 			})
 			expected := resource.PropertyMap{
-				"array": resource.MakeComputed(resource.NewArrayProperty([]resource.PropertyValue{})),
+				"array": resource.MakeComputed(resource.NewStringProperty("")),
 			}
 			assert.Equal(t, expected, actual)
 		})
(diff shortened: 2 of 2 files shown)


examples/py-user-assigned-identity/requirements.txt
```diff
@@ -0,0 +1 @@
+pulumi>=2.0.0,<3.0.0
```
kpitzen, line 1: This seems not correct, right?

danielrbradley: Hmm, just copied from the other test in there .. guessing it probably just gets ignored/overwritten somewhere in program test.

mikhailshilkov: Let's change it everywhere

examples/py-user-assigned-identity/__main__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2021, Pulumi Corporation.  All rights reserved.
```
mikhailshilkov, line 1: 2023

examples/py-user-assigned-identity/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: py-user-assigned-identity
+runtime: python
+description: A minimal Azure Python Pulumi program
```
mikhailshilkov, line 3: Please add a description

provider/pkg/resources/convert.go
```diff
@@ -325,33 +325,103 @@ func (k *SdkShapeConverter) PreviewOutputs(inputs resource.PropertyMap,
...
 		} else {
-			result[key] = resource.MakeComputed(resource.NewStringProperty(""))
+			result[key] = k.previewOutputValue(&inputValue, nil)
 		}
 	}
 	return result
 }
 
-func (k *SdkShapeConverter) previewOutputValue(inputValue resource.PropertyValue,
+func (k *SdkShapeConverter) previewOutputValue(inputValue *resource.PropertyValue,
```
mikhailshilkov, line 336: Why this change in `inputValue` type?

provider/pkg/resources/convert_test.go
```diff
@@ -530,10 +530,8 @@ func TestPreviewOutputs(t *testing.T) {
 		"more":  resource.MakeComputed(resource.NewStringProperty("")),
 		"tags":  resource.MakeComputed(resource.NewStringProperty("")),
 		"union": resource.MakeComputed(resource.NewStringProperty("")),
-		"untypedArray": resource.NewArrayProperty([]resource.PropertyValue{
-			resource.NewObjectProperty(resource.PropertyMap{"key1": resource.NewStringProperty("value1")}),
-			resource.NewObjectProperty(resource.PropertyMap{"key1": resource.NewStringProperty("value2")}),
-		}),
+		// If we have no type information for an array or dictionary, we can't do anything with it.
+		"untypedArray": resource.MakeComputed(resource.NewStringProperty("")),
```
mikhailshilkov, line 534: Why do we lose properties on arrays but not dicts?

provider/pkg/resources/convert.go
```diff
@@ -325,33 +325,103 @@ func (k *SdkShapeConverter) PreviewOutputs(inputs resource.PropertyMap,
...
 	prop *AzureAPIProperty) resource.PropertyValue {
+	if prop == nil {
```
mikhailshilkov, line 338: It's surprising to see so many changes here without any changes in tests. Why is that? Is this somehow just pure refactoring with no functional changes? Or are we missing tests?

mikhailshilkov (commented):
Took the first pass and asked questions. A few high-level remarks:
1. Reviews would be easy if commits were even smaller, with refactorings split into separate ones.
2. The metadata diff in text isn't quite heplful... Can't we commit the file now? Is it still above 100MB?

provider/pkg/resources/convert.go
danielrbradley: There are tests - we're just changing very little of the behaviour. There's one change I made to the tests - untyped arrays will just come back as unknowns during preview as we've got no type information to verify that the output will match the input shape. 

The first iteration of the code had a good few failing tests but I maintained back compatibility where possible.

provider/pkg/resources/convert_test.go
```diff
@@ -542,21 +542,50 @@ func TestPreviewOutputs(t *testing.T) {
...
+		},
+	}
+	inputMap := resource.NewPropertyMapFromMap(inputs)
+	metadata := map[string]AzureAPIProperty{
+		"name": {
+			Type: "string",
+		},
+		"userAssignedIdentities": {
+			Type: "object",
+			AdditionalProperties: &AzureAPIProperty{
```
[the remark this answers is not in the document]
danielrbradley: Not for the output type, no.

provider/pkg/gen/properties.go
```diff
@@ -246,47 +246,44 @@ func (m *moduleGenerator) genProperty(name string, schema *spec.Schema, context
...
 		IsStringSet:          isStringSet,
 	}
 
 	// Input types only get extra information attached
 	if !isOutput {
 		if m.isEnum(&propertySpec.TypeSpec) {
-			apiProperty = resources.AzureAPIProperty{Type: "string"}
+			apiProperty = resources.AzureAPIProperty{Type: "string", ForceNew: forceNew}
 		} else {
-			// Set additional properties when it's an input
```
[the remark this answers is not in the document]
danielrbradley: Ah, before squashing there was a step where I took the assignments out of this conditional block - must have got lost when putting them back.

provider/pkg/resources/convert.go
mikhailshilkov: Thank you, that's great. Can you elaborate on why you needed to make these changes then?

provider/pkg/resources/convert_test.go
mikhailshilkov: I'm a bit scared that something relied on the past behavior, given there was a test for it, and that we may break it inadvertently. Should we dig up a PR which introduced the original test and see if anything may be affected?

provider/pkg/resources/convert.go
```diff
@@ -325,33 +325,103 @@ func (k *SdkShapeConverter) PreviewOutputs(inputs resource.PropertyMap,
...
+		return resource.MakeComputed(resource.NewStringProperty(""))
+	}
+	if prop.Const != nil {
+		return resource.NewStringProperty(prop.Const.(string))
+	}
+	if inputValue == nil {
```
mikhailshilkov, line 344: What does this case represent? It looks like this will never be true in the current implementation? (you never pass a nil)

provider/pkg/resources/convert.go
danielrbradley: The changes are required for the fall-through case - we were defaulting to returning the input if none of the conditions matched. Now we're being more stringent on our type checking and then falling back to returning a computed property if we can't guarentee the input valyue matches the output shape.

provider/pkg/resources/convert.go
danielrbradley: Yeah, pretty sure we don't ever get to a nil here, but the type is a pointer, so seems safest to include the check to avoid any chance of runtime panics - especially if this is gets called differently in the future.

provider/pkg/resources/convert.go
danielrbradley: Now refactored to put it back to a non-pointer in a follow-up commit.

provider/pkg/gen/properties.go
```diff
@@ -246,47 +246,44 @@ func (m *moduleGenerator) genProperty(name string, schema *spec.Schema, context
...
+		WillReplaceOnChanges: forceNew,
+	}
+
 	if isStringSet {
-		typeSpec.Type = "array"
-		typeSpec.AdditionalProperties = nil
-		typeSpec.Items = &pschema.TypeSpec{
+		propertySpec.Type = "array"
+		propertySpec.AdditionalProperties = nil
+		propertySpec.Items = &pschema.TypeSpec{
```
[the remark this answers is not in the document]
danielrbradley: I've now renamed these variables which makes the whole section much easier to understand.

danielrbradley:
@mikhailshilkov I think I've now addressed all your feedback inline.

1. With this work this wasn't much of a refactoring job - more of a straight up rewrite so I'm afraid there wasn't really any intermediate steps when reworking the core logic.
2. We _could_ start committing the metadata again, though I think it would be of limited value as-is. We'd probably want to commit it without compaction - which would push the size up, and the diffs would likely often be hidden due to being in a single file. Perhaps we could explore this as a separate topic at some other point - perhaps splitting into smaller files (per service maybe) or even using a different format to make the diffs better.

danielrbradley:
Just for completeness, here's what we see when we run the new tests against the old preview implementation.

- All of the valid cases (except string set) function the same.
- All of the mismatch cases now function differently (these weren't really tested before).
- We're now a little smater with const types - and will just return the const value at preview time.

image

kpitzen (approved):

PR #2533, opened 2023-06-21
Refactor version removal & generation

### LINKED ISSUE #2502: Some invokes are still not removed from 2.0
### What happened?

E.g. [`GetACIService`](https://github.com/pulumi/pulumi-azure-native/blob/v2.0-pre/sdk/dotnet/MachineLearningServices/V20200515Preview/GetACIService.cs) is still present in `V20200515Preview`, while its counterpart resource is removed. Also, it's marked as obsolete in 1.0: https://github.com/pulumi/pulumi-azure-native/blob/master/sdk/dotnet/MachineLearningServices/V20200515Preview/GetACIService.cs#L12

### Expected Behavior

Invoke above does not exist in 2.0.

### Steps to reproduce

-

### Output of `pulumi about`

-

### Additional context

_No response_

### Contributing

danielrbradley:
Closes #2502 

- Improve version switching code for reading in version files. Use a single number to control the version instead of a bool + environment variable. Parse this automatically from the main build version we already pass in.
- When we calculate invokes to remove based on resource paths being removed, output these to a file in the versions folder so we can see what's happening.
- Remove V2 prefix from variables.
- Move resource & invoke exclusion logic to function behind interface which is called from within the main gen.PulumiSchema function rather than trying to modify the set of input specs or already-generated schema.
- Fix aliases so they're not removed if the resource is removed.


provider/pkg/gen/schema.go
```diff
@@ -639,57 +648,58 @@ func (g *packageGenerator) genResourceVariant(prov string, apiSpec *openapi.Reso
...
 	functionTok := fmt.Sprintf(`%s:%s:get%s`, g.pkg.Name, module, resource.typeName)
-
-	var readOp *spec.Operation
-	switch {
-	case resource.PathItemList != nil:
-		if resource.PathItemList.Post != nil {
-			readOp = resource.PathItemList.Post
-		} else {
-			readOp = resource.PathItemList.Get
+	if g.constraints.ShouldInclude(prov, g.apiVersion, resource.typeName, functionTok) {
```
danielrbradley, line 643: This whole block is just indent changes having added the new `if ...` block.

provider/pkg/gen/schema.go
```diff
@@ -40,8 +40,12 @@ import (
 // Note - this needs to be kept in sync with the layout in the SDK package
 const goBasePath = "github.com/pulumi/pulumi-azure-native-sdk/v2"
 
+type Constrains interface {
```
thomas11, line 43: typo Constrain*t*s

provider/pkg/versioning/gen.go
```diff
@@ -194,3 +200,40 @@ func ReadRequiredExplicitResources(path string) ([]string, error) {
 	lines := strings.Split(string(bytes), "\r")
 	return lines, nil
 }
+
+func findRemovedInvokesFromResources(providers openapi.AzureProviders, removedResources openapi.RemovableResources) openapi.RemovableResources {
```
thomas11, line 243: This seems somewhat complicated. Now that we have `Constraints`, can't we make it handle Invokes, too? I assume the path would need to be added as a parameter.

danielrbradley: Hmm, yeah, might be able to drop this entirely now given that way that we'll skip adding the invoke for the resource if the resource is excluded.

thomas11:
The interaction between the new `Constraints` and schema generation has no tests currently - would be great to add some.

provider/pkg/versioning/gen.go
danielrbradley: I don't think we can easily remove this. There's still additional invokes (the ones not prefixed `get...`) which are added to the schema independently to the resource. Moving this in to be part of the constraints would require it to be stateful in very bad ways.

PR #2534, opened 2023-06-22
Remove removable invokes for generated resources

### LINKED ISSUE #2502: Some invokes are still not removed from 2.0

thomas11:
- Downgrade default value warnings to higher log level
- Remove generated invokes, like from discriminated union types, directly from schema

This is a follow-up to #2494.
Resolves #2502.


danielrbradley (commented):
Very nice! 👏 

I'm wondering how we go abou deprecating these resources in v1 given we can only locate these resources after generation?

danielrbradley (approved):

thomas11:
Replaced by #2533 and #2535.

PR #2535, opened 2023-06-23
Downgrade default value warnings to higher log level

thomas11:


danielrbradley (approved):
We should do this in more places!

PR #2536, opened 2023-06-23
Re-calculate removable resources based on schema-tools squeeze

danielrbradley:
- Add method to generate special version of the schema especially for the squeeze.
- Convert removable-resources to JSON to be compatible with the schema-tools output.

Stacked on top of: #2533


provider/pkg/versioning/build_schema.go
```diff
@@ -56,8 +57,13 @@ func BuildSchema(args BuildSchemaArgs) (*BuildSchemaResult, error) {
 	if err != nil {
 		return nil, err
 	}
+	if args.OnlyExplicitVersions {
+		versionMetadata = VersionMetadata{}
```
thomas11, line 61: This overwrites the just loaded data - could save the loading time and only do it in the other branch.

thomas11 (approved):
LGTM so far!

danielrbradley:
Breaking changes look fine
- IpCommunityList renamed to IpCommunity
- IpPrefixList renamed to IpPrefix

PR #2556, opened 2023-07-07
Fix exclusions ordering

danielrbradley:
This causes exclusions to be missed because the preview was considered to be newer than the excluded stable version.


mikhailshilkov (approved):
Makes a ton of sense, LGTM

thomas11 (approved):

Issue #594, opened 2021-03-02
Add the Key Vault 'accessPolicies' resource to enable incremental updated


indigogiant:
It would be nice to be able to incrementally apply access policies to Key Vault using the `accessPolicies` resource. See https://github.com/Azure/azure-quickstart-templates/blob/master/101-keyvault-add-access-policy/azuredeploy.json and https://www.codeisahighway.com/how-to-incrementally-update-keyvault-access-policies-with-arm-templates for examples.

We have a central key vault with certain secrets in it, and as we deploy components it would be great to be able to incrementally add access policies using the nextgen provider.

andrewdmoreno:
@mikhailshilkov Any updates to share as to whether this feature to incrementally add access policies is on the roadmap?

mikhailshilkov:
@andrewdmoreno It's not included in the current roadmap for 1.0

andrewdmoreno:
Thanks for the quick response. In the meantime is there any workaround or approach you can think of to handle the case where you do not want to interfere with access policies that may have been added outside the stack? 

Would some variation of calling `GetVault` and iterating over the `AccessPolicyEntryResponseArgs`  returned in `Properties.AccessPolicies` potentially be a strategy that could work?

mikhailshilkov:
@andrewdmoreno You should likely use the classic Azure provider for this particular resource here. You can keep using Azure-Native for everything else.

krewi1:
Just to help somebody who will find this issue and have similiar use case as we have, to avoid "dead end". Our use case is: Import already existing keyvault to pulumi and then manage its access policies from other pulumi stacks. We have used azure native for managing that imported keyvault and set ignoreChanges on accessPolicies property, bcs they are managed in other stacks dynamicaly. IT WAS BAD DECISION. It seems like keyvault resource remember all the state which is connected to it. So when we keyvault is created, pulumi stores that access policies (f.e.: []). Then when we create access policies elsewhere everything seems fine until we have to do some change on that keyvault (lets say change some property), then pulumi apply its stored value for accessPolicy, which is [] and deletes dynamicaly created ones.

Prakash2024:
>  You can set configs with a separate resource https://www.pulumi.com/docs/reference/pkg/azure-nextgen/web/webapplicationsettings/

@mikhailshilkov  May I know what should I use in case of azure native for creating app settings as a independent resource? I am trying to
- Create a function app using azure native
[...]

PR #2588, opened 2023-07-18
Key vault access policy

danielrbradley:
Design Spike for #594 

The endpoint is very odd and does not naturaly map to a resource. 
- The URL is `/.../Microsoft.KeyVault/vaults/{vaultName}/accessPolicies/{add|replace|delete}`. 
- The identifier for the actual policy is in the body of the request: tenantId, objectId and applicationId
- The operation (add/replace/delete) is a part of the URL rather than the method.
- This appears to be a non-standard workaround for not supporting JSON patch for accessPolicies on the main keyvault resource.
- The first spike I did revolved around starting from the spec but it produced a very odd resource shape with "operation" as a field (and identifier).

Therefore my recommendation would be to build a custom resource implemented against the SDK rather than the raw specifications.

Thoughts:
- This doesn't include a GetAccessPolicy and GetAccessPolicyOutputs method. Looks like we'd have to implement this separately as an invoke, if we want it.
- We _could_ derive some of the fields & docs from the specification but it's not clear exactly how to make this work given we'd need some special casing when parsing the URL and combining it with the body parameters.


provider/pkg/resources/custom_keyvault_accesspolicy.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/keyvault/mgmt/keyvault"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+func keyVaultAccessPolicy(client keyvault.VaultsClient) *CustomResource {
+	return &CustomResource{
+		tok:  "azure-native:keyvault:AccessPolicy",
+		path: "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}/objectId/{objectId}/applicationId/{applicationId}",
+		Schema: &schema.ResourceSpec{
```
mikhailshilkov, line 19: ARM specs define the access policy types, right? Should we use that one directly (without hard coding it manually) to make sure we stay up to date over time?

provider/pkg/resources/custom_keyvault_accesspolicy.go
```diff
@@ -0,0 +1,168 @@
...
+	"regexp"
+
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/keyvault/mgmt/keyvault"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+func keyVaultAccessPolicy(client keyvault.VaultsClient) *CustomResource {
+	return &CustomResource{
+		tok:  "azure-native:keyvault:AccessPolicy",
```
mikhailshilkov, line 17: Is the plan to ship the default api version only?

mikhailshilkov (commented):
This makes sense if we want to go on the path of an entirely manual implementation. We assumption was that we may want to code around the endpoint weirdness and hide operation from SDK shapes, but otherwise keep automatic ARM integration, including support for all api versions.

provider/pkg/resources/custom_keyvault_accesspolicy.go
danielrbradley: Yup, I think that's what we've done for other manual implementations (e.g. storage objects).

provider/pkg/resources/custom_keyvault_accesspolicy.go
danielrbradley: Not really - the types are wrong because they describe an operation rather than a resource.

mikhailshilkov: What is wrong with https://github.com/Azure/azure-rest-api-specs/blob/14d24d17491d8c2bde24532cb8cc2d663c0ffd9f/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2023-02-01/keyvault.json#L1511 or https://github.com/Azure/azure-rest-api-specs/blob/14d24d17491d8c2bde24532cb8cc2d663c0ffd9f/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2023-02-01/keyvault.json#L1405?

`operationKind` is just a single path parameter, that would be easy to ignore as a special case.

PR #2631, opened 2023-08-03
Keep previous default resources

danielrbradley:
The intention when moving to v2 is that all the default resources in v1 would be available as explicit versions.

Somewhere along the way most of these go squeezed out. This PR ensures that any explicit version from the previous versions's lock is included as an explicit version's resource.

This adds approximately 1776 resources to each SDK (as well as corresponding get methods).

Fixes https://github.com/pulumi/pulumi-azure-native/issues/2634

### DIFF
--- a/provider/pkg/arm2pulumi/internal/test/templates_test.go
+++ b/provider/pkg/arm2pulumi/internal/test/templates_test.go
@@ -4,7 +4,7 @@ package test
 
 import (
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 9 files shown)


provider/pkg/versioning/gen.go
```diff
@@ -37,6 +37,13 @@ func (v VersionMetadata) ShouldInclude(provider string, version string, typeName
 			}
 		}
 	}
+	if resources, ok := v.PreviousLock[provider]; ok {
```
thomas11, line 40: This repeated triple-nested lookup could be a nice helper on DefaultVersionLock

thomas11, line 40: I think this warrants a comment like "Include all resources in the previous major version's lock so users are not forced to upgrade"

thomas11 (approved):

mikhailshilkov:
Sorry I didn't review it on time. 

This looks about right to me. Questions:

- Can we add tests to ShouldInclude to cover this change?
- Can we amend one of ProgramTest tests to cover a version of this change?
- What is the impact on SDK size?

mikhailshilkov:
One more question: 

> schema-tools stats -p azure-native

says we have 

> "TotalResources": 1658

How is it possible this PR adds 1776 resources?

danielrbradley:
Schema tools stats doesn't include explicit versions as it's only the docs schema.

It looks like it was actually closer to 1,200 resources based on the the SDK diff changes.

PR #2635, opened 2023-08-04
Fix timeout handling in example program gen

thomas11:
The previous approach using `defer` did not work in my local testing.
I believe that is because the panic occurs in another goroutine than `recover`.

This approach is a standard technique, described e.g. in https://gobyexample.com/timeouts.

### DIFF
--- a/provider/pkg/gen/examples.go
+++ b/provider/pkg/gen/examples.go
@@ -313,6 +313,8 @@ func generateExamplePrograms(example resources.AzureAPIExample, body *model.Body
 }
 
 func recoverableProgramGen(name string, program *hcl2.Program, fn programGenFn) (files map[string][]byte, err error) {
+	const timeout = 3 * time.Minute
+
 	defer func() {
 		if r := recover(); r != nil {
 			err = fmt.Errorf("panic recovered during generation: %v", r)
@@ -344,8 +346,10 @@ func recoverableProgramGen(name string, program *hcl2.Program, fn programGenFn) 
 				log.Printf("failed to write diagnostics: %v", err)
 			}
 		}
-	case <-time.After(3 * time.Minute):
-		log.Printf("%s timed out after 3m", name)
+	case <-time.After(timeout):
+		msg := fmt.Sprintf("%s timed out after %v", name, timeout)
+		log.Println(msg) // caller doesn't print error due to verbosity
+		err = fmt.Errorf(msg)
 		break
 	}
 


mikhailshilkov:
@thomas11 Could you please open (or find?) a corresponding issue with problem description and link it to the PR? Thank you!

thomas11:
@mikhailshilkov done.
Fixes #2650

provider/pkg/gen/examples.go
```diff
@@ -312,44 +312,43 @@ func generateExamplePrograms(example resources.AzureAPIExample, body *model.Body
...
+		files map[string][]byte
+		d     hcl.Diagnostics
+		err   error
 	}
-	if d.HasErrors() {
-		err := program.NewDiagnosticWriter(os.Stderr, 0, true).WriteDiagnostics(d)
-		if err != nil {
-			log.Printf("failed to write diagnostics: %v", err)
+
+	c := make(chan programGenResult, 1)
```
rquitales, line 328: I might have opted to just use an error channel here to avoid needing to create a struct object since we're already using named return vars, but that might make the code harder to read/follow.

rquitales (approved):
LGTM. The previous implementation would have failed since the panic occurs in a separate goroutine.

provider/pkg/gen/examples.go
thomas11: Ah, an inverse approach where the happy path happens inside the Goroutine and the error handling outside, if an error is sent through the channel?

rquitales: Something like this is what I was thinking of (don't even need an error channel actually):
```
func recoverableProgramGen(name string, program *hcl2.Program, fn programGenFn) (files map[string][]byte, err error) {
...
done := make(chan struct{}, 1)
go func() {
		var d hcl.Diagnostics
                 // files will be returned due to named return variable
		files, d, err = fn(program)
                 if err != nil {
                        // err is assignined to the named return error variable, so no need to explicitly handle
                        done <- struct{}{}
                 }
		if d.HasErrors() {
			err = program.NewDiagnosticWriter(os.Stderr, 0, true).WriteDiagnostics(res.d)
			if err != nil {
				log.Printf("failed to write diagnostics: %v", err)
			}
		}
                 done <- struct{}{}
	}()
select {
case <-done:
    break
case <- time.After(3 * time.Minute):
    log.Printf("%s timed out after 3m", name)
    break
}
```

provider/pkg/gen/examples.go
```diff
@@ -312,44 +312,43 @@ func generateExamplePrograms(example resources.AzureAPIExample, body *model.Body
...
 		}
+		files = res.files
+		if res.d.HasErrors() {
+			err = program.NewDiagnosticWriter(os.Stderr, 0, true).WriteDiagnostics(res.d)
+			if err != nil {
+				log.Printf("failed to write diagnostics: %v", err)
+			}
+		}
+	case <-time.After(3 * time.Minute):
+		log.Printf("%s timed out after 3m", name)
```
rquitales, line 348: On a second pass, do we actually want to return an error if it timed out?

rquitales (changes requested):

provider/pkg/gen/examples.go
thomas11: you're right, good catch

rquitales (approved):

Issue #2467, opened 2023-05-24
Specify custom API version on resources


danielrbradley:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

In version 2 of the provider we're removing selected resources to reduce SDK size where the next version of the resource is compatible with the previous verion. In theory API behaviour can vary between versions and a user might want to use a specific version.

To ensure that the provider allows a user to interact with the full Azure surface area, we could add a string `ApiVersion` field to all resources which would allow the same SDK class to be used with other versions. 

Initially this could be just a string field but could also be implemented to include an enum of known versions for that resource. 

If the user specifies an invalid API version or a version that's incompatible with the resources shape of the SDK resource class then errors would just be propegated by the Azure API - we would not validate this. We could also consider disabling validation of properties if the version is specified in case transformations are also being used to manipulate the shape of the payload.

### Affected area/feature

- Schema generation
- Provider

thomas11:
resolves #1877

danielrbradley:
@thomas11 I don't think this does solve the specific use case in #1877 - that is for resources which don't exist in any way in the SDK rather than a different version of an existing resource.

thomas11:
@danielrbradley the "resource doesn't exist in the SDK" case is covered by #68. Whereas #1877 is explicitly about API versions - although the semantics of the use case may differ. We should clarify this.

danielrbradley:
Discussion points with @mikhailshilkov 

1. Consider property naming to make it clear it's an "escape hatch". E.g. `OverrideApiVersion`
2. Add to the 2.0 migration guide for maintaining any v1 resource versions.
3. Add to the provider documentation to explain how it works & what it's for. Include details about using transformations to modify payload if required.

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3552

EronWright:
I would advocate for closing this ticket in favor of: https://github.com/pulumi/pulumi-azure-native/issues/68

Is anyone able to provide a use case that would really be solved by adding an override api version?

danielrbradley:
TODO: Re-spec this as a hidden escape hatch accessible via transforms.

danielrbradley:
## Updated design

[...]

PR #2637, opened 2023-08-04
Keep required resources

### LINKED ISSUE #2636: V2 is missing explicit resources with known usage
Similar to https://github.com/pulumi/pulumi-azure-native/issues/2634

We have a list of explicit resources that have known usage but we are not using this to retain the resources in the v2 schema.

danielrbradley:
Fixes #2636 

We have a list of explicit resources which we know are used. We should be ensuring these resources are not removed.

- Fix the parsing of the text file (line breaks were broken).
- Turn lines into a string set for efficient searching.
- This adds around 600 new resources.
- Update conversion tests.


danielrbradley:
Closing as going to introduce a version resource property instead to allow access to specific preview versions:
- https://github.com/pulumi/pulumi-azure-native/issues/2467

Issue #2669, opened 2023-08-22
Migration azure-native to v2 leads to ManagedCluster replacing


vladimirpodr:
### What happened?

I am migrating from azure-native provider v1 to v2.0.3. Typescript. And for containerservice.ManagedCluster resource sku name 'Basic'  is deprecated, and I use the name value 'Base'. so that I don't do it anyway Pulumi tries to replace/redeploy cluster resource.
 I changed several deprecated parameters/values but the issue with SKU only. Look at diff outputs:
I have tried to update provider in two steps: 1.70 -> 1.104 (latest minor release) -> 2.0.3. But result the same:
```
% pulumi preview --diff

    +-azure-native:containerservice:ManagedCluster: (replace)
        [id=/subscriptions/**********/resourcegroups/rg-service-platform-sandbox/providers/Microsoft.ContainerService/managedClusters/aks-sandbox]
        [urn=urn:pulumi:sandbox::service-platform::azure-native:containerservice:ManagedCluster::aks-sandbox]
        [provider=urn:pulumi:sandbox::service-platform::pulumi:providers:azure-native::default_2_3_0::5c8310bc-fc8e-48d2-811b-d7f3ad4a2cc9]
      ~ identity         : {
          - userAssignedIdentities: {
              - /subscriptions*************/resourcegroups/rg-service-platform-sandbox/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aks-sandbox-identitycd3cea73: {}
            }
          + userAssignedIdentities: [
          +     [0]: "/subscriptions/**************/resourcegroups/rg-service-platform-sandbox/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aks-sandbox-identitycd3cea73"
            ]
        }
      ~ networkProfile   : {
          - dockerBridgeCidr: "172.17.0.0/16"
        }
      ~ sku              : {
          ~ name: "Basic" => "Base"
          ~ tier: "Paid" => "Standard"
        }
 ```

     

### Expected Behavior

pulumi up without diffs

### Steps to reproduce

change package version "@pulumi/azure-native": "^1.0.0" -> "@pulumi/azure-native": "2.3.0"
npm install
pulumi refresh
pulumi up

### Output of `pulumi about`

```
% pulumi about
CLI          
Version      3.68.0
Go Version   go1.20.4
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host     
OS       darwin
Version  13.4.1
[...]

Issue #2600, opened 2023-07-23
Unable to migrate containerservice to v2: Basic managed cluster SKU name is invalid. 'Basic' has been replaced by 'Base' since v2023-02-01. 


dpunkturban:
### What happened?

We are currently trying to update our azure projects to V2 with the upgrade guide, but we are currently struggling with azure-kubernetes.

We updated the libs and see only 1 change (networkProfile), so it looks good.

                      ~ networkProfile: {
                          - dockerBridgeCidr: "172.17.0.1/16"
                        }                                                                                                                                                                                                                                             Resources:  
    ~ 1 to update      
    12 unchanged

But when we try to run pulumi, we get the following error message:

```
error: Code="SkuNotAvailable" Message="Basic managed cluster SKU name is invalid. 'Basic' has been replaced by 'Base' since v2023-02-01. "
```

We deployed the clusters with “Basic” SkuName and “Paid” SkuTier, which looks like was renamed.

When I check the cluster via az cli, the new name is already applied.

```bash
az aks show --resource-group RG --name CN | jq '.sku'
{
  "name": "Base",
  "tier": "Standard"
}
```

We deployed aks with the following version of containerservice:

```
 github.com/pulumi/pulumi-azure-native-sdk/containerservice v1.103.0 
 ```


If I simply try to replace Basic with Base and Paid with Standard in the code, Pulumi wants to replace the cluster, which we don’t want :confused:

I also tried to export the stack, replace the values and import the stack again, but it does not work.


### Expected Behavior

There is a migration path to change the SkuName and SkuTier without replacing the entire aks cluster.

### Steps to reproduce

Spin up aks with:

SkuName: Basic
SkuTier: Paid

```
github.com/pulumi/pulumi-azure-native-sdk/containerservice v1.103.0 
```

Upgrade dependency to:

```
	github.com/pulumi/pulumi-azure-native-sdk/containerservice/v2 v2.1.1
```

Try to run `pulumi up` 



### Output of `pulumi about`

```
CLI          
Version      3.76.0
Go Version   go1.20.6
Go Compiler  gc

Plugins
NAME          VERSION
azure-native  2.1.1
azure-native  2.1.1
azure-native  2.1.1
azure-native  2.1.1
azure-native  2.1.1
azure-native  2.1.1
azure-native  2.1.1
azuread       5.38.0
go            unknown
kubernetes    3.29.1
random        4.13.2
tls           4.10.0

Host     
OS       darwin
[...]

PR #2688, opened 2023-08-29
Custom diff handling for Manager Cluster SKU

mikhailshilkov:
Provides a workaround for diffing of AKS manager cluster SKUs across API versions. Diff override is similar to the existing location diffing, but it's looking for a `sku` property with `name` and `tier` sub properties, both of which match a hard-coded list of know equivalent values.

Fixes https://github.com/pulumi/pulumi-azure-native/issues/2600
Fixes https://github.com/pulumi/pulumi-azure-native/issues/2669

### Testing

1. Added a unit test to check the diff calculation.
2. Ran a manual test of upgrading a managed cluster from api `v20210501` to the default 2.0 version.

### DIFF
--- a/provider/pkg/provider/diff.go
+++ b/provider/pkg/provider/diff.go
@@ -223,6 +223,12 @@ func applyAzureSpecificDiff(diff *resource.ObjectDiff) {
 				continue
 			}
 		}
+		// Another special case is "sku" in AKS clusters.
+		if string(k) == "sku" && v.Old.IsObject() && v.New.IsObject() {
+			if sameManagedClusterSku(v.Old.ObjectValue(), v.New.ObjectValue()) {
+				continue
+			}
+		}
 		updates[k] = v
 	}
 	diff.Updates = updates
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
Happy for this tactical fix to go in, on the condition we have an issue written up to track removing this and what it would need to be replaced with in the future.

smcmanis:
Unfortunately this missed the case I was encountering. Which was: 
`sku`: `Base` -> `Basic`
`tier`: `Free` -> `Free` (unchanged)

mikhailshilkov:
@smcmanis Sorry for the trouble! Let me add that test case.

tap1992:
Is there anything else that needs to be done for this workaround to work? I've updated to v2.6.0 which seems to include this fix but I'm still seeing Pulumi trying to recreate our cluster due to the below (this is using the default API in the provider):

~ sku                      : {
              ~ name: "Basic" => "Base"

mikhailshilkov:
@tap1992 What is the tier of your cluster?

tap1992:
Free

mikhailshilkov:
Opened https://github.com/pulumi/pulumi-azure-native/pull/2746 to cover the Free use case

PR #2695, opened 2023-08-31
Ignore additional enum values which only vary by case

### LINKED ISSUE #2694: Nightly upgrade failed
https://github.com/pulumi/pulumi-azure-native/actions/runs/6020761074/job/16332772320?pr=2690

Dotnet build failing with:

```
Error: /home/runner/work/pulumi-azure-native/pulumi-azure-native/sdk/dotnet/KeyVault/Enums.cs(284,51): error CS0102: The type 'KeyRotationPolicyActionType' already contains a definition for 'Rotate' [/home/runner/work/pulumi-azure-native/pulumi-azure-native/sdk/dotnet/Pulumi.AzureNative.csproj]
Error: /home/runner/work/pulumi-azure-native/pulumi-azure-native/sdk/dotnet/KeyVault/V20230201/Enums.cs(284,51): error CS0102: The type 'KeyRotationPolicyActionType' already contains a definition for 'Rotate' [/home/runner/work/pulumi-azure-native/pulumi-azure-native/sdk/dotnet/Pulumi.AzureNative.csproj]
```

### SDK Change

```diff
+         /// <summary>
+         /// Rotate the key based on the key policy. Key Vault only. Managed HSM uses camelCase 'rotate' instead.
+         /// </summary>
+         public static KeyRotationPolicyActionType Rotate { get; } = new KeyRotationPolicyActionType("Rotate");
+         /// <summary>
+         /// Rotate the key based on the key policy. Managed HSM only. Key Vault uses PascalCase 'Rotate' instead.
+         /// </summary>
          public static KeyRotationPolicyActionType Rotate { get; } = new KeyRotationPolicyActionType("rotate");
```

It's adding a second property called `Rotate` except that the property value has Title casing.

### Upstream Change

- https://github.com/Azure/azure-rest-api-specs/pull/24475

```diff
        "enum": [
+           "Rotate",
            "rotate",
-           "notify"
+           "Notify"
        ],
```

Looks like the case was changed to be title case but the lowercase `rotate` was left in.

## Actions

- [x] Confirm codegen expects values to not vary by only case.
- [x] Open issue to codify this with a schema validation check to fail fast with a helpful message.
  - https://github.com/pulumi/pulumi/issues/13840
- [x] This looks like a mistake on the part of Microsoft, so we should report this upstream.
  - Commented back on the PR where there's also discussion from Hashicorp who have been broken by this change.
- [x] Fix by ignoring subsequent enum values which already exist when ignoring case
  - https://github.com/pulumi/pulumi-azure-native/pull/2695

danielrbradley:
Fixes #2694

### DIFF
--- a/provider/pkg/gen/types_test.go
+++ b/provider/pkg/gen/types_test.go
@@ -1,0 +1,37 @@
+package gen
+
+import (
+	"testing"
+
+	"github.com/go-openapi/spec"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/openapi"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestEnumCaseCollapsing(t *testing.T) {
+	generator := moduleGenerator{
+		pkg: &schema.PackageSpec{
+			Name:  "azure-native",
+			Types: map[string]schema.ComplexTypeSpec{},
+		},
+		module: "MyModule",
+	}
+	// Doesn't return the enum - it's added to the Types map
+	_, err := generator.genEnumType(&spec.Schema{
+		SchemaProps: spec.SchemaProps{
+			Enum: []interface{}{
+				"foo",
+				"Foo",
+				"bar",
+			},
+		},
+	}, &openapi.ReferenceContext{}, map[string]interface{}{"name": "myProp"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Len(t, generator.pkg.Types, 1)
+	assert.Len(t, generator.pkg.Types["azure-native:MyModule:MyProp"].Enum, 2)
+	assert.Equal(t, generator.pkg.Types["azure-native:MyModule:MyProp"].Enum[0].Value, "foo")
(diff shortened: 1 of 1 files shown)


provider/pkg/gen/types.go
```diff
@@ -277,7 +277,17 @@ func (m *moduleGenerator) genEnumType(schema *spec.Schema, context *openapi.Refe
 			if strings.HasPrefix(m.module, "datafactory") && enumName == "ScriptActivityParameterDirection" {
 				enumVal.Name = fmt.Sprintf("Value%s", val)
 			}
-			enumSpec.Enum = append(enumSpec.Enum, enumVal)
+			// Ignore additional enum values that only vary by case as this isn't supported by Pulumi.
+			exists := false
+			for _, existing := range enumSpec.Enum {
```
thomas11, line 282: This loop will make the first enum win. I didn't trace `enumSpec.Enum` all the way back, but if it isn't sorted, that might introduce non-determinism.

thomas11 (approved):
LGTM pending the comment

provider/pkg/gen/types.go
danielrbradley: As far as I could see it's always kept as a list so should be consistent every time.

Issue #2691, opened 2023-08-30
Add unit tests for codegen


danielrbradley:
Rough plan:
1. Ensure we can run codegen in a unit test
2. Use snapshot testing for fast end-to-end coverage
3. Design test suite to cover key codegen features

Possible areas of testing:
- Parsing specifications
- Translation into Pulumi schema
- Calculation and application of default versions
- Parsing examples
- Generating example PCL
- Deprecation messages
- Version exclusion (normal and variants)
- Type merging for default version

iwahbe:
When we say "codegen", what do we mean here? Does this mean testing schema generation?

danielrbradley:
@iwahbe yup - testing the internal parts of the schema generation process including:
1. Parsing specifications
2. Translation into Pulumi schema
3. Parsing examples
4. Generating example PCL

danielrbradley:
Closing this out as complete for now as we've managed to:
1. Refactor to allow testing of:
    1. Specs parsing
    2. Versioning
    3. Schema + Metadata construction
    4. Example generation
2. Increased coverage by >10% in #2755 
3. Extract reports for better tracking of notable behaviour

PR #2727, opened 2023-09-07
Refactor deprecation messages into version interface

danielrbradley:
Related to #2691 

GenerateUpgradeWarnings touches the file system directly so we want to do this outside of the generation process so we can mock and test the core generation code.

1. Rename gen.Constraints to gen.Versioning
2. Add GetDeprecation to Versioning interface
3. Load NextResourcesToRemove as part of VersionSources to implement GetDeprecation
4. Use GetDeprecation in schema.go and format the message there.
5. Remove old upgradePath implementation

Aside: replace ioutil.ReadAll to fix warnings


provider/pkg/versioning/gen.go
```diff
@@ -28,6 +28,10 @@ type VersionMetadata struct {
 	RemovedInvokes                ResourceRemovals
 }
 
+// Ensure our VersionMetadata type implements the gen.Constraints interface
```
thomas11, line 31: "the gen._Versioning_ interface" now, right?

provider/pkg/versioning/gen.go
```diff
@@ -56,6 +60,16 @@ func (v VersionMetadata) ShouldInclude(provider string, version string, typeName
 	return true
 }
 
+func (v VersionMetadata) GetDeprecationMessage(token string) (string, bool) {
```
thomas11, line 63: In terms of responsibilities and separation of concerns, it would make more sense to me to have an `IsDeprecated bool` method in the version metadata, and leave the message and formatting to the caller.

provider/pkg/gen/schema.go
```diff
@@ -40,12 +40,13 @@ import (
 // Note - this needs to be kept in sync with the layout in the SDK package
 const goBasePath = "github.com/pulumi/pulumi-azure-native-sdk/v2"
 
-type Constraints interface {
+type Versioning interface {
 	ShouldInclude(provider string, version string, typeName, token string) bool
+	GetDeprecationMessage(token string) (string, bool)
```
thomas11, line 45: A little odd and potentially confusing that one of the two methods has a `token` argument that's not fully qualified, and the other has one that is.

provider/pkg/versioning/gen.go
danielrbradley: In addition to indicating if the resource is deprecated, internal to the Metadata, we have information on the new token that the resource should be upgraded to.

I guess the tweak would be to return something like `({ newToken string }, bool)` rather than just a plain string which is the message - so we can move the formatting of the message into the schema.go?

provider/pkg/gen/schema.go
danielrbradley: This is because the inclusion is evaluated on both the originating provider/version/type, and the generated token.

Guess we could add all the other variables to the GetDeprecationMessage for consistency, but they would just be unused.

Another option would be to combine the two methods into a single call, but with a return type something like 
`({ include: bool, deprecatedByToken: string })` though they're called from different places and each property is looked up from a different data source, so that would make it less efficient as it would introduce duplication of effort.

I'd hope this area will be refactored further as we look at the whole default version pipeline and where we go about applying these.

thomas11: I think I misunderstood and thought these were different kinds of tokens. If not, my point is moot.

thomas11 (approved):

PR #2728, opened 2023-09-07
Convert console logs into reports

danielrbradley:
Related to #2691 

In preparation for improving unit testing, move all generation process console logging and warning into reports.

This has the added benefit of highlighting when warnings change during upgrades.

Stacked on top of #2727

### DIFF
--- a/provider/pkg/openapi/discover.go
+++ b/provider/pkg/openapi/discover.go
@@ -179,6 +179,10 @@ func ReadAzureProviders(specsDir, namespace, apiVersions string) (AzureProviders
 	// Collect all versions for each path in the API across all Swagger files.
 	providers := AzureProviders{}
 	for _, location := range swaggerSpecLocations {
+		relLocation, err := filepath.Rel(specsDir, location)
+		if err != nil {
+			return nil, diagnostics, errors.Wrapf(err, "failed to get relative path for %q", location)
+		}
 		swagger, err := NewSpec(location)
 		if err != nil {
 			return nil, diagnostics, errors.Wrapf(err, "failed to parse %q", location)
@@ -190,7 +194,7 @@ func ReadAzureProviders(specsDir, namespace, apiVersions string) (AzureProviders
 		}
 		sort.Strings(orderedPaths)
 		for _, path := range orderedPaths {
-			namingDisambiguations := providers.addAPIPath(location, path, swagger)
+			namingDisambiguations := providers.addAPIPath(specsDir, relLocation, path, swagger)
 			diagnostics.NamingDisambiguations = append(diagnostics.NamingDisambiguations, namingDisambiguations...)
 		}
 	}
@@ -321,14 +325,14 @@ var excludeRegexes = []*regexp.Regexp{
 
 // addAPIPath considers whether an API path contains resources and/or invokes and adds corresponding entries to the
 // provider map. `providers` are mutated in-place.
-func (providers AzureProviders) addAPIPath(fileLocation, path string, swagger *Spec) []resources.NameDisambiguation {
+func (providers AzureProviders) addAPIPath(specsDir, fileLocation, path string, swagger *Spec) []resources.NameDisambiguation {
 	for _, re := range excludeRegexes {
 		if re.MatchString(fileLocation) {
 			return nil
 		}
 	}
 
-	prov := resources.ResourceProvider(fileLocation, path)
+	prov := resources.ResourceProvider(filepath.Join(specsDir, fileLocation), path)
 	if prov == "" {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


provider/cmd/pulumi-gen-azure-native/main.go
```diff
@@ -82,9 +82,20 @@ func main() {
 		}
 
 		if namespaces == "*" && apiVersions == "" {
-			if err = buildSchemaResult.Version.WriteTo("versions"); err != nil {
+			written, err := buildSchemaResult.Version.WriteTo("versions")
+			if err != nil {
 				panic(err)
 			}
+			for _, v := range written {
+				fmt.Printf("Emitted %q\n", v)
```
thomas11, line 90: [nit] why print the path as a Go-escaped string? It might not be copy-passable into the shell or other tools.

provider/pkg/resources/resources.go
```diff
@@ -262,38 +261,34 @@ func ResourceName(operationID, path string) string {
 	// Manual override to resolve ambiguity between public and private RecordSet.
 	// See https://github.com/pulumi/pulumi-azure-native/issues/583.
 	// To be removed with https://github.com/pulumi/pulumi-azure-native/issues/690.
-	if resourceName == "RecordSet" && strings.Contains(path, "privateDns") {
+	if resourceName == "RecordSet" && strings.Contains(path, "/providers/Microsoft.Network/privateDnsZones/") {
 		newName := "PrivateRecordSet"
-		log.Printf("Disambiguating %s at %s to %s", resourceName, path, newName)
```
thomas11, line 267: Are these "disambiguating..." messages just dropped? Or am I missing where they go into a report?

If they're dropped that'll bite us later when we're debugging these resources.

provider/cmd/pulumi-gen-azure-native/main.go
danielrbradley: In practice this just wraps paths in quotations for us - doesn't escape anythinig. This matches the other "emmitted" logs. But yeah, could just drop the quote marks everywhere.

provider/pkg/resources/resources.go
danielrbradley: Ok, I'm still bashing my head against this one. My problem with these messages is that they're basisically saying "I'M WORKING EXACTLY AS INTENDED" in every run which isn't helpful for highlighting what needs to be done. In fact the only time an action would likely need to be taken is if these _weren't printed_ - indicating that they're changed something to make us miss e.g. casing change.

Moving these out to reports is probably the answer as it would highlight well in the diff. What put me off before is because this function is quite deep and we have no global context. The only option here would be to attach all the functions in the call stack to a context object so we can write out to the report.

provider/cmd/pulumi-gen-azure-native/main.go
```diff
@@ -82,9 +82,20 @@ func main() {
...
+			}
+			written, err = buildSchemaResult.Reports.WriteTo("reports")
+			if err != nil {
+				panic(err)
+			}
+			for _, v := range written {
+				fmt.Printf("Emitted %s\n", v)
+			}
 		} else {
 			fmt.Println("Note: skipping writing version metadata because DEBUG_CODEGEN_NAMESPACES or DEBUG_CODEGEN_APIVERSIONS is set.")
```
thomas11, line 100:
```suggestion
			fmt.Println("Note: skipping writing version metadata and reports because DEBUG_CODEGEN_NAMESPACES or DEBUG_CODEGEN_APIVERSIONS is set.")
```

provider/pkg/openapi/discover.go
```diff
@@ -355,7 +362,7 @@ func addAPIPath(providers AzureProviders, fileLocation, path string, swagger *Sp
 	}
 
 	pathItemList, hasList := swagger.Paths.Paths[path+"/list"]
-
+	var nameDisambiuations []resources.NameDisambiguation
```
thomas11, line 365:
```suggestion
	var nameDisambiguations []resources.NameDisambiguation
```

reports/namingDisambiguations.json
```diff
@@ -0,0 +1,758 @@
+[
+  {
```
thomas11, line 2: This file has duplicates. Probably we look up each resource multiple times and simply append each disambiguation?

thomas11 (approved):
Would be good to address the disambiguation duplicates, otherwise LGTM

reports/namingDisambiguations.json
danielrbradley: I think it's related to the versions of the API, but we don't have enough context in the function to include this. I'd rather keep the duplicates but add the version or related file path.

provider/cmd/pulumi-gen-azure-native/main.go
danielrbradley: Thanks - added this in the update to the outputs commit!

reports/namingDisambiguations.json
danielrbradley: I've updated the implementation to add the relative path for each entry which makes it loads more useful!

provider/pkg/openapi/discover.go
danielrbradley: Done thanks!

Issue #2476, opened 2023-05-25
[codegen] - Enable generating pyproject.toml files in order to publish Python wheels


kpitzen:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Pulumi codegen is adding the opt-in ability to generate `pyproject.toml` files in order to facilitate generating wheels as a part of python SDK generation. We should enable that feature once it's available and ship wheels for our provider Python packages

Pulumi PR:
- [ ] https://github.com/pulumi/pulumi/pull/12805

Required work:
Modify this provider's schema to set `pyproject.enabled` to `true`

### Affected area/feature

SDK Codegen

t0yv0:
Blocked on https://github.com/pulumi/pulumi-azure-native/issues/2726

t0yv0:
Actually blocked on https://github.com/pulumi/pulumi/issues/13921

The small things I can resolve within the repo but the above issue means we're losing docstrings; we should either accept that or wait for a fix in platform.

PR #2730, opened 2023-09-07
Roll out Python wheel build and publishing

t0yv0:
Fixes https://github.com/pulumi/pulumi-azure-native/issues/2476

As part of https://github.com/pulumi/home/issues/2883 we are enabling:

- generating pyproject.toml instead of setup.py
- building wheels alongside source distributions
- making sure both wheels and source distributions are published on pypi

If all goes well this feature-* PR should result in a PyPI version that can be checked to have both sdist and wheel distros published.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -155,7 +155,7 @@ jobs:
           twine upload
           -u "pulumi"
           -p "${PYPI_PASSWORD}"
-          "${{ github.workspace }}/sdk/python/bin/dist/*.tar.gz"
+          "${{ github.workspace }}/sdk/python/bin/dist/*"
           --skip-existing
           --verbose
 


t0yv0:
There is something *very* suspicious in [8eb98e4](https://github.com/pulumi/pulumi-azure-native/pull/2730/commits/8eb98e46491adf6678b660531a8f8026b422299d) I can't explain. Honestly I think we should maybe try to upgrade pulumi version to 3.81.0 in a separate PR before rolling this out.  I've not worked much in this repo yet and wonder if there's any pointers on how to do it right or there's an automated job that updates it.

t0yv0:
https://github.com/pulumi/pulumi-azure-native/actions/runs/6109850011 ah and so the update has recently failed. Let's consider this PR blocked on that.

PR #2737, opened 2023-09-08
Use coverpkg=./... to pull in coverage of more files

mjeffryes:
By default, Go coverage reports ignore files in directories with no tests. Setting coverpkg=./... tells go that you want coverage on all go files under the current directory


mjeffryes:
Just waiting for a review to land this one. Should reflect coverage for all of the provider directory now

mikhailshilkov (approved):

Issue #2734, opened 2023-09-08
Replace azidentity's certificate parsing with go-pkcs12 [was: Old version of hashicorp/go-azure-helpers causing certificate authentication issues]


floyd-may:
### What happened?

Unable to authenticate using a client certificate; error message `the Client Certificate Path is not a valid pfx file: pkcs12: unknown digest algorithm: 2.16.840.1.101.3.4.2.1]`. There is an closed terraform bug on this here:

https://github.com/hashicorp/terraform-provider-azurerm/issues/16228

If you follow the chain, the difference appears to be that `pulumi-azure-native` is using an older version of `terraform/go-azure-helpers` that has since been updated to properly handle a wider variety of certificates. 

### Expected Behavior

Authentication equivalent to the terraform azure module.

### Steps to reproduce

see above

### Output of `pulumi about`

irrelevant

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikhailshilkov:
Thank you for the note @floyd-may , I opened a PR to upgrade it

floyd-may:
Thank you for jumping on that so quickly @mikhailshilkov! Can you point me in the right direction of tracing that PR to a release? Would like to know when I can abandon my workaround.

mikhailshilkov:
@floyd-may It turns out the update is not as straightforward, since 0.52 of `terraform/go-azure-helpers` had a breaking change that we can't easily get past. Do you know which version had the fix that you need?

floyd-may:
@mikhailshilkov Not surprised that there's a breaking change there 😖 Appreciate the attempt, though! I don't know that I can give you a precise version, but I did see [this commit](https://github.com/hashicorp/go-azure-helpers/commit/db006dddcee231e0f3250c54762218aeaa3f7219) that seems to indicate some pretty significant changes related to authentication in `go-azure-helpers`. The commit message seems to indicate that there was some wheel-reinvention and that the better approach was to defer to auth functionality in go-azure-sdk.

thomas11:
We now have a new authentication backend using Azure's own [azidentity]() package (currently behind the feature toggle `PULUMI_ENABLE_AZCORE_BACKEND=true`). Unfortunately, the new package has the same limitation.

> [...] it can't decrypt keys in PEM format or PKCS#12 certificates that use SHA256 for message authentication

[...]

PR #2738, opened 2023-09-09
Update go-azure-helpers version

mikhailshilkov:
Fix https://github.com/pulumi/pulumi-azure-native/issues/2734


mikhailshilkov:
It's not as easy as I thought, I'll have to do it differently...

PR #2743, opened 2023-09-12
Upgrade actions/install versions and remove obsolete exceptions

thomas11:
A quick drive-by cleanup as I was borrowing this actions for some other repos.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -45,3 +45,5 @@ runs:
     - name: Setup Gradle
       if: inputs.skip_dotnet_and_java != 'true'
       uses: gradle/gradle-build-action@v2
+      with:
+        gradle-version: "8.3"


.github/actions/install/action.yml
```diff
@@ -49,6 +45,3 @@ runs:
     - name: Setup Gradle
       if: inputs.skip_dotnet_and_java != 'true'
       uses: gradle/gradle-build-action@v2
-      with:
-        # Pin to known good version until #2262 is resolved
-        gradle-version: "7.6"
```
danielrbradley, line 54: I wonder if we should still pin gradle just for consistency of builds, but upgrade it.

thomas11: Right, good point, not to be surprised like last time.

danielrbradley (approved):

PR #2747, opened 2023-09-12
Add special case to fix nightly failure

danielrbradley:
ScVmm has just added a new version which adds two new resources, but one of the shared types has also changed to make two fields required. Remove the additional required fields to ensure the type is compatible with the existing version we're using.

Fixes https://github.com/pulumi/pulumi-azure-native/issues/2742


thomas11:
So next time we bump the default version of this resource, this change can and should be removed?

danielrbradley:
@thomas11 yup - in theory all these special cases should go. I'll open an issue with a 3.0 label

thomas11 (approved):

PR #2755, opened 2023-09-13
Maintain nested resources which are also settable on the parent resource

### LINKED ISSUE #611: Virtual Network expects all subnets to be defined inline, causes deletion-recreation churn otherwise
A repository demonstrating the following can be found at https://github.com/MerelyAPseudonym/subnet-churn-repro.

I expected the familiar all-or-nothing constraint regarding inline subnet definitions to carry over to the nextgen provider, but even if I _only_ use standalone subnet resources I'm observing unstable behavior. The initial run works, but then **without changing the code** subsequent runs delete and recreate the subnet.

Initialize both network resources:
```sh-session
$ pulumi up --refresh --non-interactive --yes --skip-preview --suppress-permalink
Updating (dev)

 ~  pulumi:pulumi:Stack subnet-churn-repro-dev refreshing 
 ~  pulumi:pulumi:Stack subnet-churn-repro-dev refreshing 
 ~  azure-nextgen:resources/v20200601:ResourceGroup resourceGroup refreshing 
 ~  azure-nextgen:resources/v20200601:ResourceGroup resourceGroup refresh 
    pulumi:pulumi:Stack subnet-churn-repro-dev running 
    azure-nextgen:resources/v20200601:ResourceGroup resourceGroup  
 +  azure-nextgen:network/v20200601:VirtualNetwork network creating 
 +  azure-nextgen:network/v20200601:VirtualNetwork network created 
 +  azure-nextgen:network/v20200601:Subnet subnet creating 
 +  azure-nextgen:network/v20200601:Subnet subnet created 
@ Updating....
    pulumi:pulumi:Stack subnet-churn-repro-dev  
 
Resources:
    + 2 created
    2 unchanged
```

Double-check that the subnet exists via the Azure CLI:
```sh-session
$ az network vnet subnet show --resource-group=josh-repro --vnet-name=net --name=net-sub1
{
  "addressPrefix": "10.0.0.0/24",
…
```

Rerun Pulumi without changing the code:
```sh-session
$ pulumi up --refresh --non-interactive --yes --skip-preview --suppress-permalink
Updating (dev)

 ~  pulumi:pulumi:Stack subnet-churn-repro-dev refreshing 
 ~  pulumi:pulumi:Stack subnet-churn-repro-dev refreshing 
 ~  azure-nextgen:network/v20200601:Subnet subnet refreshing 
 ~  azure-nextgen:resources/v20200601:ResourceGroup resourceGroup refreshing 
 ~  azure-nextgen:network/v20200601:VirtualNetwork network refreshing 
 ~  azure-nextgen:resources/v20200601:ResourceGroup resourceGroup refresh 
 ~  azure-nextgen:network/v20200601:VirtualNetwork network refresh [diff: ~__inputs,etag,subnets]
 ~  azure-nextgen:network/v20200601:Subnet subnet refresh 
    pulumi:pulumi:Stack subnet-churn-repro-dev running 
    azure-nextgen:resources/v20200601:ResourceGroup resourceGroup  
 ~  azure-nextgen:network/v20200601:VirtualNetwork network updating [diff: -subnets]
 ~  azure-nextgen:network/v20200601:VirtualNetwork network updated [diff: -subnets]
    azure-nextgen:network/v20200601:Subnet subnet  
    pulumi:pulumi:Stack subnet-churn-repro-dev  
 
Resources:
    ~ 1 updated
    3 unchanged
```

Run the same Azure CLI command, finding the subnet absent:
```sh-session
$ az network vnet subnet show --resource-group=josh-repro --vnet-name=net --name=net-sub1
ResourceNotFoundError: Resource /subscriptions/123e4567-e89b-12d3-a456-426614174000/resourceGroups/josh-repro/providers/Microsoft.Network/virtualNetworks/net/subnets/net-sub1 not found.
```

Again, without changing the code, rerun Pulumi:
```sh-session
$ pulumi up --refresh --non-interactive --yes --skip-preview --suppress-permalink
Updating (dev)

 ~  pulumi:pulumi:Stack subnet-churn-repro-dev refreshing 
 ~  pulumi:pulumi:Stack subnet-churn-repro-dev refreshing 
 ~  azure-nextgen:resources/v20200601:ResourceGroup resourceGroup refreshing 
 ~  azure-nextgen:network/v20200601:Subnet subnet refreshing 
 ~  azure-nextgen:network/v20200601:VirtualNetwork network refreshing 
 ~  azure-nextgen:resources/v20200601:ResourceGroup resourceGroup refresh 
 ~  azure-nextgen:network/v20200601:Subnet subnet refresh 
 ~  azure-nextgen:network/v20200601:VirtualNetwork network refresh 
    pulumi:pulumi:Stack subnet-churn-repro-dev running 
    azure-nextgen:resources/v20200601:ResourceGroup resourceGroup  
    azure-nextgen:network/v20200601:VirtualNetwork network  
 +  azure-nextgen:network/v20200601:Subnet subnet creating 
 +  azure-nextgen:network/v20200601:Subnet subnet created 
    pulumi:pulumi:Stack subnet-churn-repro-dev  
 
Resources:
    + 1 created
    3 unchanged
```

It just goes back and forth.

### LINKED ISSUE #1112: Parent-child resource relationship awareness
This issue tracks an umbrella use case when two resources are in parent-child relationship and manage the same piece of configuration in ARM. It works like this:

The parent resource `/foo/fooName` that manages `bars` as a property:

```ts
const foo = new Foo("fooName", {
  someParentProp: "value1",
  bars: [{
    name: "barName",
    buzz: "123",
  }],
});
```

However, there is a separate ARM resource `/foo/fooName/bar/barName`, so the bar can be managed as

```ts
const bar = new Bar("barName", {
  buzz: "123",
});
```

The problem is that these resources are completely independent in Pulumi state but map to the same thing in Azure state. For example, if `bars` were to be omitted in the first snippet above, Pulumi state would have them empty but the second snippet would set its value in Azure. When a user runs `pulumi refresh`, that state is read from Azure and the users sees a diff. Re-running the program after the refresh would then make the `Foo` resource to try to clean the collection of `bars` thus deleting the state of `Bar` and causing another refresh diff.

We should likely introduce this knowledge as a first-class concept in the Azure Native provider. It should either keep the properties in sync automatically, or omit the parent properties and require all the management to be done via child resources.

Examples of concrete issues:

- https://github.com/pulumi/pulumi-azure-native/issues/611
- https://github.com/pulumi/pulumi-azure-native/issues/1087
- https://github.com/pulumi/pulumi-azure-native/issues/688

danielrbradley:
Resolves #611 
Resolves #1112 

1. Add metadata to indicate which properties need maintaining.
2. Before updating a resource, read remote state of nested resource properties and maintain the existing value.
3. When reading, avoid ignored properties from being added to the user's inputs.

## Identification of nested resources

1. For each resource spec, find all nested resource specs (where the path is within the current resource path)
2. For each nested resource spec, find the ref for the body payload.
4. When generating properties of the parent, tag any property which is a list of a nested resource body.
    - Note: We only do this for input properties.

The result is tagging any parent resource property which is a list of child resources (matching by type reference). 

## Provider behaviour changes

During an update, after generating the request body based on the user's inputs:

1. Find properties (in the metadata) directly within the body which are tagged for `IgnoreIfUnset`.
    - Note: We don't recurse the whole type structure for nested `IgnoreIfUnset` properties.
3. Filter the properties to those which are unset.
5. If any properties match the criteria above, fetch the current state from the cloud.
6. Set the values of each matched property in the body from the recieved GET response.

During a refresh:

1. Create a copy of the SDK-converted GET response.
2. Remove any `IgnoreIfUnset` which were unset in the old inputs.

We continue to use the GET response with all properties included for the new outputs. 

Note: We don't exclude the `IgnoreIfUnset` during an import, but perhaps we could - and assume that users should import these resources independently.

### Performance effect

- For every resource update, we loop through each of its properties in the meta data.
- For resources with `IgnoreIfUnset` properties on their root body _and_ have these properties missing, we have an added roundtrip to the cloud provider.

## Affected Resources

The metadata shows 132 instances of `ignoreIfUnset` being added. 13 of these are in types which means they'll never affect behaviour.

Ignoring multiple versions of the same resources, these are the 24 unique resource properties affected:

```js
resources["azure-native:deviceupdate:Account"].privateEndpointConnections
resources["azure-native:digitaltwins:DigitalTwin"].privateEndpointConnections
resources["azure-native:eventhub:Namespace"].privateEndpointConnections
resources["azure-native:hardwaresecuritymodules:CloudHsmCluster"].privateEndpointConnections
resources["azure-native:network:ExpressRouteCircuit"].authorizations
resources["azure-native:network:ExpressRouteCircuit"].peerings
resources["azure-native:network:ExpressRouteCircuitPeering"].connections
resources["azure-native:network:ExpressRouteGateway"].expressRouteConnections
resources["azure-native:network:LoadBalancer"].backendAddressPools
resources["azure-native:network:LoadBalancer"].inboundNatRules
resources["azure-native:network:NetworkInterface"].tapConfigurations
resources["azure-native:network:NetworkSecurityGroup"].defaultSecurityRules
resources["azure-native:network:NetworkSecurityGroup"].securityRules
resources["azure-native:network:Profile"].endpoints
resources["azure-native:network:RouteFilter"].rules
resources["azure-native:network:RouteTable"].routes
resources["azure-native:network:ServiceEndpointPolicy"].serviceEndpointPolicyDefinitions
resources["azure-native:network:VirtualHub"].virtualHubRouteTableV2s
resources["azure-native:network:VirtualNetwork"].subnets
resources["azure-native:network:VirtualNetwork"].virtualNetworkPeerings
resources["azure-native:network:VirtualNetworkGateway"].natRules
resources["azure-native:network:VpnGateway"].connections
resources["azure-native:network:VpnGateway"].natRules
resources["azure-native:network:VpnServerConfiguration"].configurationPolicyGroups
resources["azure-native:networkfunction:AzureTrafficCollector"].collectorPolicies
resources["azure-native:powerbi:PowerBIResource"].privateEndpointConnections
resources["azure-native:relay:Namespace"].privateEndpointConnections
resources["azure-native:servicebus:Namespace"].privateEndpointConnections
resources["azure-native:synapse:Workspace"].privateEndpointConnections
```

### DIFF
--- a/examples/keyvault/index.ts
+++ b/examples/keyvault/index.ts
@@ -95,7 +95,7 @@ const key = new keyvault.Key("mykey", {
 const roleName = new random.RandomString("roleName", {
     length: 10,
     upper: false,
-    number: false,
+    numeric: false,
     special: false,
 }).result;
 


provider/pkg/gen/schema.go
```diff
@@ -1010,7 +1057,7 @@ func (m *moduleGenerator) escapeCSharpNames(typeName string, resourceResponse *p
 }
 
 func (m *moduleGenerator) genMethodParameters(parameters []spec.Parameter, ctx *openapi.ReferenceContext,
-	namer *resources.AutoNamer, bodySchema *openapi.Schema) (*parameterBag, error) {
+	namer *resources.AutoNamer, bodySchema *openapi.Schema, nestedResourceBodyRefs []string) (*parameterBag, error) {
```
thomas11, line 1060: new param isn't used?

thomas11 (commented):
Just a few first remarks so far.

provider/pkg/gen/schema.go
danielrbradley: Ah yup - good catch - that's a left over from a revert after I changed to attach the values to the `gen` context instead 👍

provider/pkg/provider/provider.go
```diff
@@ -1081,6 +1082,57 @@ func (k *azureNativeProvider) Read(ctx context.Context, req *rpc.ReadRequest) (*
...
+				if !oldInputs.HasValue(propKey) {
+					delete(result, key)
+				}
+			}
+		}
+	}
+	return result
+}
+
+func deepCopy(input map[string]interface{}) map[string]interface{} {
```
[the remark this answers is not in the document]
danielrbradley: Ah, maybe ... I looked but couldn't find anything like this. Will take a read of it 👍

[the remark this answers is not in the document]
danielrbradley: Yup, this looks good. Just had to add an extra cast in.

provider/pkg/resources/resources.go
```diff
@@ -61,6 +61,8 @@ type AzureAPIProperty struct {
 	IsStringSet bool `json:"isStringSet,omitempty"`
 	// Default is the default value for the parameter, if any
 	Default interface{} `json:"default,omitempty"`
+	// If the user hasn't specified this value, read it from the current cloud state.
+	IgnoreIfUnset bool `json:"ignoreIfUnset,omitempty"`
```
danielrbradley, line 65: Other ideas:
- RefreshIfUnset
- DontOverwriteIfUnset
- ReadIfUnset

provider/pkg/gen/properties.go
```diff
@@ -195,6 +195,15 @@ func (m *moduleGenerator) genProperty(name string, schema *spec.Schema, context
 	if err != nil {
 		return nil, nil, err
 	}
+	ignoreIfUnset := false
+	if !isOutput && schema.Items != nil && schema.Items.Schema != nil {
```
danielrbradley: Add comment: Why are we only considering arrays?

provider/pkg/provider/provider.go
```diff
@@ -1291,6 +1295,101 @@ func (k *azureNativeProvider) Cancel(context.Context, *pbempty.Empty) (*pbempty.
...
+		currentContainer := bodyParams
+		for _, containerName := range prop.Containers {
+			innerContainer, ok := currentContainer[containerName]
+			if !ok {
+				innerContainer = map[string]interface{}{}
+				currentContainer[containerName] = innerContainer
+			}
+			currentContainer = innerContainer.(map[string]interface{})
+		}
+		currentContainer[propName] = state[propName]
```
danielrbradley: TODO: Squash the testing & fix for state containers

provider/pkg/resources/resources.go
mikhailshilkov: ReadOnUpdateIfUnset?

danielrbradley: MaintainValueIfUnset?

mikhailshilkov (commented):
First batch of minor comments

provider/pkg/resources/resources.go
danielrbradley: MaintainSubresourceIfUnset?

danielrbradley: ++ comments

provider/pkg/provider/provider.go
```diff
@@ -1161,6 +1161,10 @@ func (k *azureNativeProvider) Update(ctx context.Context, req *rpc.UpdateRequest
 			return nil, azureError(err)
 		}
 	default:
+		err := k.readIgnoredPropertiesIfNotSet(ctx, &res, id, bodyParams)
+		if err != nil {
+			return nil, fmt.Errorf("failed reading state for ignored unset properties: %w", err)
```
danielrbradley, line 1208: Update based on new naming

provider/pkg/provider/provider.go
```diff
@@ -1081,6 +1083,46 @@ func (k *azureNativeProvider) Read(ctx context.Context, req *rpc.ReadRequest) (*
...
+	if len(propertiesToRemove) == 0 {
+		return sdkResponse
+	}
+
+	// Take a deep copy so we don't modify the original which is also used later for diffing.
+	copy := deepcopy.Copy(sdkResponse)
+	result, ok := copy.(map[string]interface{})
+	if !ok {
+		// This should never happen.
+		fmt.Println("warning: failed to cast cloned sdkResponse to map[string]interface{}, unable to remove ignored unset properties. Please report this issue.")
```
danielrbradley, line 1116: Add verbose log with same and value

provider/pkg/gen/gen_vnet_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/gkampitakis/go-snaps/snaps"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/openapi"
+)
+
+func TestVnetGen(t *testing.T) {
+	wd, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+	// We take a snapshot of the schema as we're focusing on testing how we parse the spec
```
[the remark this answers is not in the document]
danielrbradley: We want to use the same version as we currently have in default for best coverage, but don't want to break nightly upgrades any time network recieves an update. The purpose is to test the generation algorithm with some real-ish input - not to check we can still generate - that's done nightly when we regenerate the whole schema.

danielrbradley:
@thomas11 and @mikhailshilkov that should be all feedback addressed and two additional program tests added.

Tests are just failing due to some random provider issues.

thomas11 (approved):
LGTM!

EronWright:
In retrospect, it seems to me that the root cause of the problem was how the CRUD client updates the resource using PUT, but doesn't actually use the old object as the basis. Instead the client seems to formulate a put based purely on resource inputs, and is then surprised that sub-resources are summarily removed.  This PR took the awkward approach of identifying a subset of remote state that should be "maintained", rather than retaining all state in a systematic way.

PUT has a "replace" semantic and usually uses optimistic concurrency (i.e. an etag) to prevent data loss. I would have used a "roundtrip" strategy for updates:
1. Use the current state as the basis for the update.
2. Mutate the object from state with additions/removals/updates based on the program inputs.
3. PUT the updated object with an `If-Match` header to condition the update on the etag. _Update: am unable to find evidence that the Azure API supports optimistic concurrency._
4. Use the refresh command to update the current state to reflect remote state, to recover from etag mismatches.

In this way, the client would roundtrip the remote state and the sub-resources would survive.   Even fields not recognized by the client would survive. This would also support arrays (unlike the PR's solution) and would not require metadata. WDYT?

PR #2766, opened 2023-09-21
Upgrade pulumi dependency v3.79.0 -> v3.82.1

t0yv0:
Since an automated Pulumi update failed, I am attempting to append commits to make it work. Changes:

-  remove `cp scripts/install-pulumi-plugin.js bin/scripts` from the Makefile (accommodate breaking change in v3.80.0)
- update examples/go.mod
- switch Go SDK generation from generating `go 1.17` to generating `go 1.18` - required for tests to pass it appears (not fully explainable yet)


thomas11 (commented):
Latest is v3.84 by now, can we jump there directly while a PR is open?

t0yv0:
I'd prefer not to, since an automated job does that normally when nothing is breaking. If we land this PR I think the automated job will catch up. I can trigger it manually right after this PR so as not to have a delay.

thomas11 (approved):

Issue #959, opened 2021-06-30
Pulumi update issues with AKS clusters


michaelbpalmer:
We keep hitting issues with AKS and Pulumi. Namely, when we try to update properties on existing AKS clusters (managed by Pulumi), we often get into error cases where we end up having to manually edit our stack. Here are some examples:

1. We are unable to update the vm size. pulumi up fails and we have to manually delete the old AKS cluster, edit the stack and then run pulumi up to create the new cluster.
2. Updating maCount on spot instance agent pools

## Steps to reproduce

Issue 1:

1. Create an AKS cluster in Pulumi with any VM size.
2. Change the VM Size and run pulumi up.

Expected: VM size updates (even if cluster needs to be replaced).
Actual: Error and update is not performed. Only way to get out of this is to manually delete the cluster, edit the stack and rerun pulumi up.

Issue 2:

1. Create an AKS cluster with a spot instance AgentPool. Set the maxCount property to 1.
2. Update the maxCount property to 2 and run pulumi up.

Expected: pulumi up to succeed with maxCount updated.
Actual: pulumi up fails

`azure-native:containerservice:AgentPool (dg-gpuspot-agentpool-cpu-staging-dev):
error: Code="PropertyChangeNotAllowed" Message="Changing property 'properties.nodeTaints' is not allowed." Target="properties.nodeTaints"`

We weren’t trying to update the node taints. It seems like Pulumi knows it was a change to maxCount:

`~  azure-native:containerservice:AgentPool dg-gpuspot-agentpool-cpu-staging-dev updating [diff: ~maxCount]; error: Code="PropertyChangeNotAllowed" Message="Changing property 'properties.nodeTaints' is not allowed." Target="properties.nodeTaints"
 ~  azure-native:containerservice:AgentPool dg-gpuspot-agentpool-cpu-staging-dev **updating failed** [diff: ~maxCount]; error: Code="PropertyChangeNotAllowed" Message="Changing property 'properties.nodeTaints' is not allowed." Target="properties.nodeTaints"`

Taking a look at the stack output for this resource, I think the issue is that Pulumi is trying to remove the taints that Azure automatically adds to spot node pools. In this case we haven’t been able to get this to work even after editing the stack.

`{
                "urn": "urn:pulumi:spark-staging::spark-deployment::azure-native:containerservice:AgentPool::dg-gpuspot-agentpool-cpu-staging-dev",
                "custom": true,
[...]

PR #2774, opened 2023-09-26
Disable AKS cluster recreation on agent profile change

danielrbradley:
Quick fix for https://github.com/pulumi/pulumi-azure-native/issues/1808

Remove all manually curated agentPoolProfile properties from the forceNewMap as we're currently also forcing recreation when adding or removing agent pools. This method of replacing agent pools is documented here: https://learn.microsoft.com/en-us/azure/aks/use-system-pools?tabs=azure-cli#system-and-user-node-pools

If the current behaviour is still wanted (to replace the whole cluster instead of updating the agent pools in place) then the replaceOnChanges resource option can be used: https://www.pulumi.com/docs/concepts/options/replaceonchanges/

This was originally introduced by https://github.com/pulumi/pulumi-azure-native/pull/1210 to fix https://github.com/pulumi/pulumi-azure-native/issues/959 but it didn't take into account rotating agent pools within the existing cluster.

### DIFF
--- a/provider/pkg/gen/replacement.go
+++ b/provider/pkg/gen/replacement.go
@@ -19,25 +19,6 @@ var forceNewMap = map[string]map[string]codegen.StringSet{
 	},
 	"ContainerService": {
 		"ManagedCluster": codegen.NewStringSet(
-			// agentPoolProfiles
-			"availabilityZones",
-			"enableEncryptionAtHost",
-			"enableFIPS",
-			"enableNodePublicIP",
-			"kubeletConfig",
-			"linuxOSConfig",
-			"maxPods",
-			"name",
-			"nodeLabels",
-			"nodePublicIPPrefixID",
-			"nodeTaints",
-			"osDiskSizeGB",
-			"osSKU",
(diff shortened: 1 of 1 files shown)


thomas11 (commented):
Does this PR not resurrect #959, then? For instance, `vmSize` is one of the properties mentioned there and it's removed here.

danielrbradley:
@thomas11 yes, partially, because the premise of that issue was incorrect. You are not allowed to update those properties within an existing agent pool, but you can create a new agent pool within the AKS cluster with the new setting, then delete the old agent pool. This is a much better solution in production workloads than re-creating the whole cluster.

The issue with setting these sub-properties to force re-creation is that there's no way of opting out of the behaviour. However there are options to force creation, so leaving them as not forcing creation is the only way to make both modes of upgrades available to users.

thomas11 (approved):

mikhailshilkov (approved):
LGTM. It's reassuring that we remove our own overrides that turned out to be non bullet-proof.

Issue #2686, opened 2023-08-29
Remove `__inputs` in state


danielrbradley:
Follow on from:
- https://github.com/pulumi/pulumi-azure-native/issues/2552

In the subsequent major release of the provider we can:
1. Assert that the [`SendsOldInputs`](https://github.com/pulumi/pulumi/pull/13139) and [SendsOldInputsOnDelete](https://github.com/pulumi/pulumi/pull/14051) flags are set during configuration - failing with an error that the version of the engine must be upgraded.
2. Remove the writing of the `__inputs` output property.
3. Remove `__inputs` property parsing from `parseCheckpointObject` because old inputs is now guarenteed to be available on the gRPC request object.

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3552

PR #2775, opened 2023-09-26
Prefer using old inputs from the request, if available

### LINKED ISSUE #2552: Use original inputs for operations
Currently, the engine didn't provide us old inputs to diff and update. We worked around this by creating a special output property called `__inputs` which contained the inputs used so we could unpack them from the old outputs in the next deployment

Following https://github.com/pulumi/pulumi/pull/13139 being merged, we now get the actual old inputs passed to diff and update if using the new engine. This means we will be able to phase out the use of `__inputs`.

We do need to maintain backward compatibility with previous versions of the engine in case a stack is deployed with a newer version of the engine, then a subsequent deployment is using an older version of the engine. Therefore the migration to using old inputs is in two steps.

## 1. Use old inputs if available

Within the provider, we should upgrade our pulumi internal dependencies and use the old inputs from the request if they are available. These will be nil or empty if using an older version of the engine. If they are nil or empty, we should fall back to continuing to use `__inputs` to match the existing behaviour.

When writing the outputs we should still always include the `__inputs` property to ensure backward compatibility.

This enables users who have upgraded to the latest version of the engine to make manual state edits to the inputs section, and the `__inputs` can be safely deleted or ignored during manual edits.

## 2. Drop `__inputs` in future release

- https://github.com/pulumi/pulumi-azure-native/issues/2686

danielrbradley:
Fixes #2552

When using version v3.74.0 or greater of the Pulumi CLI, the `inputs` field from the state file will be used for comparing against, rather than the encrypted `outputs.__inputs` field. The `inputs` can safely be edited and, until #2686 is completed, `outputs.__inputs` will be re-generated on the next operation.

This is required when needing to prune imported resource's inputs or edit properties which might be flagged for replacement due to a major version upgrade.

~Currently blocked on https://github.com/pulumi/pulumi/pull/14051~

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -2002,9 +2002,11 @@ func checkpointObject(inputs resource.PropertyMap, outputs map[string]interface{
 	return object
 }
 
-// parseCheckpointObject returns inputs that are saved in the `__inputs` field of the state.
+// parseCheckpointObject returns the old inputs either from the RPC request or that are saved in the `__inputs` field of the state.
 func parseCheckpointObject(reqOldInputs *structpb.Struct, obj resource.PropertyMap, label string) (resource.PropertyMap, error) {
-	// Prefer the old inputs from the RPC request if available.
+	// Prefer the old inputs from the RPC request if available because these are much easier for a user to edit in the state file.
+	// In the next major version of the provider, we'll remove the `__inputs` field from the state and require that the user is on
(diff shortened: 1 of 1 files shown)


Frassle:
You'll want to wait for https://github.com/pulumi/pulumi/pull/14051 to release and then in Configure only turn on this new behavior if the engine sends true for `SendsOldInputsToDelete`.

thomas11 (approved):
I'm not sure I understand all the nuances behind this change TBH but it looks fine from what I do know.

PR #2786, opened 2023-09-29
Run example tests from provider module

danielrbradley:
Fixes https://github.com/pulumi/pulumi-azure-native/issues/2788

Author example tests within the provider module so the provider and unit tests are executing in the same process.

1. Start the provider host from the test setup.
2. Grab the host port and pass to the ProgramTest PULUMI_DEBUG_PROVIDERS env.
3. Use a deferred, cancellable context to shut down the provider at the end of the test.

Benefits:
- Diagnose provider behaviour by easily starting program tests in debug mode within an IDE and breakpoint straight through to provider calls.
- Generates test-coverage data on which parts of the provider are or aren't tested via program tests.
- Shorten testing feedback loop by removing requirement to build SDKs and provider binary before re-running tests.

Specifics:
- Implicitly label non-e2e tests via build tags as `unit`.
   - Local runs via `make test_provider` exclude integration tests by default for speed.
   - CI runs all tests by default
   - All tests are runnable within a development environment

Aside: remove unused schema string in provider setup.

## Long Term (inspirational) Plan

1. Move all program tests into provider module as YAML.
2. Build simple matrix testing to automatically convert YAML to each language and confirm correctness.
3. Use [HTTP session capture and replay](https://pkg.go.dev/github.com/google/go-replayers/httpreplay) to enable very fast testing of whole provider end-to-end without hitting the real backend for local development.
4. Run capture only on nightly job and run all PRs in replay mode by default.
5. Remove SDKs and leave only the single provider module to get rid of mono-repo layout (also fixing go.work oddities).


provider/pkg/arm2pulumi/converter_test.go
```diff
@@ -1,4 +1,5 @@
 // Copyright 2021, Pulumi Corporation.  All rights reserved.
+//go:build unit || all
```
t0yv0, line 2: Is this necessary? I am still struggling a bit with tags as they reduce out-of-the-box IDE usability. Is there a way to avoid the build tags? Not blocking on this but in the future?

provider/pkg/provider/provider_example_test.go
```diff
@@ -0,0 +1,130 @@
+// Copyright 2023, Pulumi Corporation.  All rights reserved.
+//go:build integration || all
```
t0yv0, line 2: Build tags here also.

t0yv0 (approved):
One of the most important PRs I've seen in months! Thank you.

provider/pkg/arm2pulumi/converter_test.go
danielrbradley: Yup - this approach is specifically to get around breaking the IDE usability. 

Here's how I got here...

Requirements:
1. Edit and run all tests from within the IDE without special settings.
2. Only run unit tests by default in the local `make test_provider`
3. Run unit and integration tests in a single execution in CI so we get accurate coverage numbers.

## Considered options

Option 1: Only set `integration` build flag

- ✅ Doesn't touch other tests
- ✅ Make skips integration tests by default
- ❌ IDE breaks

Option 2: Set `integration || all` build flag
- ✅ IDE works
- ❌ Can't target only unit tests from make

Option 3: Add build flags to categorise every unit test (plus `all` for all)
- ✅ IDE works (using `all`)
- ✅ Make only targets `unit`
- ✅ CI/CD targets all
- Technically the explicit `integration` flag isn't used, but it's there for consistency.

## Alternative 1: Environment variables

Add an environment variable `DISABLE_INTEGRATION_TESTS` which can be set to `true`. 
- ✅ IDE works with no special ENV
- ✅ make sets to `true` in make by default, but needs to allow overriding somehow. 
- ✅ CI/CD doesn't set the ENV

Downside: this feels like abuse of environment variables when there's a specific tool designed for this.

## Alternative 2: Negative build flags

I _think_ it's possible to use build flag exclusion runs to only exclude integration tests if `unit` is flagged, but I've not tested this.

1. Integration test would have `//go:build -unit`
2. IDE _should_ work because it's not targeting `unit`, but `all` .. which all unit tests have by default.
3. Make targets `all,unit` to exclude integration tests from the all.
4. CI/CD just targets `all`.

This would avoid having to remember to add `//go:build unit || all` but would require more explanation as to why it's like this 🙃

t0yv0: I'm curious how your IDE works on "integration||all" .. Emacs doesn't work unless I set env vars to set the Go tags. But if we cover VSCode that's probably enough, Emacs can take care of itself.

There's a few more options to distinguish tests.

1. Naming conventions. Running `go test -run '^TestUnit'` would filter down by regex.

2. There is a testing.Short() flag that can be checked inside a test and you skip the test if it's not what you want. You opt into that by `go test -short`. 

3. It is possible to add new ad-hoc flags. For example, https://github.com/hexops/autogold/blob/030d4676331d215ada737b5876e3a9b3947ee811/autogold.go#L38 uses `-update` flag to make tests rewrite their expected values (kind of like we do with PULUMI_ACCEPT env var sometimes).

t0yv0: I like negative build flags actually too if they work. What would really help me basically is that by default no tests is turned off, so if I want to run it, it runs, and I don't need to define any tags for it to run.

danielrbradley: The negative tags seems to work, so I've gone with that and will merge once green 👍 
Have added some docs along with it.

Notes on the other approaches:
- Using naming standards could work, but it's less explicit and might be missed by people adding new tests.
- We use `-short` for other purposes - separating long program tests from short program tests.
- We should look at the custom flag when we wrap this up into program test itself - probably overengineering for in just one provider.

thomas11 (approved):
This is awesome!

My only nits are naming-related:
- `getBaseOptions` isn't (only) what the function actually does anymore
- Is this the time to rename the confusing `examples` to something like `e2eTests` or `integrationTests`?

PR #2809, opened 2023-10-09
List all available API versions for a resource in the registry

thomas11:
For each resource, this PR adds all Azure API versions that are available in the SDK to the resource's description. The goal is to make these additional versions more easily discoverable for users.

I want to show a registry preview here but the last few attempts have failed due to what looks like an oom. In progress.

- Drive-by fix: wrong make target in message
- Format the versions more nicely.
- Refactor to make provider available during package gen
- Show only available, not all compatible other API versions.
- Regenerate SDKs


provider/pkg/gen/schema.go
```diff
@@ -946,49 +947,59 @@ func (g *packageGenerator) genPostFunctions(typeName, path string, pathItem *spe
...
 			}
-			sdkVersions[i] = sdkVersion
+			tok := g.generateTok(typeName, openapi.ApiToSdkVersion(v))
+			if g.shouldInclude(typeName, tok, v) {
+				includedVersions = append(includedVersions, v)
+			}
+		}
+
+		if len(includedVersions) > 0 {
+			fmt.Fprintf(&b, "\n<br/>Other available API versions: %s.", strings.Join(includedVersions, ", "))
```
danielrbradley, line 1001: Nit: Can we avoid using HTML formatting as this doesn't appear nicely in the SDKs - dotnet actually double encodes this to `&lt;br$gt;` which looks odd. Could we just add two new lines for spacing?

danielrbradley (approved):
Overall looks like a good result, only 1 question on the formatting.

provider/pkg/gen/schema.go
iwahbe: I'm not 100% sure for the specific case. In general, we emit registry docs in markdown (which is a superset of HTML). Markdown has various rules for which whitespace is ignored vs counts as a paragraph delineation. To get paragraph behavior, I believe that you need two newlines: `fmt.Fprintf(&b, "\n\nOther av...")`.

thomas11: Two newlines work, thank you! I didn't realize we're emitting Markdown here rather than plain text.

PR #2810, opened 2023-10-09
Debugging hack: env variable to capture the intermediate .pp of certain examples

thomas11:
Capture the generated intermediate representation of code samples from a given file. Useful to debug codegen issues and create tests for p/p.

### DIFF
--- a/provider/pkg/gen/examples.go
+++ b/provider/pkg/gen/examples.go
@@ -315,6 +315,10 @@ func generateExamplePrograms(example resources.AzureAPIExample, body *model.Body
 	return languageExample, nil
 }
 
+// writeDebugHCL writes the given example program to DEBUG_CODEGEN_EXAMPLE_HCL/ as a .pp file if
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):
I think this could be helpful for determining why our examples aren’t stable too - narrow it down to if it’s our gen code or the language-specific generation in pu/pu.

It's not clear what we need to add in the value of `DEBUG_CODEGEN_EXAMPLE_HCL`. It seems to be comparing against path.Match, but then match isn't used. Perhaps a comment in `writeDebugHCL` with an example of what to set it to would help?

thomas11:
> I think this could be helpful for determining why our examples aren’t stable too - narrow it down to if it’s our gen code or the language-specific generation in pu/pu.
> 
> It's not clear what we need to add in the value of `DEBUG_CODEGEN_EXAMPLE_HCL`. It seems to be comparing against path.Match, but then match isn't used. Perhaps a comment in `writeDebugHCL` with an example of what to set it to would help?

Added docs. `match` is used, it's just a bool.

Issue #2785, opened 2023-09-29
Missing resource FirewallRule, Database etc. for Postgres Single Server in version 2.*


quantum-byte:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Starting in version 2.* of pulumi-azure-native  the default resources for Postgres became the flexible server variants (See [here](https://www.pulumi.com/registry/packages/azure-native/from-v1-to-v2/#mysql-and-postgresql-server-and-flexible-server) and [here](https://github.com/pulumi/pulumi-azure-native/issues/2206)).

One can still create/manage a new Postgres Single Server with version 2.* by using 
```python
from pulumi_azure_native.dbforpostgresql import v20171201 as dbforpostgresql_single_server

dbforpostgresql_single_server.Server(
    *** 
)
```

But there do not exist resources like `FirewallRule` or `Database` in `pulumi_azure_native.dbforpostgresql.v20171201`.

Using the flexible server variants like `pulumi_azure_native.dbforpostgresql.FirewallRule` also does not work, because there is no api endpoints for it in newer versions. This causes errors during `pulumi refresh` like this:

```
  azure-native:dbforpostgresql:FirewallRule (db_access_via_aks):
    error: Preview failed: autorest/azure: Service returned an error. Status=400 Code="NoRegisteredProviderFound" Message="No registered resource provider found for location 'germanywestcentral' and API version '2022-12-01' for type 'servers'. The supported api-versions are '2017-12-01-preview, 2017-12-01'. The supported locations are 'australiacentral, australiacentral2, australiaeast, australiasoutheast, brazilsouth, canadacentral, canadaeast, centralindia, centralus, eastasia, eastus2, eastus, francecentral, francesouth, germanywestcentral, japaneast, japanwest, jioindiawest, koreacentral, koreasouth, northcentralus, northeurope, southafricanorth, southafricawest, southcentralus, southindia, southeastasia, norwayeast, swedencentral, switzerlandnorth, uaenorth, uksouth, ukwest, westcentralus, westeurope, westindia, westus, westus2, westus3, germanynorth'."
```

This makes migration from the deprecated Single Server Variant to the Flexible Server Variant quite difficult, since one can not have them both in a pulumi project with `pulumi-azure-native` 2.* version at the same time.

mikhailshilkov:
[...]

Issue #2813, opened 2023-10-11
Do not squeeze API versions of resources when the path differs


thomas11:
This issue is prompted by #2785 where the fix involved manually restoring resources. This manual edit will be overwritten during the next major version generation. Therefore, we need a permanent way to preserve these resources. There are two basic options:
1. Add the capability to preserve certain resources to the version curation.
2. Automatically detect that the resources should be preserved, in this case by seeing that the path in the version in question is different from the path in the default version.

Option (2) is preferable because it's automatic.

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3552

danielrbradley:
@thomas11 can we close this now we don't have resources with different paths?

PR #2814, opened 2023-10-11
Restore missing MySQL and PostgreSQL single-server resources in API version 20171201

thomas11:
This PR is simply a manual edit of `versions/v2-removed-resources.json`. For v2 of this provider, this will work because we won't squeeze the schema again.

For v3 and beyond, we need to find a more permanent solution which is tracked by #2813.


thomas11:
resolves #2785

danielrbradley (approved):

Issue #579, opened 2020-12-11
Can't add Agent Pool to AKS declaratively


kralikba:
Hello,

When the set of agent pools in ManagedCluster.AgentPoolProfiles is changed, the provider sends an incorrect request to the Azure API instead a set of requests:

```
azure-nextgen:containerservice/latest:ManagedCluster (k8s-cluster):
    error: Code="BadRequest" Message="A new agent pool was introduced. Adding agent pools to an existing cluster is not allowed through managed cluster operations. For agent pool specific change, please use per agent pool operations: https://aka.ms/agent-pool-rest-api" Target="agentPoolProfiles"
```

mikhailshilkov:
@kralikba Apologies for the no response, the issue slipped out. Do you still have this problem? If so, could you share your code and repro steps?

In general, any update to a resource results in a single PUT request, that's how the provider works (similarly to ARM templates).

kralikba:
Hello,
We have decided not to use AKS so I can't report on this. Maybe this behaviour cannot be correctly described using the schemas used to generate the provider?

mikhailshilkov:
Yes, this is the idea.

As it's not a problem for you anymore, I'll close the issue for now. If anyone is facing a similar problem and has a repro, please comment below.

Bj3MaS:
```` 
cluster:
`const managedCluster = new azure_native.containerservice.ManagedCluster("managedCluster", {
	resourceGroupName: resourceGroup.name,
	resourceName: `resourceName`,
	addonProfiles: {},
	agentPoolProfiles: [{
		count: 3,
		maxPods: 110,
		mode: "System",
		name: `someName`, <-- changing this
		osDiskSizeGB:30,
		osType:"Linux",
		type:"VirtualMachineScaleSets",
		vmSize:"Standard_D4_v3",
		vnetSubnetID: subnet.id
	}],
	dnsPrefix: "AzureNativeprovider",
	enableRBAC: true,
	kubernetesVersion: "1.19.7",
	servicePrincipalProfile: {
		clientId: application.applicationId,
		secret: servicePrincipalPassword.value
	},
	networkProfile: {
		networkPlugin: "azure",
		serviceCidr: "10.10.0.0/16",
		dnsServiceIP: "10.10.0.10",
		dockerBridgeCidr: "172.17.0.1/16"
	}
},
{
	dependsOn: subnetAssignment,
	deleteBeforeReplace: true,
	parent: this
});`
````

[...]

PR #2828, opened 2023-10-13
Add extra custom docs to AKS managed cluster

danielrbradley:
- Add folder with additional custom resource docs.
- Add some guidance on how to manage AKS inline agent pools.

Related to:
- https://github.com/pulumi/pulumi-azure-native/issues/579


thomas11 (approved):
Neat approach to adding docs!

PR #2842, opened 2023-10-23
Upgrade

### LINKED ISSUE #2839: Workflow failure: nightly-sdk-generation
## Workflow Failure

- [2023-10-20T04:45:14.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/6583446493)

- [2023-10-23T04:46:13.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/6608784494)

- [2023-10-24T04:45:15.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/6622412318)

danielrbradley:
Fixes #2839 

The `Migrate.VmUptimeArgs` type is incompatible between 2019-10-01 and 2023-03-15 as the property `HoursPerDay` changed from a number to an integer, which is a breaking change. To avoid this, we exclude pulling the new `AssessmentOperation` resource into the default version.

To investigate the issue, I edited the flow so we still return and output the calculated versions on a failure so we can see what new resources are being added at what version.


versions/v2-spec.yaml
```diff
@@ -1307,7 +1346,6 @@ Security:
         DeviceSecurityGroup: "2019-08-01"
         GovernanceAssignment: 2022-01-01-preview
         GovernanceRule: 2022-01-01-preview
-        IngestionSetting: 2021-01-15-preview
```
thomas11: Where did this resource go? Removed upstream?

danielrbradley: Yup, 2021-01-15-preview of Security was deleted

danielrbradley: https://github.com/Azure/azure-rest-api-specs/pull/25970

thomas11 (approved):

PR #2867, opened 2023-11-06
Bump schema-tools version

thomas11:
Missing out on a whole bunch of releases, e.g., with function signature checks.

@danielrbradley do we even need to pin this, since it's our tool for internal use? We'd get updates for free.


iwahbe (approved):
LGTM when tests are passing.

danielrbradley (approved):
Not pinning has bitten us in the past with weird failures so would prefer to focus on automating these kinds of bumps.

Issue #2864, opened 2023-11-04
2.15.0 failed to release cleanly


mikhailshilkov:
arm2pulumi step failed which also prevented the docs update: https://github.com/pulumi/pulumi-azure-native/actions/runs/6748088952/job/18346962980#step:4:63

Probably broken by https://github.com/pulumi/pulumi-azure-native/pull/2856

PR #2871, opened 2023-11-07
Remove arm2pulumi

danielrbradley:
The realse job is broken because goreleaser tries to call a makefile target that doesn't exist.

We don't need to release this any more because we have the new ARM converter template instead.

The web hosted tool also has been unable to be updated in a long time and should be replaced entirely with the ARM converter plugin.

We should also update https://github.com/pulumi/arm2pulumi to point to the new tool: https://github.com/Zaid-Ajaj/pulumi-converter-arm

Fixes https://github.com/pulumi/pulumi-azure-native/issues/2864


mikhailshilkov (approved):
LGTM

PR #2915, opened 2023-11-29
Allow removing network rule sets from storage accounts

### LINKED ISSUE #2507: Props removed from the code are not removed from the real resource
### What happened?

Customer reports there are several Azure Native resource types that do not conform our Pulumi declarative model: when an input property is completely removed from the code, and `pulumi up` is ran, Pulumi reports it succesfully applied the change, but nothing changed on the real Azure resource.

### Expected Behavior

When the input property is removed, the reported change should really be applied.

### Steps to reproduce

* Create a new project with `pulumi new azure-typescript`
* Update the template code to this:

```ts
import * as pulumi from "@pulumi/pulumi";
import * as storage from "@pulumi/azure-native/storage";

// Create an Azure Resource Group
const resourceGroup = new resources.ResourceGroup("<customer>-bug-delete-not-applied");

// Create an Azure resource (Storage Account)
const storageAccount = new storage.StorageAccount("sa", {
    resourceGroupName: resourceGroup.name,
    sku: {
        name: storage.SkuName.Standard_LRS,
    },
    kind: storage.Kind.StorageV2,
    networkRuleSet: {
        defaultAction: "Deny",
        ipRules: [
            {
                action: "Allow",
                iPAddressOrRange: "78.23.36.13"
            }
        ]
    }
});
```

* Run `pulumi up`:

```
pulumi up
Previewing update (team-ce/dev)

View in Browser (Ctrl+O): https://app.pulumi.com/team-ce/delete-not-applied/dev/previews/cd2d49ee-8937-4e84-b258-7e65c3d48c75

     Type                                     Name                        Plan       
 +   pulumi:pulumi:Stack                      delete-not-applied-dev      create     
 +   ├─ azure-native:resources:ResourceGroup <customer>-bug-delete-not-applied  create     
 +   └─ azure-native:storage:StorageAccount   sa                          create     


Resources:
    + 3 to create

Do you want to perform this update? details
+ pulumi:pulumi:Stack: (create)
    [urn=urn:pulumi:dev::delete-not-applied::pulumi:pulumi:Stack::delete-not-applied-dev]
    + azure-native:resources:ResourceGroup: (create)
        [urn=urn:pulumi:dev::delete-not-applied::azure-native:resources:ResourceGroup::<customer>-bug-delete-not-applied]
        [provider=urn:pulumi:dev::delete-not-applied::pulumi:providers:azure-native::default_1_103_0::04da6b54-80e4-46f7-96ec-b56ff0331ba9]
        location            : "WestEurope"
        resourceGroupName   : "<customer>-bug-delete-not-applied7f1745d2"
    + azure-native:storage:StorageAccount: (create)
        [urn=urn:pulumi:dev::delete-not-applied::azure-native:storage:StorageAccount::sa]
        [provider=urn:pulumi:dev::delete-not-applied::pulumi:providers:azure-native::default_1_103_0::04da6b54-80e4-46f7-96ec-b56ff0331ba9]
        accountName         : "sabe751de1"
        kind                : "StorageV2"
        location            : "WestEurope"
        networkRuleSet      : {
            bypass       : "AzureServices"
            defaultAction: "Deny"
            ipRules      : [
                [0]: {
                    action          : "Allow"
                    iPAddressOrRange: "78.23.36.13"
                }
            ]
        }
        resourceGroupName   : output<string>
        sku                 : {
            name: "Standard_LRS"
        }

Do you want to perform this update? yes
Updating (team-ce/dev)

View in Browser (Ctrl+O): https://app.pulumi.com/team-ce/delete-not-applied/dev/updates/1

     Type                                     Name                        Status              
 +   pulumi:pulumi:Stack                      delete-not-applied-dev      created (1s)        
 +   ├─ azure-native:resources:ResourceGroup <customer>-bug-delete-not-applied  created (0.83s)     
 +   └─ azure-native:storage:StorageAccount   sa                          created (21s)       


Resources:
    + 3 created

Duration: 27s

```
* The storage account is created and there is a firewall rule configured with the given IP address:

Screenshot 2023-06-09 at 12 04 27

* Now remove the complete `networkRuleSet` input property from the `StorageAccount` resource.
* Run `pulumi up` again:

```
$ pulumi up
Previewing update (team-ce/dev)

View in Browser (Ctrl+O): https://app.pulumi.com/team-ce/delete-not-applied/dev/previews/e0c54801-58aa-4262-a645-d2acd2ca62f3

     Type                                    Name                    Plan       Info
     pulumi:pulumi:Stack                     delete-not-applied-dev             
 ~   └─ azure-native:storage:StorageAccount  sa                      update     [diff: -networkRuleSet]


Resources:
    ~ 1 to update
    2 unchanged

Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::delete-not-applied::pulumi:pulumi:Stack::delete-not-applied-dev]
    ~ azure-native:storage:StorageAccount: (update)
        [id=/subscriptions/32b9cb2e-69be-4040-80a6-02cd6b2cc5ec/resourceGroups/<customer>-bug-delete-not-applieda2855651/providers/Microsoft.Storage/storageAccounts/saeaa0e26a]
        [urn=urn:pulumi:dev::delete-not-applied::azure-native:storage:StorageAccount::sa]
        [provider=urn:pulumi:dev::delete-not-applied::pulumi:providers:azure-native::default_1_103_0::f1b4a4f5-dda9-45a2-953e-fd106ddfcf05]
      - networkRuleSet: {
          - bypass       : "AzureServices"
          - defaultAction: "Deny"
          - ipRules      : [
          -     [0]: {
                  - action          : "Allow"
                  - iPAddressOrRange: "78.23.36.13"
                }
            ]
        }

Do you want to perform this update? yes
Updating (team-ce/dev)

View in Browser (Ctrl+O): https://app.pulumi.com/team-ce/delete-not-applied/dev/updates/2

     Type                                    Name                    Status              Info
     pulumi:pulumi:Stack                     delete-not-applied-dev                      
 ~   └─ azure-native:storage:StorageAccount  sa                      updated (0.88s)     [diff: -networkRuleSet]


Resources:
    ~ 1 updated
    2 unchanged

Duration: 4s

```

* Expecation is now that the firewall rule should be removed from the `StorageAccount` resource, but it is still there.

### Output of `pulumi about`

CLI          
Version      3.70.0
Go Version   go1.20.4
Go Compiler  gc

Plugins
NAME          VERSION
azure-native  1.103.0
nodejs        unknown

Host     
OS       darwin
Version  13.3.1
Arch     arm64

This project is written in nodejs: executable='/Users/ringods/.volta/bin/node' version='v16.19.1'

Current Stack: team-ce/delete-not-applied/dev

TYPE                                  URN
pulumi:pulumi:Stack                   urn:pulumi:dev::delete-not-applied::pulumi:pulumi:Stack::delete-not-applied-dev
pulumi:providers:azure-native         urn:pulumi:dev::delete-not-applied::pulumi:providers:azure-native::default_1_103_0
azure-native:resources:ResourceGroup  urn:pulumi:dev::delete-not-applied::azure-native:resources:ResourceGroup::<customer>-bug-delete-not-applied
azure-native:storage:StorageAccount   urn:pulumi:dev::delete-not-applied::azure-native:storage:StorageAccount::sa

Found no pending operations associated with team-ce/dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/ringods
User           ringods
Organizations  ringods, team-ce, pulumiverse, demo

Dependencies:
NAME                  VERSION
@pulumi/azure-native  1.103.0
@pulumi/pulumi        3.70.0
@types/node           16.18.34

### Additional context

_No response_

### Contributing

thomas11:
Fixes #2507 

In this issue, the user removed a property from their resource definition and expected it to get removed or reset in Azure as well. Azure, however, interpreted the omission of the property on the update request as "no changes to make here". 

This leads to out-of-sync Pulumi and Azure state. Refreshing doesn't work to fix this because when the updated state comes back, the removed property is still there, so then the value gets maintained across a refresh because it treats it like a default value.

The spec does not describe this behavior, so I added a custom exception: on update of a storage account, if the property was set previously but now is not, explicitly set it to the default. I copied that default from an Azure response.

The added test is essentially the user's repro from #2507. After the second update where the network rules were removed, we assert that they are back to the default value in outputs.

So far, this PR is not generalized at all. We could certainly do that, hiding the details and making it easily re-usable for other APIs that have the same problem, but I figured that's premature optimization and we can do that when we encounter more cases.

### DIFF
--- a/provider/pkg/resources/resources.go
+++ b/provider/pkg/resources/resources.go
@@ -108,7 +108,7 @@ type AzureAPIResource struct {
 	// Containers within the request body that are required even if no properties are set within it.
 	RequiredContainers [][]string `json:"requiredContainers,omitempty"`
 	// Default values to be used when the property is removed or in importing. Must be top-level properties.
-	DefaultProperties map[string]interface{}
+	DefaultProperties map[string]interface{} `json:"defaultProperties,omitempty"`
 }
 
 func (res *AzureAPIResource) LookupProperty(key string) (AzureAPIProperty, bool) {


provider/pkg/provider/provider.go
```diff
@@ -1195,6 +1199,32 @@ func (k *azureNativeProvider) Update(ctx context.Context, req *rpc.UpdateRequest
 		}, nil
 	}
 
+	// Special case for storage accounts: if networkRuleSet was specified and then removed, set it
+	// back to the default value because Azure will interpret omission as "no changes".
+	// https://github.com/pulumi/pulumi-azure-native/issues/2507
+	if !req.GetPreview() && resourceKey == "azure-native:storage:StorageAccount" {
```
danielrbradley, line 1205: Why do we exclude previews here? Would this not result in different values being shown at preview from what it will actually do?

provider/pkg/provider/provider.go
```diff
@@ -1195,6 +1199,32 @@ func (k *azureNativeProvider) Update(ctx context.Context, req *rpc.UpdateRequest
...
+		oldState, err := readOlds()
+		if err != nil {
+			return nil, err
+		}
+
+		networkRuleSet := resource.PropertyKey("networkRuleSet")
+		if !inputs.HasValue(networkRuleSet) && oldState.HasValue(networkRuleSet) {
+			const defaultNetworkRuleSetJson = `{
```
danielrbradley, line 1213: Why do we define this as JSON which we then have to unmarshal? Could we just write this directly as a map or is it more messy that way?

provider/sa-payload.json
```diff
@@ -0,0 +1,59 @@
+{
```
danielrbradley, line 1: I can't see this file used anywhere - was it committed by mistake?

provider/pkg/provider/provider.go
danielrbradley, line 1205: Does this only apply to the default resource? Could we extend it to also match any of the explicitly versioned resources too?

danielrbradley (commented):
A couple of questions inline - overall approach looks sound though.

It seems likely that this problem also affects other resources. How tricky would it be to extract this hard-coded switch to something that could be pulled from the metadata? 

I'm also wondering about how this issue overlaps with the problem of not knowing default values during import. Could we extend this approach to allow us to start curating more of the default values so that we could have a route to reducing the number of inputs created during an import?

What I thinking is that adding support for complex default values in the metadata could solve this specific issue but also allow us to utilize more of the default values from the specification (which aren't supported in schema) and think about a mechanism for curating more manual defaults similar to this one. This doesn't have to block getting this PR in, but we could just capture it for future work.

provider/pkg/provider/provider.go
thomas11: That was an oversight from when I had initially just copy-pasted the Azure response.

provider/pkg/provider/provider.go
thomas11: Done, although the code's rather ugly. I couldn't find any existing helpers to parse resource token strings, only to construct them.

I was maybe a little paranoid with not overmatching here. Maybe `keyParts[2] == "StorageAccount" && strings.HasPrefix(keyParts[1], "storage")` would suffice?

provider/pkg/provider/provider.go
```diff
@@ -1265,29 +1283,24 @@ func (k *azureNativeProvider) Update(ctx context.Context, req *rpc.UpdateRequest
...
 
-// Add or modify certain inputs of certain resourcesto make updates conform to the Pulumi resource model.
-func adjustInputsForSpecialCases(inputs resource.PropertyMap, resourceKey string, readOlds func() (resource.PropertyMap, error)) error {
-	// Special case for storage accounts: if networkRuleSet was specified and then removed, set it
-	// back to the default value because Azure will interpret omission as "no changes".
-	// https://github.com/pulumi/pulumi-azure-native/issues/2507
-	keyParts := strings.Split(resourceKey, ":")
-	if len(keyParts) == 3 && keyParts[2] == "StorageAccount" && (keyParts[1] == "storage" || strings.HasPrefix(keyParts[1], "storage/")) {
-		oldState, err := readOlds()
+func restoreDefaultInputsForRemovedProperties(inputs resource.PropertyMap, res resources.AzureAPIResource, readOlds func() (resource.PropertyMap, error)) error {
```
danielrbradley, line 1286: This looks like a good candidate for a few unit tests.

provider/pkg/provider/provider.go
```diff
@@ -1094,6 +1099,19 @@ func (k *azureNativeProvider) Read(ctx context.Context, req *rpc.ReadRequest) (*
 	return &rpc.ReadResponse{Id: id, Properties: checkpoint, Inputs: inputsRecord}, nil
 }
 
+// Converts oldState into a serializable object map, with the resource's default values from metadata removed.
+func mappableOldState(res resources.AzureAPIResource, oldState resource.PropertyMap) map[string]interface{} {
```
danielrbradley, line 1103: This looks fairly simple, but perhaps a test would be helpful to demonstrate intended behaviour.

danielrbradley (approved):
This looks good with the defaults in the schema. Might be nice to have a couple of targeted unit tests for the simple functions to test.

PR #2922, opened 2023-12-01
Add selected GET invokes

### LINKED ISSUE #2419: Bring azure.monitoring.getDiagnosticCategories to azure-native
## Hello!

## Issue details

Regarding the `diagnostic settings` in [azure-native](https://www.pulumi.com/registry/packages/azure-native/api-docs/insights/diagnosticsetting). I wanted to create a resource of type [azure-native.insights.DiagnosticSetting](https://www.pulumi.com/registry/packages/azure-native/api-docs/insights/diagnosticsetting) when doing so I needed to provide the fields `logs` and `metrics`, they require a [category](https://www.pulumi.com/registry/packages/azure-native/api-docs/insights/diagnosticsetting/#category_python
) that needs to be retrieved for each different resource type. 

My problem is that there is no way to get the diagnostic categories of a resource with azure-native. The doc specifies that: 
> To obtain the list of Diagnostic Log categories for a resource, first perform a GET diagnostic settings operation.

But no such method exists in azure-native, I eventually used the azure-classic method: [azure.monitoring.getDiagnosticCategories
](https://www.pulumi.com/registry/packages/azure/api-docs/monitoring/getdiagnosticcategories)

Would it be possible to include such a method in azure-native too?

Thanks for your time 😃

thomas11:
We only create `getFoo` or `listFoo` functions in combination with a `Foo` resource. We don't turn all GET endpoints to functions because most aren't very useful. 

In some cases, like #2419 to start with, we do however miss a GET function that's users need. This PR adds the one from that issue and lays the groundwork for adding more in a well-defined place.

Reviewers, the core change is schema.go line 305 where we now also consider GET functions.

Resolves #2419


provider/pkg/gen/schema.go
```diff
@@ -917,9 +936,9 @@ func (g *packageGenerator) generateExampleReferences(resourceTok string, path *s
 	return nil
 }
 
-// genPostFunctions defines functions for list* (listKeys, listSecrets, etc.)
-// and get* (getFullUrl, getBastionShareableLink, etc.) POST endpoints.
-func (g *packageGenerator) genPostFunctions(typeName, path string, pathItem *spec.PathItem, swagger *openapi.Spec) {
```
danielrbradley, line 922: Nit: Having split the operation into its own argument, the only use of `pathItem` is `pathItem.Parameters` - so we could just pass `pathParameters` in the place of `pathItem`.

provider/pkg/gen/schema.go
```diff
@@ -375,6 +367,33 @@ version using infrastructure as code, which Pulumi then uses to drive the ARM AP
 	}, nil
 }
 
+func (g *packageGenerator) genPostInvokes(invokes map[string]*openapi.ResourceSpec) {
+	g.genInvokes(invokes, true)
+}
+
+func (g *packageGenerator) genGetInvokes(invokes map[string]*openapi.ResourceSpec) {
+	g.genInvokes(invokes, false)
```
danielrbradley, line 375: Nit: It would be nice to give context on what the `true` and `false` refer to. It's not terrible given the function being called is right below though. Perhaps just a `/* post */` type of inline comment could be enough here ... or we could use a string or enum for verbosity e.g. `g.genInvokes(invokes, "GET")`?

provider/pkg/openapi/discover.go
```diff
@@ -60,7 +69,7 @@ type DefaultVersionLock = map[ProviderName]map[DefinitionName]ApiVersion
 
 func (v VersionResources) All() map[string]*ResourceSpec {
 	specs := map[string]*ResourceSpec{}
-	for s, resourceSpec := range v.Invokes {
+	for s, resourceSpec := range v.POST_Invokes {
 		specs[s] = resourceSpec
 	}
 	for s, resourceSpec := range v.Resources {
```
danielrbradley, line 75: `All` should include the new GET_Invokes too. Can't remember where this is used, but it would be odd not to include it.

provider/pkg/openapi/discover.go
```diff
@@ -512,12 +519,50 @@ func (providers AzureProviders) addAPIPath(specsDir, fileLocation, path string,
...
 				Swagger:  swagger,
 			}
 		}
 	}
+
+	disambiguation := addGetFunctionIfRequired(version, pathItem, path, swagger)
+	if disambiguation != nil {
+		disambiguation.FileLocation = fileLocation
+		nameDisambiguations = append(nameDisambiguations, *disambiguation)
+	}
```
danielrbradley, line 534: Any way to de-duplicated code with above so it's in a method maybe?

danielrbradley (commented):
Overall approach looks okay to me. Some comments inline.

provider/pkg/openapi/discover.go
thomas11: Good call, there are actually many of these. Double-check my handling of the slice in the closure, but a local test looks good.

PR #2924, opened 2023-12-01
Add special case to fix `testbase` schema gen nightly failure

rquitales:
testbase has just added a new version of the resource `Package` which made the `targetOSs` field required. This is incompatible to with the existing version we're using so this special case handling removes this field as a required type.

Fixes: #2917


blampe (approved):
Seems consistent with the other exceptions.

PR #2949, opened 2023-12-16
Manually add invokes for Microsoft.MarketplaceOrdering sign and cancel

thomas11:


danielrbradley (changes requested):
Any modifications to cloud state need to be make via declarative resources, not invokes.

PR #2950, opened 2023-12-18
Allow maintaining nested sub-resources at arbitrary depth

thomas11:
This PR is a generalization of #2755, and groundwork for #594. I recommend reading their descriptions first.

The properties pointing to sub-resources that can be maintained as stand-alone resources can now be at any level, with `#/type` references being followed. Traversal methods have been added to `resources.AzureAPIProperty` that can gather the paths leading to sub-resource properties (and can also be used for other traversals).

In the provider, a new function `findUnsetSubResourceProperties` uses the collected sub-resource paths to look up these properties in API-shaped objects and collects the ones that are not set. In `Read()`, we remove them from outputs so they are not projected into inputs. In `Update()`, we use them to determine where we need to insert cloud state to preserve values not set in our inputs.

In order to be able to unit test code that looks up types, this PR also introduces an abstraction `resources.TypeLookupFunc` over the previous direct access to `azureNativeProvider.resourceMap.Types` which is hard to mock.

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -985,7 +985,10 @@ func (k *azureNativeProvider) setUnsetSubresourcePropertiesToDefaults(res resour
 		for _, pathEl := range p.path[:len(p.path)-1] {
 			curObj, ok := cur[pathEl]
 			if !ok {
-				break
+				newContainer := map[string]any{}
+				cur[pathEl] = newContainer
+				cur = newContainer
+				continue
 			}
 			cur, ok = curObj.(map[string]any)
 			if !ok {
@@ -993,7 +996,7 @@ func (k *azureNativeProvider) setUnsetSubresourcePropertiesToDefaults(res resour
 			}
 		}
 
-		cur[p.propertyName] = []any{}
+		cur[p.path[len(p.path)-1]] = []any{}
 	}
 }
 
--- a/provider/pkg/provider/provider_test.go
+++ b/provider/pkg/provider/provider_test.go
@@ -529,3 +529,43 @@ func setUpResourceWithRefAndProviderWithTypeLookup() (*resources.AzureAPIResourc
 
 	return &res, &provider
 }
+
+func TestSetUnsetSubresourcePropertiesToDefaults(t *testing.T) {
+	res, provider := setUpResourceWithRefAndProviderWithTypeLookup()
+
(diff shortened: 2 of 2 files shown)


provider/pkg/resources/resources.go
```diff
@@ -109,19 +109,79 @@ type AzureAPIResource struct {
...
+func (res *AzureAPIResource) LookupProperty(key string) (AzureAPIProperty, bool) {
+	if body := res.BodyParameter(); body != nil {
+		if prop, ok := body.Body.Properties[key]; ok {
+			return prop, true
+		}
 	}
 	return AzureAPIProperty{}, false
 }
 
+func (res *AzureAPIResource) BodyParameter() *AzureAPIParameter {
```
danielrbradley, line 176: Add bool `ok` to return.

provider/pkg/resources/resources.go
```diff
@@ -109,19 +109,79 @@ type AzureAPIResource struct {
...
+	for propName, prop := range props {
+		if prop.Ref != "" {
+			refType, ok, err := lookupType(prop.Ref)
+			if !ok || err != nil {
+				fmt.Printf("Cannot traverse properties of %s: failed to find ref %s: %v\n", propName, prop.Ref, err)
+				continue
+			}
+			if _, visited := seen[prop.Ref]; !visited {
+				seen[prop.Ref] = struct{}{}
+				traverseProperties(refType.Properties, lookupType, append(path, propName), seen, f)
```
danielrbradley, line 138: Include containers before `propName`.

provider/pkg/resources/resources.go
```diff
@@ -109,19 +109,79 @@ type AzureAPIResource struct {
...
+	if body == nil {
+		return
+	}
+
+	result := [][]string{}
+	TraverseProperties(
+		body.Body.Properties,
+		typeLookup,
+		func(propName string, prop AzureAPIProperty, path []string) {
+			if prop.MaintainSubResourceIfUnset {
```
danielrbradley, line 157: Make `MaintainSubResourceIfUnset` private.

danielrbradley (commented):
Minor comments inline. Update test snapshot to fix failing test.

danielrbradley (approved):
Looks good to go once we've resolved the failing test.

danielrbradley (approved):
Everything's green, all the discussed changes seem to be there and working. Let's ship it! :shipit:

PR #2951, opened 2023-12-18
Add a custom resource for Key Vault access policies

thomas11:
This PR implements #594, allowing access policies (APs) to be managed as individual Pulumi resources.

_This PR builds on top of #2950 and shares its commits._

It adds a new custom resource following the design of the existing ones: use Azure's KeyVault client to directly implement the AP CRUD operations. Note that access policies do not have standard CRUD semantics. Instead, the "add" operation merges if the AP already exists, and the "replace" operation creates if it does not exist. Therefore, the custom resource uses "replace" for both create and update, with a prior read to ensure that the AP does/does not exist yet.

In addition to the custom resource, some other changes were required:
- The KV client authenticates via the newer `azcore` interfaces. Our existing tokens are valid, so I added a wrapper to implement the azcore interface.
- When a resource has properties pointing to independent sub-resources, set them to their empty default value on create if not specified. This was required because `Vault` requires the `accessPolicies` property, but we don't want to explicitly set it to an empty `[]` and record this in Pulumi state. Otherwise, every update to the vault would clear out the APs.


danielrbradley (approved):
Core implementation and integration tests all look good to go 👍 

I think there's opportunity to add some basic unit tests around the simple funcitons like `parseKeyVaultPathParams`, `sdkPolicyToMap`, `sdkPermissionsToMap` and `propertyPermissionsToSdk` but happy for this to ship with the core integration tests for now if preferred.

PR #2961, opened 2023-12-19
Fix nightly upgrade due to incompatible refs

### LINKED ISSUE #2946: Workflow failure: nightly-sdk-generation
## Workflow Failure

- [2023-12-14T04:45:56.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/7204548136)

- [2023-12-14T22:08:07.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/7204548136)

- [2023-12-15T04:45:33.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/7218032766)

- [2023-12-18T04:45:59.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/7243694827)

- [2023-12-19T04:46:54.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/7257441428)

danielrbradley:
Closes #2946

## Improve how type merging fails on ref mismatch

Catch where we're trying to merge two different refs (which is currently impossible) and return a regular error rather than causing a panic by falling through to `type = "object"` even though we don't have AdditionalProperties.

- Fix deprecation warning by using Assertf.

## Extend HybridContainerService excludes

Also exclude the new 2024-01-01 version as it causes the same errors as 2023-11-15-preview:

```
panic:
failed to generate 'azure-native:hybridcontainerservice:ProvisionedClusterInstance':
response type:
incompatible type "azure-native:hybridcontainerservice:ControlPlaneProfileResponse" for resource "ProvisionedClusterInstance" ("azure-native:hybridcontainerservice:ProvisionedClusterInstance"):
refs do not match: #/types/azure-native:hybridcontainerservice:ControlPlaneProfileResponseControlPlaneEndpoint vs #/types/azure-native:hybridcontainerservice:ControlPlaneEndpointProfileResponseControlPlaneEndpoint
```

## Extend ServiceFabric excludes

Also exclude the new 2023-11-01-preview version as it causes the same errors as 2021-06-01:

```
panic:
failed to generate 'azure-native:servicefabric:ManagedCluster':
response type:
incompatible type "azure-native:servicefabric:ApplicationTypeVersionsCleanupPolicyResponse" for resource "ManagedCluster" ("azure-native:servicefabric:ManagedCluster"):
types not mergable: {integer  <nil> <nil> [] <nil> false} vs {number  <nil> <nil> [] <nil> false}
```

## Unify integer & number in type merging

Number (float) and integer can both be represented reasonably safely in our system as floats as their JSON representation is the same. (See above error)

## Exclude servicefabric (non-managed) 2023-11-01-preview

There's already servicefabricmanagedclusters 2023-11-01-preview.
servicefabric and servicefabricmanagedclusters both contain identically named types which are not compatible.
This causes a failure when building the explicit 2023-11-01-preview version of their resources.

This was caused by https://github.com/Azure/azure-rest-api-specs/pull/26671

### DIFF
--- a/provider/pkg/gen/merging.go
+++ b/provider/pkg/gen/merging.go
@@ -1,7 +1,6 @@
 package gen
 
 import (
-	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -138,7 +137,7 @@ func isPrimitiveType(t schema.TypeSpec) bool {
 
 // unionOneOfTypes checks that two oneOf types are allowed to be represented as a single schema type.
 func unionOneOfTypes(oneOf, other schema.TypeSpec) (*schema.TypeSpec, error) {
-	contract.Assert(oneOf.OneOf != nil)
+	contract.Assertf(oneOf.OneOf != nil, "Type %v is not a oneOf", oneOf)
 	if isPrimitiveType(other) {
 		// Check if type exists in oneOf.
 		for _, t := range oneOf.OneOf {
@@ -235,10 +234,10 @@ func unionOneOfTypes(oneOf, other schema.TypeSpec) (*schema.TypeSpec, error) {
 // 	}
 // 	return nil, errors.Errorf("cannot union oneOf with specified type %v", other)
 // }
-
-func oneOfTypeKey(t schema.TypeSpec) string {
-	// convert to json
-	b, err := json.Marshal(t)
-	contract.Assert(err == nil)
-	return string(b)
-}
+//
+// func oneOfTypeKey(t schema.TypeSpec) string {
+// 	// convert to json
+// 	b, err := json.Marshal(t)
+// 	contract.Assert(err == nil)
+// 	return string(b)
+// }
--- a/provider/pkg/gen/merging_test.go
+++ b/provider/pkg/gen/merging_test.go
@@ -34,6 +34,55 @@ func Test_mergeTypes(t *testing.T) {
 		wantErr bool
 	}{
 		{
+			name: "object and enum",
+			args: args{
+				t1: objectComplexType(map[string]schema.PropertySpec{
+					"foo": {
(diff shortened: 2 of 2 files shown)


thomas11 (approved):

provider/pkg/gen/merging.go
```diff
@@ -81,15 +81,23 @@ func mergeTypeSpec(t1, t2 schema.TypeSpec) (*schema.TypeSpec, error) {
...
-	// Simple case: both types are the same ref.
-	if t1.Ref == t2.Ref && t1.Ref != "" {
-		return &t1, nil
+	if t1.Ref != "" && t2.Ref != "" {
+		// Simple case: both types are the same ref.
+		if t1.Ref == t2.Ref {
+			return &t1, nil
+		}
+		// Refs don't match - fail with error.
+		return nil, errors.Errorf("refs do not match: %s vs %s", t1.Ref, t2.Ref)
```
mjeffryes, line 96: It would be really nice if the errors produced in merge linked to a short playbook for adding new exclusions as needed.

mjeffryes (approved):
Thanks!

provider/pkg/gen/merging.go
danielrbradley: Yeah, this would be nice. Unfortunately the solutions seems to be quite varied still. 

Options that come to mind right now are:
1. Excluding a version from the default version selection through exclusions in the config.
2. Modifying the merge logic to handle a new case.
3. Add an exception before calling merge to pick one type or the other.
4. Exclude an API version entirely from discovery.

mjeffryes: Even just describing the options would be a good first step; it would be great not to have these always fall to you or Thomas

PR #2962, opened 2023-12-19
Add a dummy playbook for resolving merge errors

mjeffryes:
Merge errors on a new API release seem to be somewhat common, so I'm starting a guide to enumerate our options for fixing them. Would love some help filling in the details, but even if we just have bullet points and an example that's something to help whoever is on rotation to investigate workflow failures


thomas11 (commented):
We already have playbooks, including for azure-native, at https://github.com/pulumi/platform-providers-team/tree/main/playbooks. I think it should go there rather than adding a new location.

mjeffryes:
> We already have playbooks, including for azure-native, at https://github.com/pulumi/platform-providers-team/tree/main/playbooks. I think it should go there rather than adding a new location.

I wanted to put it here because it's very specific to this provider, but open to moving it there if you disagree.

danielrbradley (approved):

PR #2777, opened 2023-09-27
Investigate applying forceNew on nested properties

danielrbradley:
In reference to the comment:

> Note: a non-updatable property at a subtype level (a property of a property of a resource) does not
> mandate the replacement of the whole resource. Anyway, it's used very seldom (2 places at the time of writing).

The report shows 722 current properties skipped. This is 207 unique properties (ignoring multiple versions of the same property).

This is also in reference to: https://github.com/pulumi/pulumi-azure-native/issues/2577 where we now need to utilize forceNew for a sub-property.

PR #2968, opened 2023-12-21
Diff keyed arrays

### LINKED ISSUE #2781: Make Diff aware of keyed arrays
## Hello!

## Issue details

Azure's specification [has a custom field called `x-ms-identifiers`](https://azure.github.io/autorest/extensions/#x-ms-identifiers) which can be used to specify a unique key for items in the array.

> - The default identifying property is id. You don’t need to add this annotataion if there is the id property in the object model.
> - If there is no appropriate identifying properties, leave it as an empty array.
> - You can provide multiple properties to form the composite identifier.
> - Use [json pointer](https://datatracker.ietf.org/doc/html/rfc6901) if the identifying property is an inner property. E.g., "x-ms-identifiers": ["/properties/subnet/id"]

Where this feature is used, the order of the array might not be maintained by Azure - the order returned by a GET request might not match the order as sent in our PUT request. 

## Impact

This leads to incorrect diffs as described in these issues where diff assumes all the items have been edited:

- https://github.com/pulumi/pulumi-azure-native/issues/1808#issuecomment-1255438944

As a workaround, the user might be able to re-order their declarations to match how Azure will order them when returned - if this ordering is known and stable.

We can't safely enable forcing resource recreation on sub-property changes due to these diffs appearing where there's actually no change and therefore forcing a re-creation on every deployment. This affects https://github.com/pulumi/pulumi-azure-native/issues/2577

## Implementation

1. When parsing the specification, we need to capture details about the identifiers for array into the metadata.
2. When performing the diff, we need to incorporate the identifier metadata when comparing the array's changes.

Currently, the diff is implemented in two steps:
1. Calling the generic `resource.PropertyMap.Diff`. 
2. Taking the output of the generic diff and attempt to modify it with Azure-specific details (identify forceNew properties and calculates replacements).

Therefore, to implement this key-aware diff we will almost certainly need to create our own Diff implementation which treats keyed arrays more like maps to avoid creating erroneous diffs in the first step. 

### Affected area/feature

- Diff
- Codegen/metadata

thomas11:
Azure's specification [has a custom field called x-ms-identifiers](https://azure.github.io/autorest/extensions/#x-ms-identifiers) which can be used to specify a unique key for items in the array. Where this feature is used, the order of the array might not be maintained by Azure - the order returned by a GET request might not match the order as sent in our PUT request. For details see #2781.

This PR changes diffing to handle these keyed arrays more like sets, comparing elements based on their ids.

It also contains @danielrbradley's PR #2777 to generate a report of these properties.

In order to be able to customize the array diffing, it was necessary to copy some pulumi/pulumi diffing code to this repository. Quoting #2781,
> the diff is implemented in two steps:
> 1. Calling the generic resource.PropertyMap.Diff.
> 2. Taking the output of the generic diff and attempt to modify it with Azure-specific details (identify forceNew properties and calculates replacements).

We needed to modify step 1 here.

Moving the diffing code here also allows us to integrate and simplify the Azure-specific post-processing of diffs of step 2. That's not part of this PR.

Resolves #2781

### DIFF
--- a/provider/pkg/provider/properties_diff_test.go
+++ b/provider/pkg/provider/properties_diff_test.go
@@ -1,0 +1,341 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package provider
+
+// NOTE: this file was copied from pulumi/pulumi/sdk/go/common/resource as part of #2968 which
+// required customizing the diffing logic.
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	. "github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+func assertDeepEqualsIffEmptyDiff(t *testing.T, val1, val2 PropertyValue) {
+	diff := val1.Diff(val2)
+	equals := val1.DeepEquals(val2)
+	assert.Equal(t, diff == nil, equals, "DeepEquals <--> empty diff")
+}
+
+func TestNullPropertyValueDiffs(t *testing.T) {
+	t.Parallel()
+	d1 := NewNullProperty().Diff(NewNullProperty())
+	assert.Nil(t, d1)
+	d2 := NewNullProperty().Diff(NewProperty(true))
+	assert.NotNil(t, d2)
+	assert.Nil(t, d2.Array)
+	assert.Nil(t, d2.Object)
+	assert.True(t, d2.Old.IsNull())
+	assert.True(t, d2.New.IsBool())
+	assert.Equal(t, true, d2.New.BoolValue())
+}
+
+func TestBoolPropertyValueDiffs(t *testing.T) {
+	t.Parallel()
+	d1 := NewProperty(true).Diff(NewProperty(true))
+	assert.Nil(t, d1)
+	d2 := NewProperty(true).Diff(NewProperty(false))
+	assert.NotNil(t, d2)
+	assert.Nil(t, d2.Array)
+	assert.Nil(t, d2.Object)
+	assert.True(t, d2.Old.IsBool())
+	assert.Equal(t, true, d2.Old.BoolValue())
+	assert.True(t, d2.New.IsBool())
+	assert.Equal(t, false, d2.New.BoolValue())
+	d3 := NewProperty(true).Diff(NewNullProperty())
+	assert.NotNil(t, d3)
+	assert.Nil(t, d3.Array)
+	assert.Nil(t, d3.Object)
+	assert.True(t, d3.Old.IsBool())
+	assert.Equal(t, true, d3.Old.BoolValue())
+	assert.True(t, d3.New.IsNull())
+}
+
+func TestNumberPropertyValueDiffs(t *testing.T) {
+	t.Parallel()
+	d1 := NewProperty(42.0).Diff(NewProperty(42.0))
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/diff_test.go
```diff
@@ -617,3 +616,200 @@ func calculateChangesAndReplacementsForOneAddedProperty(t *testing.T, value stri
...
+				case del:
+					delete(newObj, k)
+				}
+				changes[i][k] = recordedOp{operation, i}
+			}
+			new[i].V = newObj
+		}
+
+		// fmt.Println("OLD", old)
+		// fmt.Println("UPDATES", changes)
```
t0yv0, line 769: t.Logf(..) could be good to keep.

provider/pkg/provider/diff.go
```diff
@@ -16,9 +18,185 @@ import (
...
+// mapProperties traverses the given APIResource and creates a map from all properties, including
+// in referenced types, to their property spec. The map key is a JSONPath to the property.
+func mapProperties(res resources.AzureAPIResource, lookupType resources.TypeLookupFunc) map[string]resources.AzureAPIProperty {
+	properties := map[string]resources.AzureAPIProperty{}
+	for _, p := range res.PutParameters {
+		if p.Location == body && p.Body != nil {
+			resources.TraverseProperties(p.Body.Properties, lookupType,
+				false, // don't include flattened containers, we're comparing the SDK shape
+				func(name string, prop resources.AzureAPIProperty, path []string) {
+					path = append(path, name)
```
t0yv0, line 34: Just curious, property path fragments are never array elements or some such nameless things (like in bridged providers)?

t0yv0:
Popping in here, very nice use of rapid, but I don't have enough background context on the azure-native structures to review thoroughly. Let me know if there's any generic learnings from this rollout. Bridged providers have TF-defined set semantics that we get wrong sometimes as well and it's a little bit of a struggle. I also feel we could do better logging and explaining what the diff algo is deciding so as to be able to figure out bugs faster when they are reported in the field. LMK if anything here could apply to the bridged world as well.

provider/pkg/provider/diff_test.go
thomas11: The output of this is easily 100+ lines, so I opted for a different approach, keeping this code in an unused func that developers can call when they need more data.

provider/pkg/provider/diff.go
thomas11: I haven't come across that case. `TraverseProperties` only recurses through properties who have a reference to `#/types/azure-native...` which resolves again to a type with named properties.

provider/pkg/provider/diff.go
```diff
@@ -16,9 +19,189 @@ import (
...
+
+		if new, has := newInputs[k]; has {
+			// If a new exists, use it; for output properties, however, ignore differences.
+			if new.IsOutput() {
+				sames[k] = old
+			} else if diff := valueDiff(properties, old, new, addToPath(path, k)); diff != nil {
+				if !old.HasValue() {
+					adds[k] = new
+				} else if !new.HasValue() {
+					deletes[k] = old
```
danielrbradley, line 94: These lines are currently uncovered by the new unit tests. I'm not exactly sure what data's required to go down here. Is it a specific variation of special values to trigger these three paths?

provider/pkg/provider/diff.go
```diff
@@ -16,9 +19,189 @@ import (
...
+		return &resource.ValueDiff{
+			Old: old,
+			New: new,
+			Array: &resource.ArrayDiff{
+				Adds:    adds,
+				Deletes: deletes,
+				Sames:   sames,
+				Updates: updates,
+			},
+		}
```
danielrbradley, line 180: Would be good to have a unit test covering the non-nil case - with at least one add, delete and update.

provider/pkg/provider/diff.go
thomas11: This is code that I copied from pulumi/pulumi/sdk/go/common/resource/properties_diff.go to be able to modify the array handling, as outlined in #2781.

It may be best to just copy pulumi/pulumi/sdk/go/common/resource/properties_diff_test.go so we don't lose coverage, WDYT?

danielrbradley: Yup, that sounds sensible - that will hopefully serve as a good spec for the intended behaviour.

Issue #2577, opened 2023-07-14
`VnetConfiguration` on `App.ManagedEnvironment` not applied on actual resource


ringods:
### What happened?

When updating an existing `App.ManagedEnvironment` resource with a `VnetConfiguration` input, the vnet configuration is not applied on the managed environment resource.

### Expected Behavior

The managed environment resource to be updated with the given vnet configuration.

### Steps to reproduce

Code sample used to reproduce the issue:

```csharp
using System.Collections.Generic;
using Pulumi;
using Pulumi.AzureNative.App;
using Pulumi.AzureNative.App.Inputs;
using Pulumi.AzureNative.Network;
using Pulumi.AzureNative.Network.Inputs;
using Pulumi.AzureNative.Resources;

return await Pulumi.Deployment.RunAsync(() =>
{
    var resourceGroupName = "managedenv"; // Provide existing resource group name

    // Assuming the existing resource group is available under the given resourceGroupName
    var resourceGroup = new ResourceGroup(resourceGroupName);

    // Create a virtual network
    // var virtualNetwork = new VirtualNetwork("virtualNetwork", new VirtualNetworkArgs
    // {
    //     ResourceGroupName = resourceGroup.Name,
    //     Location = resourceGroup.Location,
    //     AddressSpace = new AddressSpaceArgs
    //     {
    //         AddressPrefixes =
    //                 {
    //                     "10.0.0.0/16",
    //                 },
    //     },
    // });

    // Create a subnet in the virtual network
    // var subnet = new Subnet("subnet", new Pulumi.AzureNative.Network.SubnetArgs
    // {
    //     ResourceGroupName = resourceGroup.Name,
    //     VirtualNetworkName = virtualNetwork.Name,
    //     AddressPrefix = "10.0.0.0/22",
    // });

    var managedEnvironment = new ManagedEnvironment("TestManagedEnv", new ManagedEnvironmentArgs
    {
        ResourceGroupName = resourceGroup.Name,
        EnvironmentName = "TestManagedEnv",
        Location = resourceGroup.Location,
        // VnetConfiguration = new Pulumi.AzureNative.App.Inputs.VnetConfigurationArgs
        // {
        //     InfrastructureSubnetId = subnet.Id,
        // },
        Tags =
                {
                    { "Environment", "Dev" },
                },
    });

    return new Dictionary<string, object?>
    {
        ["ManagedEnvironmentName"] = managedEnvironment.Name,
    };
});
```

[...]

PR #2970, opened 2023-12-21
Set ForceNew on referenced types

### LINKED ISSUE #2782: Enable auto re-creation on sub-property changes
## Hello!

## Issue details

Currently, we skip setting `forceNew` for properties within types (sub properties of a resource):

https://github.com/pulumi/pulumi-azure-native/blob/8f12571fc1f62281465f77385145248ddbfa99e6/provider/pkg/gen/properties.go#L350-L357

This is due to it potentially causing resource recreation when it is not actually needed, which a user cannot opt out of. 

As far as I am aware there's two causes for this over-zealous recreation:
1. Our diffs are incorrect for arrays due to not taking into account the items identifier:
    - https://github.com/pulumi/pulumi-azure-native/issues/2781
2. Our application of the mutability metadata is inaccurate.

### Correct application of mutability metadata

When a property is an array of sub-objects and one of the sub-objects is added or removed, this should not flag for the whole resource recreation. If the property of the sub-object is marked as "not updatable", then we only can't change that value for that item within the array.

Our current implementation just looks at the naive diff to see if there's any properties with changes, but this doesn't account correctly for:
- New items being to the collection
- Items being removed from the collection
- Items being re-ordered within the collection.

None of these actions should trigger replacement. The only scenario which should trigger recreation is the modification of a sub-object value when comparing to the old state for the same sub-object identifier.

### Example

Suppose there was a schema with a resource with a single property called `foos`, and each `foo` was keyed by `name` but has a property called `value` which is not updatable.

```json
  {
    "Resource": {
      "properties": {
        "foos": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Foo"
          },
          "x-ms-identifiers": ["name"],
        }
      }
    },
    "Foo": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "value": {
          "type": "string",
          "x-ms-mutability": [
            "create",
            "read"
          ]
        }
      },
      "required": [
        "name"
      ],
    }
  }
```

Initial investigtion and test work is captured in this draft PR: https://github.com/pulumi/pulumi-azure-native/pull/2777

### Affected area/feature

- Diff
- Codegen
- Metadata

thomas11:
_This PR is on top of #2968._

This PR enables auto re-creation on sub-property changes which resolves #2782. As that issue points out, the reason this was disabled so far was to prevent over-eager recreation of resources when unnecessary. With correct diffing of keyed arrays, this might be fixed. Further testing needed, therefore this is a draft PR.

The core implementation is a rather small change in properties.go. We already had the data about referenced types to find properties where `forceNew` should be set, but avoided doing so. Now we can enable this. We use a new metadata flag rather than the existing `forceNew` so we can treat the referenced type properties differently, namely to provide an escape hatch from the new forced recreation. 

Since unwanted recreation of resources is very disruptive, this PR includes an escape hatch, i.e., a way to opt-out of this change. For now, it's actually an opt-in to be cautious about breaking user programs in a minor version. To this end, a new environment variable `PULUMI_FORCE_NEW_FROM_SUBTYPES` can be set to "true" to enable the behavior described above. _Question: should this be a proper Pulumi config setting?_

It works in manual testing for the Managed Environment VNet example from #2577.  Flipping a property on the environment's `vnetConfiguration` object results in

```
❯ PULUMI_DEBUG_PROVIDERS="azure-native:58241" pulumi up -s dev2
Previewing update (dev2)

     Type                                    Name                   Plan        Info
     pulumi:pulumi:Stack                     azn-2577-vnet-ts-dev2              3 warnings
 +-  └─ azure-native:app:ManagedEnvironment  appEnvironment         replace     [diff: ~vnetConfiguration]

Resources:
    +-1 to replace
    4 unchanged

Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev2::azn-2577-vnet-ts::pulumi:pulumi:Stack::azn-2577-vnet-ts-dev2]
    ++azure-native:app:ManagedEnvironment: (create-replacement)
        [id=/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/resourceGroupc34a3e63/providers/Microsoft.App/managedEnvironments/appEnvironment7f502e85]
        [urn=urn:pulumi:dev2::azn-2577-vnet-ts::azure-native:app:ManagedEnvironment::appEnvironment]
        [provider=urn:pulumi:dev2::azn-2577-vnet-ts::pulumi:providers:azure-native::default_2_0_0_alpha_0_dev::b2b85f57-0239-406b-81f5-d1c764b43587]
      ~ vnetConfiguration: {
          ~ internal: false => true
        }
    +-azure-native:app:ManagedEnvironment: (replace)
        [id=/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/resourceGroupc34a3e63/providers/Microsoft.App/managedEnvironments/appEnvironment7f502e85]
        [urn=urn:pulumi:dev2::azn-2577-vnet-ts::azure-native:app:ManagedEnvironment::appEnvironment]
        [provider=urn:pulumi:dev2::azn-2577-vnet-ts::pulumi:providers:azure-native::default_2_0_0_alpha_0_dev::b2b85f57-0239-406b-81f5-d1c764b43587]
      ~ vnetConfiguration: {
          ~ internal: false => true
        }
    --azure-native:app:ManagedEnvironment: (delete-replaced)
        [id=/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/resourceGroupc34a3e63/providers/Microsoft.App/managedEnvironments/appEnvironment7f502e85]
        [urn=urn:pulumi:dev2::azn-2577-vnet-ts::azure-native:app:ManagedEnvironment::appEnvironment]
        [provider=urn:pulumi:dev2::azn-2577-vnet-ts::pulumi:providers:azure-native::default_2_0_0_alpha_0_dev::b2b85f57-0239-406b-81f5-d1c764b43587]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,12 @@ quality and higher fidelity with the Azure platform.
 
 To learn how to configure credentials refer to the [Azure configuration options](https://www.pulumi.com/registry/packages/azure-native/installation-configuration/#configuration-options).
 
+## Other Configuration
+
+In addition to the configuration options in the official documentation linked above, the following environment variables can be used to tweak lower-level behavior of the provider:
+
+  - `PULUMI_FORCE_NEW_FROM_SUBTYPES`: if, after a change in your program, the provider does not replace a resource that should be replaced because it cannot be updated, setting this variable to `true` might allow the provider to infer the correct behavior. For more details please see PR #2970. We're planning to make this behavior the default in the future.
+
 ## Building
 
 See [contributing](CONTRIBUTING.md) for details on how to build and contribute to this provider.


provider/pkg/provider/diff.go
```diff
@@ -243,8 +244,9 @@ func calculateDetailedDiff(resource *resources.AzureAPIResource, lookupType reso
 }
 
 func findForceNew(base string, props map[string]resources.AzureAPIProperty, replaceKeys codegen.StringSet) {
+	forceNewFromSubtypes := os.Getenv("PULUMI_FORCE_NEW_FROM_SUBTYPES") == "true"
 	for propName, prop := range props {
-		if prop.ForceNew {
+		if prop.ForceNew || (prop.ForceNewInferredFromReferencedTypes && forceNewFromSubtypes) {
```
danielrbradley, line 249: Ah, is this why return `false` on the `forceNew` return when we're forcing from a referenced type?

danielrbradley (commented):
I _think_ the implementation looks correct, however, it feels like there should be some kind of test to reflect this behaviour change and ensure we don't regress.

Secondly, I think we should document the ENV variable in the readme or somewhere else that's discoverable.

Thirdly, the two bool returns and their interpretation are a little confusing and appeared incorrect on first reading. I wonder if there's a clearer way modelling this e.g. triggerReplacement (bool) + reason (top/nested) ... or something else?

thomas11:
> I _think_ the implementation looks correct, however, it feels like there should be some kind of test to reflect this behaviour change and ensure we don't regress.
> 
> Secondly, I think we should document the ENV variable in the readme or somewhere else that's discoverable.
> 
> Thirdly, the two bool returns and their interpretation are a little confusing and appeared incorrect on first reading. I wonder if there's a clearer way modelling this e.g. triggerReplacement (bool) + reason (top/nested) ... or something else?

Great feedback. The last two commits hopefully improve on this.

danielrbradley (approved):
Looks great :shipit:

PR #2993, opened 2024-01-04
Case-insensitive comparison of Azure ids for more correct diffs

### LINKED ISSUE #854: Pulumi reports differences caused by letters (uppercase/lowercase) in Azure URN
The command `pulumi preview` reports differences caused by letters (uppercase/lowercase) in Azure URN against the fact that URNs are case insensitive. The reporting of the difference cause is missing in a console output.

## Expected behavior
Pulumi should not differentiate between uppercase and lowercase letters in Azure URNs. I expect that URNs" `/subscriptions/<subscription id>/**resourcegroups**/<resource group name>/providers/microsoft.insights/actiongroups/<name>` and `/subscriptions/<subscription id>/**resourceGroups**/<resource group name>/providers/microsoft.insights/actiongroups/<name>` are the same. Difference between **resourcegroups** and **resourceGroups** should not lead to Warning during the `pulumi preview` action.

## Current behavior
Pulumi should not report warning caused by change in Azure URN regarding uppercase/lowercase letters. It is not warning nor error.

## Steps to reproduce

1. Create manually metric alert focused on `Microsoft.Automation/automationAccounts` resource type.
2. Import resource under Pulumi management via `pulumi import`
3. Change the part a URN string of `ActionGroupId` within the source code generated: from `**resourcegroups**` to `**resourceGroups**`.
4. Make a preview of this stack via `pulumi preview` command as `pulumi preview --logtostderr -v=9 2> out.txt.`
5. Look at the warning on console and the test `detaileddiff=map[actions[0].actionGroupId:kind:UPDATE inputDiff:true ]` within the detailed trace.

Source code of a metric alert stack:
```csharp
class MyStack : Stack
{
    public MyStack()
    {
        var alert = new AzureNative.Insights.MetricAlert(<...provide alert name...>, new AzureNative.Insights.MetricAlertArgs
        {
            Actions =
            {
                new AzureNative.Insights.Inputs.MetricAlertActionArgs
                {
                    ActionGroupId = "/subscriptions/<...provide subscription id...>/resourcegroups/<...provide resource group name...>/providers/microsoft.insights/actiongroups/<...provide action group name...>",
                },
            },
            AutoMitigate = true,
            Criteria =
            {
                { "allOf",
                {

                    {
                        { "criterionType", "StaticThresholdCriterion" },
                        { "dimensions",
                        {
                            new AzureNative.Insights.Inputs.MetricDimensionArgs
                            {
                                Name = "Status",
                                Operator = "Include",
                                Values =
                                {
                                    "Failed",
                                },
                            },
                        } },
                        { "metricName", "TotalJob" },
                        { "metricNamespace", "Microsoft.Automation/automationAccounts" },
                        { "name", "Metric1" },
                        { "operator", "GreaterThan" },
                        { "threshold", 0 },
                        { "timeAggregation", "Count" },
                    },
                } },
                { "odataType", "Microsoft.Azure.Monitor.SingleResourceMultipleMetricCriteria" },
            },
            Description = "",
            Enabled = true,
            EvaluationFrequency = "PT1M",
            Location = "global",
            ResourceGroupName = <....provide resource group name.....>,
            RuleName = <....provide rule name.....>,
            Scopes =
            {
                "/subscriptions/<...provide subscription id...>/resourceGroups/<...provide resource group name...>/providers/Microsoft.Automation/automationAccounts/<...provide automation account name...>",
            },
            Severity = 3,
            TargetResourceType = "Microsoft.Automation/automationAccounts",
            WindowSize = "PT1M",
        }, new CustomResourceOptions
        {
            Protect = true,
        });
    }
}
```
pulumi-console-warning

## Context (Environment)
The type of such warning causes slowing down the code development and resource deployment activities. Nothing is changed in Azure after deploy of a particular resource, this warning should not be there.

thomas11:
This PR aims to resolve #854. Basically, the issue is that the common prefix of Azure IDs, `/subscriptions/${sub}/resourcegroups/${rg}/providers/microsoft.${namespace}/`, is case-insensitive, but we diff it as if it weren't. This leads to unnecessary updates as well as delete-replaces in keyed arrays where the Azure id property is a key, like in #854.

The PR introduces a string comparison that, if the string looks like an Azure id, compares the common prefix case-insensitively, but still compares the part after the prefix case-sensitively since we cannot know if names of various resource providers are case-insensitive.

We already have the `applyAzureSpecificDiff` func, but it handles only top-level properties and wouldn't help with keyed arrays, so I went the route of applying the id normalization to all strings.

The downside is that might be a bit expensive. I tried to delay the regex matching where possible.

This PR fixes the repro in #854 for me.

### DIFF
--- a/provider/pkg/provider/diff.go
+++ b/provider/pkg/provider/diff.go
@@ -206,18 +206,34 @@ func valueDiff(properties map[string]resources.AzureAPIProperty,
 	return &resource.ValueDiff{Old: old, New: new}
 }
 
-var urnRegex = regexp.MustCompile(`(?i)(/subscriptions/.+?/resourcegroups/.+?/providers/microsoft.+?/)(.*)`)
+// azureIdRegex is a regular expression that matches Azure resource IDs.
+var azureIdRegex = regexp.MustCompile(`(?i)(/subscriptions/.+?/resourcegroups/.+?/providers/microsoft.+?/)(.*)`)
 
+// stringsEqualCaseInsensitiveAzureIds compares two strings for equality. If they look like Azure
+// resource IDs, the common prefix `/subscriptions/.+?/resourcegroups/.+?/providers/microsoft.+?/`
+// is compared case-insensitively. The rest is compared case-sensitively since we cannot know if
+// names of various resource providers are case-insensitive. If the strings don't look like Azure
+// resource IDs, they are compared using regular `==` string comparison.
 func stringsEqualCaseInsensitiveAzureIds(a, b string) bool {
 	// Compare plain strings first to save the regex matching cost
 	if a == b {
 		return true
 	}
-	return normalizeAzureId(a) == normalizeAzureId(b)
+	matchesA := azureIdRegex.FindStringSubmatch(a)
+	if len(matchesA) > 0 {
+		matchesB := azureIdRegex.FindStringSubmatch(b)
+		if len(matchesB) > 0 {
+			return strings.EqualFold(matchesA[1], matchesB[1]) && matchesA[2] == matchesB[2]
+		}
+	}
+	return false
 }
 
+// normalizeAzureId normalizes an Azure resource ID by lowercasing the common prefix
+// `/subscriptions/.+?/resourcegroups/.+?/providers/microsoft.+?/`. If the string doesn't look like
+// an Azure resource ID, it is returned unchanged.
 func normalizeAzureId(id string) string {
-	match := urnRegex.FindStringSubmatch(id)
+	match := azureIdRegex.FindStringSubmatch(id)
 	if len(match) > 0 {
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
This is a great one to fix - I like the splitting on the very specific pattern.

Let's add a comment to the method to explain why we're only looking for the specific section and leaving the other unchanged.

Nit: technically I think it's better to use a case insensitive comparison (using `strings.EqualFold`) instead of doing ToLower on both parts - so make `normalizeAzureId` into `compareAzureId` - split both using the regex, compare the first sections using EqualFold, then compare the last section using strict equality. But, in practice I don't think it makes any difference here because I don't think there should be any non-latin characters in this section of the URL.

PR #2997, opened 2024-01-05
Add playbook documenting existing customization points

thomas11:
I documented the existing places - that I'm aware of - where we override or customize behavior in azure-native. I hope this can achieve two purposes:
  1. Serve as a playbook for the next person who runs into one of these issues
  2. Serve as an overview to show us where we should invest in more robust code, or public extension points.

### DIFF
--- a/playbooks/customizing-resources.md
+++ b/playbooks/customizing-resources.md
@@ -10,27 +10,31 @@ Override: change the default API version in versions/v2-spec.yaml and versions/v
 See: [Specify custom API version on resources #2467](https://github.com/pulumi/pulumi-azure-native/issues/2467), [Change the default API version of AlertProcessingRuleByName to 2021-08-08 #2921](https://github.com/pulumi/pulumi-azure-native/pull/2921) (example), [Change the default version of insights/ActivityLogAlert to supported 2020-10-01 #2941](https://github.com/pulumi/pulumi-azure-native/pull/2941) (example)
 
 ### A GET method (invoke) is missing
+We only create getFoo or listFoo functions in combination with a Foo resource. We don't turn all GET endpoints to functions because most would simply duplicate the Pulumi Read of the resource. There are exceptions that are useful, however, that can be added here.
+
 Override: openapi/additionalInvokes.go.
 
 Note: don’t do this for POST endpoints, invokes need to be side-effect free.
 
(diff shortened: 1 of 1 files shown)


playbooks/customizing-resources.md
```diff
@@ -0,0 +1,57 @@
...
+First, try to improve the code to do the right thing automatically. Special cases are expensive to maintain. If that doesn't work out, see the below list of existing override/customization locations.
+
+## Availability of resources
+
+### The wrong API version is used
+Override: change the default API version in versions/v2-spec.yaml and versions/v2-lock.yaml.
+
+See: [Specify custom API version on resources #2467](https://github.com/pulumi/pulumi-azure-native/issues/2467), [Change the default API version of AlertProcessingRuleByName to 2021-08-08 #2921](https://github.com/pulumi/pulumi-azure-native/pull/2921) (example), [Change the default version of insights/ActivityLogAlert to supported 2020-10-01 #2941](https://github.com/pulumi/pulumi-azure-native/pull/2941) (example)
+
+### A GET method (invoke) is missing
```
mjeffryes, line 12: I don't understand this case; does this mean that the schema is in the spec, but just the GET method is not included in the spec?

playbooks/customizing-resources.md
```diff
@@ -0,0 +1,57 @@
...
+
+Note: don’t do this for POST endpoints, invokes need to be side-effect free.
+
+See: Add selected GET invokes #2922
+
+### An auto-generated resource name looks wrong
+Override: resources/resources.go `ResourceName()` has a list of exceptions.
+
+### A resource should not be included in Azure Native
+Override: gen/exclude.go.
```
mjeffryes, line 23: I'm very curious to understand more why these cases arise. Especially, why would we not want to include a resources that's defined in the spec?

mjeffryes (commented):
Whoo! Very happy to see this. As a n00b, it would be nice to have a 1-2 sentence description of all the cases, (you did this for some things and I found it very instructive!) but this is already a big help toward understanding some of the ways we have had to customize on top of the spec

mjeffryes (approved):

playbooks/customizing-resources.md
thomas11: added

PR #2998, opened 2024-01-05
Fix pypi credentials again

blampe:
The publish step was using a hard-coded username instead of the `PYPI_USERNAME` environment variable.

The SDK gen step doesn't publish to PyPI, so we can remove those credentials.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -160,7 +160,7 @@ jobs:
       - name: Publish PyPi Package
         run: >
           twine upload
-          -u "pulumi"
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -160,7 +160,7 @@ jobs:
       - name: Publish PyPi Package
         run: >
           twine upload
-          -u "pulumi"
+          -u "${PYPI_USERNAME}"
```
guineveresaenger, line 163: d'oh!

guineveresaenger (approved):

danielrbradley (approved):

PR #3002, opened 2024-01-08
Add report about skipped POST endpoints during schemagen

thomas11:
There's a decent number of [issues related to missing Azure functionality](https://github.com/pulumi/pulumi-azure-native/issues?q=is%3Aissue+is%3Aopen+POST+PUT) which we omit because the endpoints are POST-based and don't fit our resource model.

This PR adds a report to have a complete list of these endpoints and detect patterns that might help us to add them without coding them up one by one.


provider/pkg/openapi/discover.go
```diff
@@ -341,16 +359,16 @@ var excludeRegexes = []*regexp.Regexp{
 
 // addAPIPath considers whether an API path contains resources and/or invokes and adds corresponding entries to the
 // provider map. `providers` are mutated in-place.
-func (providers AzureProviders) addAPIPath(specsDir, fileLocation, path string, swagger *Spec) []resources.NameDisambiguation {
+func (providers AzureProviders) addAPIPath(specsDir, fileLocation, path string, swagger *Spec) DiscoveryDiagnostics {
```
danielrbradley, line 362: Let's make the return a pointer to make the returns a little cleaner and give easier access to the methods after it's returned.

danielrbradley (approved):
Looks good .. one inline tweak.

PR #3004, opened 2024-01-09
Don't delete Service Plan when deleting last Web App

### LINKED ISSUE #1529: Destroying web app also destroys service plan
## Hello!

## Issue details

When removing the last web app in an app service plan, the service plan is also deleted.
So, if the service plan was created from another stack, then that stack needs to be refreshed and re-applied before the apps can be brought back.
Looking at the documentation [here](https://docs.microsoft.com/en-us/rest/api/appservice/web-apps/delete#uri-parameters), I am wondering if Pulumi should be specifying `deleteEmptyServerFarm=false` when talking to the REST API.

### Steps to reproduce

1. Create a stack with a service plan and web app.
2. Pulumi up.
2. Remove the web app, leaving the service plan.
3. Pulumi up.

Expected result: the service plan still exists in Azure.
Actual result: the service plan was automatically removed from Azure.

thomas11:
Resolves #1529 

By default, when deleting the only Web App in a Service Plan, Azure deletes the Service Plan as well. This doesn't mesh with the Pulumi resource model since the Service Plan is a separate resource.

The default Azure behavior can be customized with a query parameter `deleteEmptyServerFarm` on the DELETE request. However, so far we don't have any support for delete parameters. I saw two ways of approaching this:
1. Add support for delete parameters to the schema, and pass them through so that the regular delete operation in the provider does the right thing.
2. Implement a custom resource for Web App, with only the delete method, re-using as much provider code as possible.

The problem with (1) is that it would expose the `deleteEmptyServerFarm` parameter via the schema but we'd still not know which value it would be set to without custom handling for the Web App case. This PR implements approach (2). I'm open to changing course, though.

In order to be able to re-use the provider's existing delete logic, the required methods are encapsulated in a new interface `AzureClient` which can be passed to the custom resources factory. This will probably also be useful to support POST-based operations, as it would be simple to add the `azurePost` method to the interface and call it from custom resources.

Tests are still TODO, but this PR fixed the issue for me in a local, manual test.

### DIFF
--- a/provider/pkg/resources/custom_webapp_delete.go
+++ b/provider/pkg/resources/custom_webapp_delete.go
@@ -10,12 +10,12 @@ import (
 const webAppResourceType = "azure-native:web:WebApp"
 
 // https://github.com/pulumi/pulumi-azure-native/issues/1529
-func customWebAppDelete(azureClient AzureClient) *CustomResource {
+func customWebAppDelete(resourceLookupper ResourceLookupper, deleter AzureDeleter) *CustomResource {
 	return &CustomResource{
 		path: "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}",
 		tok:  webAppResourceType,
 		Delete: func(ctx context.Context, id string, properties resource.PropertyMap) error {
-			res, ok, err := azureClient.LookupResource(webAppResourceType)
+			res, ok, err := resourceLookupper.LookupResource(webAppResourceType)
 			if err != nil {
 				return err
 			}
@@ -23,7 +23,7 @@ func customWebAppDelete(azureClient AzureClient) *CustomResource {
 				return fmt.Errorf("resource %q not found", webAppResourceType)
 			}
 
-			return azureClient.AzureDelete(ctx, id, res.APIVersion, res.DeleteAsyncStyle, map[string]any{
+			return deleter.AzureDelete(ctx, id, res.APIVersion, res.DeleteAsyncStyle, map[string]any{
 				// Don't delete the app service plan even if this was the last web app in it. It's
 				// a separate Pulumi resource.
 				"deleteEmptyServerFarm": false,
--- a/provider/pkg/resources/custom_webapp_delete_test.go
+++ b/provider/pkg/resources/custom_webapp_delete_test.go
@@ -8,23 +8,25 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-type TestAzureClient struct {
-	queryParamsOfLastDelete map[string]any
-}
+type MockResourceLookupper struct{}
 
-func (t *TestAzureClient) LookupResource(id string) (AzureAPIResource, bool, error) {
+func (t *MockResourceLookupper) LookupResource(id string) (AzureAPIResource, bool, error) {
 	return AzureAPIResource{}, true, nil
 }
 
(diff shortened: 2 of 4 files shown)


provider/pkg/resources/custom_webapp_delete_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (t *TestAzureClient) AzureDelete(ctx context.Context, id, apiVersion, asyncStyle string, queryParams map[string]any) error {
+	t.queryParamsOfLastDelete = queryParams
+	return nil
+}
+
+func TestSetsDeleteParam(t *testing.T) {
+	client := TestAzureClient{}
+	custom := customWebAppDelete(&client)
+	custom.Delete(context.TODO(), "id", resource.PropertyMap{})
```
blampe, line 27:
```suggestion
	custom.Delete(context.Background(), "id", resource.PropertyMap{})
```

provider/pkg/resources/custom_webapp_delete_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/stretchr/testify/assert"
+)
+
+type TestAzureClient struct {
```
blampe, line 11: Nice mock!

provider/pkg/resources/custom_webapp_delete.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+const webAppResourceType = "azure-native:web:WebApp"
+
+// https://github.com/pulumi/pulumi-azure-native/issues/1529
+func customWebAppDelete(azureClient AzureClient) *CustomResource {
+	return &CustomResource{
```
blampe, line 14: I assume this falls back on upstream create/update/etc?

blampe (approved):
Looks reasonable, only a small suggestion re: the interface.

provider/pkg/resources/customresources.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
 
+type AzureClient interface {
+	LookupResource(resourceType string) (AzureAPIResource, bool, error)
+	AzureDelete(ctx context.Context, id, apiVersion, asyncStyle string, queryParams map[string]any) error
+}
```
blampe, line 22: Agree this could be helpful for POSTs and elsewhere, but if you extended this to another method like `Create` it doesn't make sense for tests to need to implement `AzureDelete`. I suggest splitting this into a some smaller interfaces so you can mix and match only the things you need:

```
type Lookupper interface {
    Lookup(resourceType string) (AzureAPIResource, bool, error)
}

type Deleter interface {
	Delete(ctx context.Context, id, apiVersion, asyncStyle string, queryParams map[string]any) error}
}

type LookupDeleter interface {
    Lookupper
    Deleter
}
```

provider/pkg/resources/custom_webapp_delete.go
thomas11: correct

provider/pkg/resources/customresources.go
thomas11: Good point, thanks

PR #3007, opened 2024-01-10
Support DefenderForStorageSettings resource by adding default state

### LINKED ISSUE #2996: Add DefenderForStorageSettings resource
## Hello!

## Issue details

I'm trying to change the Microsoft Defender for Storage settings, but can find no way of doing that via Pulumi Azure native. It's available in the REST API (https://learn.microsoft.com/en-us/azure/storage/common/azure-defender-storage-configure?tabs=enable-subscription#rest-api-1):

PUT
https://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/providers/Microsoft.Security/DefenderForStorageSettings/current?api-version=2022-12-01-preview

````json
{
    "properties": {
        "isEnabled": true,
        "malwareScanning": {
            "onUpload": {
                "isEnabled": true,
                "capGBPerMonth": 5000
            }
        },
        "sensitiveDataDiscovery": {
            "isEnabled": true
        },
        "overrideSubscriptionLevelSettings": true
    }
}
````

### PR DESCRIPTION
**Pull request by thomas11**
Support DefenderForStorageSettings resource by adding the default state to reset it to on delete, since it doesn't have an actual delete operation.

Without the hard-coded exception in the last commit, the generated schema is incorrect. The cause is unrelated to the PR. What's happening is that [the Azure spec](https://github.com/Azure/azure-rest-api-specs/blob/fa8683f6725e5447cc808420c4542ce8375f0325/specification/security/resource-manager/Microsoft.Security/preview/2022-12-01-preview/defenderForStorageSettings.json) is defined as such (unrelated properties omitted):
```json
      "properties": {
        "isEnabled": {
          "type": "boolean",
          "description": "Indicates whether Defender for Storage is enabled on this storage account."
        },
        "sensitiveDataDiscovery": {
          "x-ms-client-flatten": true,
          "$ref": "#/definitions/SensitiveDataDiscoveryProperties"
        }
      }
```

Where SensitiveDataDiscoveryProperties is:
```json
      "properties": {
        "isEnabled": {
          "type": "boolean",
          "description": "Indicates whether Sensitive Data Discovery should be enabled."
        }
      }
```

The nested object is annotated with x-ms-client-flatten, causing the object to be flattened, causing the two identical `isEnabled` properties to collide. #3016 will fix this in v3.

In the meantime, I added a manual exception.

Resolves #2996


danielrbradley (approved):
Looks good. Does your other PR have a note so we remember to remove this when it merges?

thomas11:
> Looks good. Does your other PR have a note so we remember to remove this when it merges?

Added.

PR #3013, opened 2024-01-11
Add new report to track properties that get overwritten during flattening

thomas11:
The Azure spec has an annotation [x-ms-client-flatten](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-client-flatten) which instructs the schema generator to flatten the property into the parent object. However, there are cases where this leads to overwriting a property, creating incorrect schema and SDKs, when inner and outer property have the same name. I discovered this problem accidentally. First step is to quantify, which this report is for.

### DIFF
--- a/provider/pkg/gen/properties.go
+++ b/provider/pkg/gen/properties.go
@@ -97,11 +97,11 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, isOutput
 				return nil, err
 			}
 
+			// Check that none of the inner properties already exists on the outer type. This
+			// causes a conflict when flattening, and will probably need to be handled in v3.
 			for propName := range bag.properties {
 				if _, has := result.properties[propName]; has {
 					m.flattenedPropertyConflicts[fmt.Sprintf("%s.%s", name, propName)] = struct{}{}
-					logging.V(5).Infof("WARNING: property '%s' of %s will be overwritten when flattening '%s'\n",
-						propName, m.resourceToken, name)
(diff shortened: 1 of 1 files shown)


provider/pkg/gen/properties.go
```diff
@@ -97,6 +97,14 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, isOutput
 				return nil, err
 			}
 
+			for propName := range bag.properties {
+				if _, has := result.properties[propName]; has {
+					m.flattenedPropertyConflicts[fmt.Sprintf("%s.%s", name, propName)] = struct{}{}
+					logging.V(5).Infof("WARNING: property '%s' of %s will be overwritten when flattening '%s'\n",
```
danielrbradley, line 103: Nit: Do we need the log statement in addition to the report? Or is this likely to to just clutter the output and hide other issues?

provider/pkg/gen/properties.go
```diff
@@ -97,6 +97,14 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, isOutput
 				return nil, err
 			}
 
+			for propName := range bag.properties {
```
danielrbradley, line 100: Nit: this little block could benefit from a brief explanation as to its purpose.

danielrbradley (approved):
Core change looks good. A couple of minor nits inline.

Also, can we add a brief explanation of the report to `reports/README.md`?

PR #3016, opened 2024-01-12
[v3] Skip flattening if properties would clash

thomas11:
Skip the flattening of nested properties indicated by [x-ms-client-flatten](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-client-flatten) if it would lead to overwriting a property, creating incorrect schema and SDKs. This case happens when inner and outer property have the same name, for a report on all occurrences see #3013.

This change is breaking and could therefore only be applied to v3 of the provider. The purpose of the PR is mainly to illustrate the SDK changes that would result.

Note: remember to remove exception added in #3007.


danielrbradley (commented):
Yup, in principle the approach looks good for a breaking change in the next release. 

Did you managed to identify is this is the same approach taken by official Azure tools when encountering the conflicts with flattening?

thomas11:
superceded by #3801

PR #3032, opened 2024-01-19
New report about all REST endpoints, grouped by resource

thomas11:
I wrote this report as part of investigating missing POST endpoints. I found that having a single view of all endpoints with paths and verbs, but without schema/spec clutter, was helpful. A few jq commands, that I'm still fiddling with, could help slice and dice the report.


danielrbradley (approved):
Looks helpful already 👍 

It might be worth customising how this report is outputted - given its size. Perhaps we could actually output it as a folder within the reports - with a file per provider?

thomas11:
> Looks helpful already 👍
> 
> It might be worth customising how this report is outputted - given its size. Perhaps we could actually output it as a folder within the reports - with a file per provider?

I considered it but a) wanted to keep the reporting code simple, and b) was concerned that ad-hoc queries of the report, such as using `jq`, might become harder. I suggest we merge it as-is and can always reconsider. I added a sample `jq` query to the README that can be used as a template to shrink the file to only certain data when needed.

Issue #2798, opened 2023-10-03
Refresh indicates non-existing changes e.g. in dates or diagnostic settings


mikocot:
### What happened?

When running `pulumi refresh` immediately after pulumi deployment we get tons of changes. I can understand that many properties are simply added by Azure and will be reported as drift until the first refresh -> although I'd assume that Pulumi retrieves created resource states and updates its state with it, doesn't it?

Still, big part of the changes seem easily avoidable and related to the dates, for example (multiple different resources):

` ~ createdOn       : "2023-10-03T11:05:20.3536642Z" => "2023-10-03T11:05:20.6306655Z"`

This line does not indicate any change in the state, it's the same date just with different precision. IMO pulumi should compare the dates with reasonable precision and decide if its a change or not.

The same applies to formatting differences:

` ~ createdAt             : "2023-10-02T16:56:45.01Z" => "2023-10-02T16:56:45.01"`

Another easily avoidable change is, which is listed for pretty much every resource
    
`  + createdBy       : "90fd5960-b25a-(...)"`
  
  Even after running the refresh there are still date differences indicated in some resources (event hub consumer group):
  
```
~ createdAt: "2023-10-03T11:04:46.081Z" => "2023-10-03T11:04:46.088Z"
~ updatedAt: "2023-10-03T11:04:46.081Z" => "2023-10-03T11:04:46.088Z"
```

### Example

```
 ~ createdOn       : "2023-10-03T11:05:20.3536642Z" => "2023-10-03T11:05:20.6306655Z"

 ~ createdAt        : "2023-10-02T16:56:45.01Z" => "2023-10-02T16:56:45.01"
    
 + createdBy        : "90fd5960-b25a-(...)"
```
We can provide an example stack to look at.

### Output of `pulumi about`

> 'dotnet build -nologo .' completed successfully
> CLI
> Version      3.86.0
> Go Version   go1.21.1
> Go Compiler  gc
> 
> Plugins
> NAME          VERSION
> azure-native  2.5.0
> dotnet        unknown
> random        4.13.2
> 
> Host
> OS       Microsoft Windows 10 Enterprise
> Version  10.0.19045 Build 19045
> Arch     x86_64
> 
> This project is written in dotnet: executable='C:\Program Files\dotnet\dotnet.exe' version='7.0.401'

### Additional context

currently it's a bit scary to update a production environment with refresh, as the list of changes is impossible to review, ideally there would be a systematic solution to deal with not relevant changes

[...]

Issue #2245, opened 2023-02-15
Refresh is not idempotent for untouched resources


ringods:
### What happened?

When running `pulumi refresh` on Azure resources created with this native provider, the action is not idempotent. During refresh, a number of resource properties are read into the state which:

* are no inputs, e.g. `etag`, `lastModifiedTime`
* are the default values for that property as defined by Microsoft, e.g. `defaultEncryptionScope`, `hasImmutabilityPolicy`

### Expected Behavior

When resources are not modified (aka no drift), a `pulumi refresh` should be an idempotent action.

### Steps to reproduce

I used the following Pulumi C# application setting up 2 blob storage containers, one using the Azure Classic provider and one using this Azure Native provider:

```C#
using Pulumi;
using Pulumi.AzureNative.Resources;
using Pulumi.AzureNative.Storage;
using Pulumi.AzureNative.Storage.Inputs;
using StorageClassic = Pulumi.Azure.Storage;
using System.Collections.Generic;

return await Pulumi.Deployment.RunAsync(() =>
{
    // Create an Azure Resource Group
    var resourceGroup = new ResourceGroup("azNativeRefreshProblem");

    // Create an Azure resource (Storage Account)
    var storageAccount = new StorageAccount("sa", new StorageAccountArgs
    {
        ResourceGroupName = resourceGroup.Name,
        Sku = new SkuArgs
        {
            Name = SkuName.Standard_LRS
        },
        Kind = Kind.StorageV2
    });

    var nativeBlobContainer = new BlobContainer("nativeBlobContainer", new BlobContainerArgs{
        AccountName = storageAccount.Name,
        ContainerName = "nativeblobcontainer",
        ResourceGroupName = resourceGroup.Name,
    });

    var classicBlobContainer = new StorageClassic.Container("classicBlobContainer", new StorageClassic.ContainerArgs{
        StorageAccountName = storageAccount.Name,
        Name = "classicblobcontainer"
    });

    // Export nothing for now.
    return new Dictionary<string, object?>
    {
    };
});
```

Run `pulumi up` so both containers are created. This is our baseline.

## pulumi refresh

Starting from the baseline, we trigger the `refresh` command. Here are the details from this command as reported by the CLI and our Pulumi Service. The output is redacted to hide the Azure subscription ID.

### CLI

```
No resources will be modified as part of this refresh; just your stack's state will be.
 details
  pulumi:pulumi:Stack: (same)
[...]

PR #3042, opened 2024-01-25
Read resource state after each Create and Update operation to avoid spurious diffs

mikhailshilkov:
The code change in this PR is fairly small but it does change the fundamental flow of the requests that the provider makes to Azure API, it's worth a careful consideration. The goal is to minimize the diffs between `pulumi up` and `pulumi refresh`.

### Current situation

We use PUT operations for both resource creation and updates. The provider's Create() and Update() methods call the PUT endpoint associated with a resources and awaits for HTTP response, that may come immediately of after the long-running operation (LRO) succeeds. Still, logically, it's a response from the PUT request, as it's defined in Open API specs and implemented in the actual API. We calculate resource outputs from that response.

When we refresh (or import) resources, the provider's Read() method calls the GET (sometimes it's POST) endpoint of the same resource. It then calculates resource outputs and inputs from it. Open API (potentially) defines the type of GET response separately from PUT response.

Our implicit assumption here is that actually PUT response and GET response have the same shape and the same data for the same resource. This assumption is mostly okay, but not perfect. As #2245 and #2798 report, and `RequireEmptyPreviewAfterRefresh=false` tests demonstrate, the actual payloads of PUT vs. GET may differ. In particular, I noticed the following scenario:

1. Some resource properties have default values that are assigned by Azure.
2. A user does not set any value for such property.
3. PUT response still has no default value.
4. GET response does return the default value.
5. Refresh finds the PUT vs GET discrepancy and updates state.
6. The next Update thinks those values were removed from the code, and suggests updating them.
7. Even though it's a no-op on Azure, it's a diff that the users sees and get confused.

### Proposed change

This PR amends the flow of Create() and Update() by an extra GET call after successful PUT operation. The response of that extra GET is what is used to calculate resource outputs. This guarantees that the shapes and data of Create/Update outputs and Read outputs are going to be the same. It resolves the problematic flow above, as proved by test changes.

The only downside that I can see is that we do an extra GET request for each resource change, which may slow things down. However, we observe that GET endpoints are much faster than PUTs, especially for LRO-based resources. I assert that the correctness win has exceedingly high values compared to this minor slowdown.

Resolve https://github.com/pulumi/pulumi-azure-native/issues/2798
Resolve https://github.com/pulumi/pulumi-azure-native/issues/2245

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -55,23 +55,23 @@ const (
 type azureNativeProvider struct {
 	rpc.UnimplementedResourceProviderServer
 
-	azureClient    azure.AzureClient
-	host           *provider.HostClient
-	name           string
-	version        string
-	subscriptionID string
-	environment    azureEnv.Environment
-	// client          autorest.Client
-	resourceMap     *resources.PartialAzureAPIMetadata
-	config          map[string]string
-	schemaBytes     []byte
-	metadataBytes   []byte
-	fullPkgSpec     *schema.PackageSpec
-	fullResourceMap *resources.AzureAPIMetadata
-	converter       *convert.SdkShapeConverter
-	customResources map[string]*customresources.CustomResource
-	rgLocationMap   map[string]string
-	lookupType      resources.TypeLookupFunc
+	azureClient      azure.AzureClient
+	host             *provider.HostClient
+	name             string
+	version          string
+	subscriptionID   string
+	environment      azureEnv.Environment
+	resourceMap      *resources.PartialAzureAPIMetadata
+	config           map[string]string
+	schemaBytes      []byte
+	metadataBytes    []byte
+	fullPkgSpec      *schema.PackageSpec
+	fullResourceMap  *resources.AzureAPIMetadata
+	converter        *convert.SdkShapeConverter
+	customResources  map[string]*customresources.CustomResource
+	rgLocationMap    map[string]string
+	lookupType       resources.TypeLookupFunc
+	skipReadOnUpdate bool
 }
 
 func makeProvider(host *provider.HostClient, name, version string, schemaBytes []byte,
@@ -230,6 +230,8 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 	if err != nil {
 		return nil, fmt.Errorf("initializing custom resources: %w", err)
 	}
+
+	k.skipReadOnUpdate = k.getConfig("skipReadOnUpdate", "ARM_SKIP_READ_ON_UPDATE") == "true"
 
 	return &rpc.ConfigureResponse{
 		SupportsPreview: true,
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/provider.go
```diff
@@ -1303,6 +1310,20 @@ func (k *azureNativeProvider) Update(ctx context.Context, req *rpc.UpdateRequest
...
 
+		// Read the state of the resource from its Read endpoint. While we already have resource state from
+		// the PUT reponse, it may not match precisely the shape of Read reponses that we also use for Refresh
+		// operations. That discrepancy may cause noisy diffs that are hard for users to reconcile.
+		readResponse, err := k.azureRead(ctx, res.ReadMethod, id+res.ReadPath, res.APIVersion)
+		if err != nil {
+			// Since this read operation was introduced in a minor version of the provider, we choose to ignore
+			// any errors here to avoid user-facing regressions. If no warnings are reported, we should be able
+			// to promote this situation to an error.
+			k.host.Log(ctx, diag.Warning, urn, "Failed to read resource after Update. Please report this issue. See verbose logs for more information.")
```
thomas11, line 1321: Might be worth adding https://www.pulumi.com/docs/support/troubleshooting/#verbose-logging to the message?

thomas11 (commented):
For custom resources, you've left the flow as-is, making it the resource's author's responsibility to return the correct outputs. As we're looking at an increasing number of custom resources, maybe we should remove this responsibility and do an extra read in this case as well?

mikhailshilkov:
> For custom resources, you've left the flow as-is, making it the resource's author's responsibility to return the correct outputs. As we're looking at an increasing number of custom resources, maybe we should remove this responsibility and do an extra read in this case as well?

Good question @thomas11 

I decided against doing so because custom resources define an explicit single schema for outputs, there is no duality analogous to separate PUT/GET response Open API definitions. Since we fully control the implementation of custom resources, it seems natural to me to solve any inconsistencies inside their code.

We should probably add more tests to all custom resources to prove no changes on refresh happen.

Is this convincing to you?

provider/pkg/provider/provider.go
mikhailshilkov: I copied the message from another place, but it’s a good idea, I can adjust both.

danielrbradley (commented):
The tests show that this is a good change and it makes sense for consistency that our state should always come from the GET method response. 

Given it's wide ranging impact across all resources - some of which we don't have tests for - perhaps we should add a feature flag here for people to opt-out of the additional read in case they hit an issue not currently foreseen by us? E.g. `AZURE_SKIP_RE_READ`

mikhailshilkov:
> Given it's wide ranging impact across all resources - some of which we don't have tests for - perhaps we should add a feature flag here for people to opt-out of the additional read in case they hit an issue not currently foreseen by us? E.g. AZURE_SKIP_RE_READ

@danielrbradley  I took a different prevention measure - I basically ignore the error that is returned from GET and proceed with the PUT response, logging a warning to the user. I think this should prevent regressions unless the GET code panics. Just to confirm - do you think the flag is still useful?

danielrbradley:
> > Given it's wide ranging impact across all resources - some of which we don't have tests for - perhaps we should add a feature flag here for people to opt-out of the additional read in case they hit an issue not currently foreseen by us? E.g. AZURE_SKIP_RE_READ
> 
> @danielrbradley I took a different prevention measure - I basically ignore the error that is returned from GET and proceed with the PUT response, logging a warning to the user. I think this should prevent regressions unless the GET code panics. Just to confirm - do you think the flag is still useful?

Hmm, yeah that covers most of the potential edge cases. 

The only other edge cases I could foresee is if the read response is not actually wanted and for some reason the PUT response is what they're actually wanting ... but that seems a very slim chance. The other side is performance - if it increases deploy time significantly for some reason.

Also, does Azure guarentee that reads after modifications are always consistent? Or could we potentially intermittently get the pre-update state due to eventual consistency which could leave our state quite broken?

mikhailshilkov:
> The only other edge cases I could foresee is if the read response is not actually wanted and for some reason the PUT response is what they're actually wanting ... but that seems a very slim chance. The other side is performance - if it increases deploy time significantly for some reason.

Good brainstorming! I added a config flag `skipReadOnUpdate`/`ARM_SKIP_READ_ON_UPDATE`.

> Also, does Azure guarentee that reads after modifications are always consistent? Or could we potentially intermittently get the pre-update state due to eventual consistency which could leave our state quite broken?

As far as I remember, they guarantee consistency if you make requests to the same region. Delays are possible between regions. I'm not too worried, because Read-on-Update is a very common pattern in the classic provider (example [one](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/monitor/monitor_diagnostic_setting_resource.go#L482), [two](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/servicebus/servicebus_topic_resource.go#L218), those are the first two resources I checked, no cherry-picking).

danielrbradley (approved):

PR #3046, opened 2024-01-29
Sketch of automagic POST properties for boolean toggles

thomas11:
- First sketch of the metadata part of fully automagic properties
- First sketch of toggling automagic POST properties
- Regenerate with automagic toggle for appplatform.Service.stopped


t0yv0:
Would appreciate a walk-through if I were to review this, thanks! Unfamiliar territory.

thomas11:
> Would appreciate a walk-through if I were to review this, thanks! Unfamiliar territory.

This PR, and its approach in general, might never see the light of day. I'll let you when there's one that's more promising, though!

PR #3054, opened 2024-01-31
Reset, don't remove sub-resource properties to avoid wrong diffs

### LINKED ISSUE #3049: Refresh for VirtualNetwork with external subnets still shows subnets
I'm running [`vnet-subnets-resolution`](https://github.com/pulumi/pulumi-azure-native/tree/master/examples/vnet-subnets-resolution) example with the following flow.

1. `pulumi up` and deploy all resources, all goes well.
2. `pulumi refresh` and expect no meaningful changes, at least not in `subnets` property of the `external` network. However, I get this: 
image
image

3. Run `pulumi up` again, it shouldn't try deleting the subnets but it does suggest doing so:
image

Expected result: no refresh or update of `subnets`.

I think this regressed since https://github.com/pulumi/pulumi-azure-native/pull/2755, most likely in https://github.com/pulumi/pulumi-azure-native/pull/2950. A quick debugging session shows that the `removeUnsetSubResourceProperties` invocation is trying to delete paths `[properties subnets]` - note the top-level `properties` here. However, it can't delete it from the SDK shape, because `properties` are flattened here, so it's probably operating on incompatible data structures.

If you are interested, I have an automated test for this - will open PRs for it soon.

Azure Native version: 2.27.0.
CLI version: 1.103.1.

thomas11:
This PR fixes a defect introduced by #2950. When handling sub-resource properties that may be maintained as stand-alone resources, we need to overwrite them in several code paths. Before, we set them to their default value `[]` on Create, but removed them entirely on Read. This causes unnecessary diffs on Refresh. This PR standardizes the handling on resetting to the default value `[]`.

Resolves #3049

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -969,7 +969,8 @@ func (k *azureNativeProvider) Create(ctx context.Context, req *rpc.CreateRequest
 // present in the inputs because the user wants to manage them as standalone resources. However,
 // such a property might be required by Azure even if it's not annotated as such in the spec, e.g.,
 // Key Vault's accessPolicies. Therefore, we set these properties to their default value here,
-// an empty array.
+// an empty array. For more details, see section "Sub-resources" in CONTRIBUTING.md.
+//
 // During create, no sub-resources can exist yet so there's no danger of overwriting existing values.
 //
 // The `input` param is used to determine the unset sub-resource properties. They are then reset in
@@ -1183,6 +1184,7 @@ func mappableOldState(res resources.AzureAPIResource, oldState resource.Property
 
 // removeUnsetSubResourceProperties resets sub-resource properties in the outputs if they weren't set in the old inputs.
 // If the user didn't specify them inline originally, we don't want to push them into the inputs now.
+// For more details, see section "Sub-resources" in CONTRIBUTING.md.
 func (k *azureNativeProvider) resetUnsetSubResourceProperties(ctx context.Context, urn resource.URN, sdkResponse map[string]any,
 	oldInputs resource.PropertyMap, res *resources.AzureAPIResource) map[string]any {
 	// Take a deep copy so we don't modify the original which is also used later for diffing.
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
My PR here has the tests that caught the original bug that I filed: https://github.com/pulumi/pulumi-azure-native/pull/3056

provider/pkg/provider/provider_test.go
```diff
@@ -458,20 +430,26 @@ func TestRemoveUnsetSubResourceProperties(t *testing.T) {
...
+					"a policy",
+				},
+			},
+		}
+		actual := provider.resetUnsetSubResourceProperties(ctx, "urn", sdkResponse, oldInputs, res)
+		expected := map[string]any{
 			"properties": map[string]any{
 				"accessPolicies": []any{},
 			},
 		}
```
danielrbradley, line 452: Looks like this is the actual core of the change right? We're now expecting the access policies to still be there but be set to an empty value?

provider/pkg/provider/provider_test.go
```diff
@@ -553,33 +532,53 @@ func TestSetUnsetSubresourcePropertiesToDefaults(t *testing.T) {
 
 	t.Run("unchanged", func(t *testing.T) {
 		body := map[string]any{
-			"properties": map[string]any{
-				"accessPolicies": []any{},
+			"container": map[string]any{
+				"properties": map[string]any{
+					"accessPolicies": []any{},
+				},
```
danielrbradley, line 538: What's the reason for adding the container? Were we missing an edge case?

danielrbradley (commented):
I _think_ this is probably good, though it's still tricky to reason about how wide the impact will be.

I feel like we need a succinct description of this feature written down somewhere - like a short spec which describes the behaviour. Perhaps this could go into the contributing guide where we discuss things like the versioning mechanisms or perhaps a block alongside the tests could be easier to find? 

My gut feel is that the tests are still a little patchy and are missing a few edge cases - though I haven't gathered proper evidence for this.

provider/pkg/provider/provider_test.go
thomas11: Just expanding test coverage. I added a test case below for the SDK-shape case, without the container, so both are covered.

provider/pkg/provider/provider_test.go
thomas11: Correct. Because on `Create`, we set such properties to their default value, which is `[]`, so we want it here, too, to avoid diffs.

danielrbradley (approved):

PR #3056, opened 2024-01-31
Enable RequireEmptyPreviewAfterRefresh in all tests

mikhailshilkov:
This PR enables [`RequireEmptyPreviewAfterRefresh`](https://github.com/pulumi/pulumi/pull/15309) for all tests except three that fail with it. There is no functional provider change here. A follow-up PR with fix the provider for those three tests and re-enable them.

I had to sprinkle a few `ignoreChanges` in several places. Once https://github.com/pulumi/pulumi-azure-native/issues/3049 is fixed, we can remove some of them. The remaining ones are legit, and I added comments why.


.github/workflows/build-test.yml
```diff
@@ -86,7 +86,7 @@ jobs:
     # Use big runner for dotnet and nodejs because we need more memory and more compute, respectively
     runs-on: ${{ (matrix.language == 'dotnet' || matrix.language == 'nodejs' || matrix.language == 'go') && 'pulumi-ubuntu-8core' || 'ubuntu-latest' }}
     strategy:
-      fail-fast: true
+      fail-fast: false
```
mikhailshilkov, line 89: Temporarily toggled this to make sure I see all failures, will revert before merging.

examples/go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/pulumi/pulumi-azure-native/examples
 go 1.21
 
 require (
-	github.com/pulumi/pulumi/pkg/v3 v3.91.1
-	github.com/pulumi/pulumi/sdk/v3 v3.91.1
+	github.com/pulumi/pulumi/pkg/v3 v3.103.2-0.20240129165945-b382354fbd66
```
mikhailshilkov, line 6: Will upgrade to 3.104.0 when it's out.

thomas11 (approved):
It's not quite clear to me why we need both `ignoreChanges` and `RequireEmptyPreviewAfterRefresh = false` for these tests - shouldn't `ignoreChanges` take care of the empty preview?

mikhailshilkov:
> It's not quite clear to me why we need both ignoreChanges and RequireEmptyPreviewAfterRefresh = false for these tests - shouldn't ignoreChanges take care of the empty preview?

Yeah, good point. That's a temporary artifact that this PR is a part of my overall work where I will get to the place of removed `RequireEmptyPreviewAfterRefresh = false` but that requires code changes. I could have skipped `ignoreChanges` for those tests in this PR, they are not needed here. Let me know if you want me to do so, or if I can merge as-is and open a follow-up PR that enables all RequireEmptyPreviewAfterRefresh.

thomas11:
> > It's not quite clear to me why we need both ignoreChanges and RequireEmptyPreviewAfterRefresh = false for these tests - shouldn't ignoreChanges take care of the empty preview?
> 
> Yeah, good point. That's a temporary artifact that this PR is a part of my overall work where I will get to the place of removed `RequireEmptyPreviewAfterRefresh = false` but that requires code changes. I could have skipped `ignoreChanges` for those tests in this PR, they are not needed here. Let me know if you want me to do so, or if I can merge as-is and open a follow-up PR that enables all RequireEmptyPreviewAfterRefresh.

If you're tracking the cleanup work after your next PR, I don't care about some extra statements in this one. Thanks for getting after this.

Issue #1464, opened 2022-02-01
Consider refactoring from go-autorest to azcore


mikhailshilkov:
From [this comment](https://github.com/Azure/go-autorest/issues/634#issuecomment-982140032):

> we will be deprecating go-autorest in the coming year (exact date is still being worked out). It will be replaced by azcore. If you're interested in the LRO-specific bits you can find them here.

Consider switching to `azcore` for the provider.

PR #3062, opened 2024-02-02
Extract low-level Autorest Azure client from provider

thomas11:
This PR is a refactoring that's not supposed to change any behavior. The goal were to
  - separate the lower level of Azure request handling from the core provider logic
  - make the huge provider.go file a bit more pliable
  - prepare for the migration off of go-autorest which we need to tackle at some point: #1464

As a boundary between the `provider` module and the new `azure` module I used the following guideline, which is implemented in this PR:
  - `provider` does not depend on github.com/Azure/go-autorest/autorest (encapsulated by `azure`)
  - `azure` does not depend on github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources (resources are decomposed into primitive HTTP request arguments in `provider`)

As a wart, `provider` still depends on github.com/Azure/go-autorest/autorest/azure for handling of Azure environments (global, China, etc.), but that's a lesser concern.

I'm aware the new `azure` module has zero tests so far, but its code didn't have any before and most of it is very side-effect heavy.

### DIFF
--- a/provider/pkg/azure/client.go
+++ b/provider/pkg/azure/client.go
@@ -25,9 +25,13 @@ const (
 `
 )
 
-type AzureClient interface {
-	CanCreate(ctx context.Context, id, path, apiVersion, readMethod string, isSingletonResource, hasDefaultBody bool, isDefaultResponse func(map[string]any) bool) error
+type AzureDeleter interface {
 	Delete(ctx context.Context, id, apiVersion, asyncStyle string, queryParams map[string]any) error
+}
+
+type AzureClient interface {
+	AzureDeleter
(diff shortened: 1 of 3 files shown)


provider/pkg/azure/client.go
```diff
@@ -0,0 +1,468 @@
...
+	updateMethod string,
+	asyncStyle string) (map[string]interface{}, bool, error) {
+
+	var op autorest.PrepareDecorator
+	switch updateMethod {
+	case "PATCH":
+		op = autorest.AsPatch()
+	default:
+		op = autorest.AsPut()
+	}
```
danielrbradley, line 259: This is suprising for a method called PUT that you can then pass in another method (PATCH). I think this method switching would be better to remain in the provider to keep the client more purely modelling the HTTP operations.

danielrbradley (approved):

PR #3065, opened 2024-02-04
Add test_examples CI job to test against p/examples test suite

thomas11:
Increase test coverage for changes by extending testing to the existing test suite in pulumi/examples.

The approach is basically to
- check out p/examples in addition to p/pulumi-azure-native
- import the test suite from there, via a Go `replace` directive
- run as regular ProgramTests with the same config than the repo's own tests, to inject the locally built provider and SDKs.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -222,8 +222,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: pulumi/examples
-          # TODO,tkappler change to master when merged
-          ref: tkappler/separate-tests-by-provider
+          ref: master
           path: p-examples
 
       - name: Setup


t0yv0:
This is one of those times where I'm 👍 on pragmatic grounds, whatever it takes to get that extra coverage is super helpful. Out of curiosity though did you have to reject a previous approach you demoed sometime where the indirection was solved completely on the ProgramTest level, e.g. presumably you'd keep the CI layer intact but just point ProgramTest to a remote example source instead of a local one, is that not working as easily as it seems it should, or there's some advantages to having dedicated GHA context for these?

danielrbradley (commented):
Really like this for getting additional feedback ahead of merging 👍 

Assume we wouldn't require the example tests for merging to avoid blocking on external issues?

thomas11:
> Assume we wouldn't require the example tests for merging to avoid blocking on external issues?

That's my thinking as well.

PR #3076, opened 2024-02-11
Distinguish nil from empty map as default state in resource discovery

thomas11:
This PR fixes what I believe is a bug introduced in #2456. For the default state of a resource, we need to distinguish between `nil` (no default) and the empty object as the empty default. This is also described in the doc comment of `AzureAPIResource.DefaultBody` in resources.go:
> - omitempty is not set to distinct between nil (no default) and empty maps (empty default)

`AzureClient.CanCreate` checks `DefaultBody != nil` to detect resources that are automatically created with their parent.

The actual change in metadata is not visible in the PR because we don't commit it. It looks like this:
```diff
--- ../m-master.json    2024-02-12 20:22:54.578408624 +0100
+++ ../m.json   2024-02-12 20:21:14.881957652 +0100
@@ -1005922,7 +1005922,7 @@
           ]
         }
       },
-      "defaultBody": {},
+      "defaultBody": null,
       "putAsyncStyle": "azure-async-operation",
       "deleteAsyncStyle": "azure-async-operation"
     },
```

### DIFF
--- a/provider/pkg/openapi/discover_test.go
+++ b/provider/pkg/openapi/discover_test.go
@@ -88,10 +88,8 @@ func TestDefaultState(t *testing.T) {
 		}
 	}
 
-	t.Run("No default", func(t *testing.T) {
-		path := "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Insights/diagnosticSettingsCategories/{categoryName}"
+	parseSwagger := func(t *testing.T, path string) *ResourceSpec {
 		swagger := makeSwagger(path)
-
 		version := NewVersionResources()
 
 		addResourcesAndInvokes(version, "/file/path", path, "insights", &swagger)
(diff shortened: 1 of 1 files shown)


provider/pkg/openapi/discover_test.go
```diff
@@ -47,3 +48,73 @@ func TestListDiagnosticCategoriesShouldBeAdded(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		}
+	}
+
+	t.Run("No default", func(t *testing.T) {
+		path := "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Insights/diagnosticSettingsCategories/{categoryName}"
+		swagger := makeSwagger(path)
```
rquitales, line 93: Should we also have the following two lines in this test as well to make it more explicit the default is nil?
```
def := defaults.GetDefaultResourceState(path)
require.Nil(t, def)
```

Perhaps we could refactor the 2 test cases into a table test?

thomas11: Agree on the extra assertion. Also agree that the test is repetitive, but a table test didn't save any LOC for me because of some if/else branching whether nil is expected. I've factored out the common code into a helper method now, which I think leaves nice concise test cases.

rquitales: Agreed. A table test may be unnecessary in this case!

rquitales (approved):

PR #3079, opened 2024-02-12
Add a custom resource for blob container legal holds

### LINKED ISSUE #2840: Support for Legal Hold for Azure Native on Storage Accounts
## Hello!

## Issue details

Currently if we need to create a legal hold for a container instead of a time based retention, Azure native doesn't support this.

### Affected area/feature
Azure Native Storage Account immutable polices specifically legal hold.

 the Pulumi Service

thomas11:
Azure Storage supports [Legal holds for immutable blob data](https://learn.microsoft.com/en-us/azure/storage/blobs/immutable-legal-hold-overview). The Azure Native provider didn't include this functionality because it's based on a POST API. Due to the heterogenous design of POST APIs, we cannot include them automatically.

This PR adds a custom resource adding legal holds. It makes use of the new `AzureClient` from #3062 to make POST requests directly using the built-in provider functionality, without using yet another Azure SDK.

An e2e test is also added.

Resolves #2840

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -221,6 +221,7 @@ jobs:
       # This is essentially just copying files from bin to the provider folder
       - name: Prebuild provider prerequisites
         run: |
+          make prebuild
           make --touch codegen schema
           make provider_prebuild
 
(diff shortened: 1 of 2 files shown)


provider/pkg/resources/customresources/custom_blob_container_legal_hold_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+		inputs := resource.PropertyMap{
+			resource.PropertyKey("tags"): resource.NewArrayProperty([]resource.PropertyValue{
+				resource.NewStringProperty("tag1"),
+			}),
+		}
+		_, err := custom.Create(context.Background(), containerId+"/legalHold", inputs)
+		assert.NoError(t, err)
+
+		assert.Len(t, m.postIds, 1)
```
rquitales, line 58: nit: should we use the `github.com/stretchr/testify/require`package instead since the following checks requires/depends on `m.postIds` to not be empty?

thomas11: good point

danielrbradley (approved):
Approved with comments inline

PR #3078, opened 2024-02-12
Re-enable the go-aks e2e test and extend it with a K8s upgrade

thomas11:
This PR re-enables the go-aks since, based on my local runs, it seems to work fine now. It also modernizes its dependencies and K8s version.

In addition, it's extended to also do a K8s upgrade after the initial deployment, based on #2094 which therefore seems fixed now (the report was based on v1 of the provider and a different API version).

It does seem like the tests's run time could be a problem - to be discussed.

Resolves #304

PR #3095, opened 2024-02-20
Reorganize examples integration tests into long and short based on run times

thomas11:
This PR was motivated by #3078 because this new test has a run time of ~24m, too long for PRs. That motivated me to collect data on the run times of all our e2e tests. Until now, about half of the tests were classified as "long", but with no clear pattern.

The data shows that two thirds of the tests take 7 minutes or less, so we picked that as a cut-off. Those tests are now classified as "short". PRs will now only run the short tests, leading to faster builds with still two thirds of the previous test coverage.

### DIFF
--- a/.github/workflows/acceptance-test.yml
+++ b/.github/workflows/acceptance-test.yml
@@ -52,7 +52,7 @@ jobs:
     with:
       ref: ${{ github.ref }}
       version: ${{ inputs.version || needs.version.outputs.version }}
-      short_test: ${{ inputs.short_test || false }}
+      short_test: ${{ (github.event_name == 'pull_request' || inputs.short_test) || false }}
       retention_days: 30
     secrets: inherit
 
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -145,7 +145,7 @@ jobs:
           repo: GoTestTools/gotestfmt
 
       - name: Run tests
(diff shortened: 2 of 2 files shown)


.github/workflows/acceptance-test.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     with:
       ref: ${{ github.ref }}
       version: ${{ inputs.version || needs.version.outputs.version }}
-      short_test: ${{ inputs.short_test || false }}
+      short_test: ${{ (github.event_name == 'pull_request' || inputs.short_test) || false }}
```
danielrbradley, line 55: I think this can only be reached via the workflow call or the pull request triggers. The workflow call should always set the `inputs.short_test`, so I think we can just change the fallback:

```suggestion
      short_test: ${{ (inputs.short_test) || true }}
```

.github/workflows/build-test.yml
```diff
@@ -145,7 +145,7 @@ jobs:
           repo: GoTestTools/gotestfmt
 
       - name: Run tests
-        if: "!inputs.short_test"
+        if: ${{ ! inputs.short_test }}
```
danielrbradley, line 148: Ah that's a neater way of managing the leading `!`!

danielrbradley (approved):

PR #3102, opened 2024-02-22
Skip Microsoft.Network 2023-03-01-preview when generating the schema

### LINKED ISSUE #3100: Workflow failure: nightly-sdk-generation
## Workflow Failure

- [2024-02-22T04:44:44.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/7999536913)

thomas11:
[networkManagerRoutingConfiguration.json](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/Microsoft.Network/preview/2023-03-01-preview/networkManagerRoutingConfiguration.json) fails parsing with
  `json: cannot unmarshal bool into Go struct field SwaggerProps.definitions of type []string`

This change also moves the exclusion processing to before we read the JSON, since that's where this failure is.

I was not able to see, so far, what's actually wrong with the file. It's valid JSON but not what github.com/go-openapi/spec expects. 

Fixes #3100

### DIFF
--- a/provider/cmd/pulumi-gen-azure-native/main.go
+++ b/provider/cmd/pulumi-gen-azure-native/main.go
@@ -77,6 +77,9 @@ func main() {
 	switch languages {
 	case "schema":
 		buildSchemaResult, buildSchemaErr := versioning.BuildSchema(buildSchemaArgs)
+		if buildSchemaErr != nil {
+			panic(buildSchemaErr)
+		}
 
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
Looks good.

It appears that https://github.com/Azure/azure-rest-api-specs/issues/27758 introduced an invalid schema with `"required": true` where `required` should be a string array.

PR #3108, opened 2024-02-26
Factor CRUD handling into a client that can be independently tested and used in custom resources

thomas11:
This PR is a continuation of the idea of #3062, which is to factor the provider, especially the huge `provider.go`, more cleanly.

Besides the standard benefits of encapsulation, testability, etc., this is also opens up the opportunity to pass the newly extracted client into custom resource implementations. This essentially enables them to call back into the standard CRUD logic, making their implementation similar to the standard resources while allowing variation, rather than the current completely separate implementation. #3079 showed this by passing the `AzureClient` into the custom resources. The new CRUD client in this PR could be used in custom resources in a similar manner, allowing them to use the standard provider CRUD building blocks like type lookup, REST body and query generation, SDK shape converstion etc.

Unfortunately, I don't have a custom resource using this new client to show at this point. Originally, I assumed #3079 might, but it turned out to be unnecessary beyond the existing `AzureClient`.

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -883,7 +883,7 @@ func (k *azureNativeProvider) Create(ctx context.Context, req *rpc.CreateRequest
 			return nil, azure.AzureError(err)
 		}
 	default:
-		outputs, err = k.defaultCreate(ctx, req, inputs, id, queryParams, crudClient)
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):
Nice :shipit: 

This also looks like an excellent step in the direction towards getting better test coverage of both the client methods, but also the core provider behaviour by having the option to mock out the crud interface.

PR #3111, opened 2024-02-27
Use v4 consitently for upload/download-artifact

mjeffryes:
the `download-artifact@v3` action can't pull artifacts uploaded by `upload-artifact@v4` (and vice versa). It looks like we updated some but not all of our builds to use v4 of these actions, which is causing master builds to fail. This change moves everything to v4 which should fix the builds. 

Fixes: https://github.com/pulumi/pulumi-azure-native/issues/3109


thomas11 (approved):
Oh that was probably me. Thanks for the fix!

PR #3123, opened 2024-03-04
Portal Dashboard: Treat MarkdownPartMetadata as untyped JSON

### LINKED ISSUE #1696: portal.dashboard only works with Extension/HubsExtension/PartType/MarkdownPart
### What happened?

Currently, Pulumi only support part type `Extension/HubsExtension/PartType/MarkdownPart`. 
Other types such as:
- `Extension/HubsExtension/PartType/MonitorChartPart`
- `Extension/HubsExtension/PartType/ClockPart`
- others metadata part types can be found [here](https://docs.microsoft.com/en-us/azure/azure-portal/azure-portal-dashboards-structure)

are not supported at the moment.

image
**This is the original dashboard**

image
It shows undefined for unsupported part types in Pulumi when do a getoutput.

May I know is Pulumi has any plan to fix this issue ? because the dashboard may not be useful if only MarkdownPart is supported.

### Steps to reproduce

Create a new dashboard using Pulumi portal dashboard api with part types as below:
- `Extension/HubsExtension/PartType/MonitorChartPart`
- `Extension/HubsExtension/PartType/ClockPart`
- others metadata part types can be found [here](https://docs.microsoft.com/en-us/azure/azure-portal/azure-portal-dashboards-structure)

### Expected Behavior

The part types should be supported and show in the azure portal dashboard

### Actual Behavior

The part types are not supported throwing errors

### Versions used

_No response_

### Additional context

_No response_

### Contributing

### LINKED ISSUE #3023: Dashboard - Treat DashboardPartsArgs as untyped JSON.
The Dashboard parts schema is incomplete in the azure specs. It looks like the type field is supposed to be a discriminator on a union type, but only the MarkdownPart arm of the union is defined in the specs. (Additional discussion on this problem can be found in #3015.)

While we've filed an upstream issue, this seems like a type of resource where it maybe difficult for azure to keep a fully typed schema up to date. As noted in https://github.com/pulumi/pulumi-azure-native/issues/3015, the terraform provider for azure just exposes the dashboard parts as an untyped json blob. 

We propose to add support for Dashboard Parts using a similar approach, by enhancing the generator to support marking a property or sub-property as JSON typed. (This would also address https://github.com/pulumi/pulumi-azure-native/issues/68.)

--------------

# Original: Dashboard - DashboardPartsArgs doesn't contain properties needed for successful Parts creation

### What is happening:
I'd like to deploy Azure Dashboard with multiple Parts.
For example this one (Server request rate MonitorChart): [DashboardPart.json](https://github.com/pulumi/pulumi-azure-native/files/13952172/DashboardPart.json)
Such part must contain following Metadata properties:
```cs
{
  Inputs,
  Settings,
  Type
}
```

`Inputs` Pulumi type is `InputList<object>` which does not match expected structure (refer to attached [DashboardPart.json](https://github.com/pulumi/pulumi-azure-native/files/13952172/DashboardPart.json))

`Settings` (`MarkdownPartMetadataSettingsArgs`) had completely different structure then needed.

`Type` has following comment:
```cs
    /// <summary>
    /// The type of dashboard part.
    /// Expected value is 'Extension/HubsExtension/PartType/MarkdownPart'.
    /// </summary>
    Input<string> Type { get;set; }= null!;
```

So it's not possible to deploy any complex DashboardPart using Pulumi.

### Example

Please compare Pulumi `DashboardPartsArgs` class structure with structure retrieved from existing Azure Dashboard ([DashboardPart.json](https://github.com/pulumi/pulumi-azure-native/files/13952172/DashboardPart.json))

### Output of `pulumi about`

CLI
Version      3.100.0
Go Version   go1.21.5
Go Compiler  gc

Plugins
NAME          VERSION
azure-native  2.24.0
dotnet        unknown
random        4.15.0

Host
OS       Microsoft Windows 10 Pro
Version  10.0.19045 Build 19045
Arch     x86_64

This project is written in dotnet: executable='C:\Program Files\dotnet\dotnet.exe' version='8.0.101'

Backend
Name           pulumi.com
URL            https://app.pulumi.com/ttvrdon
User           ttvrdon
Organizations  ttvrdon, resideo
Token type     personal

Dependencies:
NAME                                                  VERSION
Beacon.BuildSupport.MSBuild.Containers.RecordCreated  1.0.4
Beacon.BuildSupport.MSBuild.Pulumi.Publish            1.0.4
JetBrains.Annotations                                 2023.3.0
Pulumi                                                3.59.0
Pulumi.AzureNative                                    2.24.0
Pulumi.Random                                         4.15.0

### Additional context

_No response_

### Contributing

mikhailshilkov:
### Problem

As explained in #1696 and #3023, the `azure-native:portal:Dashboard` resource is currently ususable due to lack of specific types for dashboard parts. The API supports a multitude of part types, but the open API spec only describes one of them. The result of this is that we generate a discriminated union which is so restricted that it can't be used in practice.

We opened https://github.com/Azure/azure-rest-api-specs/issues/27465 upstream but received no feedback yet. The work to describe all other dashboard part types looks very involved - there are many types with many-many properties. Unfortunately, we don't anticipate reliable type definitions any time soon.

### Proposal

As suggested in #3023, this PR introduces a type override to remove the only strongly typed part definition and replace it with a generic DashboardMetadataPart type. The type accepts arbitrary collections of inputs and settings, which enables passing configuration for any part type.

Note that technically this is a breaking change. However, I'm quite convinced that the previous type system prevented anyone from creating any non-trivial dashboards, so we are not breaking any practical user scenarios.

### Implementation details

Historically, we implement this kind changes as a special case in our generation logic. However, this change is kind of tricky to do in the generation pass, because we need to introduce an new type, change an existing type, and then (ideally) remove unused types too.

Because of that, and to avoid further bloat of inline generation logic, I decided to introduce a new mechanism in customer resources. It allows specifying "overrides" for schema and metadata types, that would then replace the original types with the same name.

Additionally, I added a final-pass on the schema that deletes all unused object types. The good news is that no previous types were unused, so only the Dashboard Markdown types are being removed now. This also helps with the existing snapshot test: it elides the Dashboard types and therefore the snapshot is unchanged.

### Testing

In addition to a few unit tests, I added two end-2-end tests that create a dashboard from TypeScript and C#. The TypeScript one was created by importing an elaborate dashboard created manually in Azure portal. Unfortunately, our C# import is broken beyond repair for untyped dictionaries, so I defined a much simpler dashboard manually.

Resolves #1696 
Resolves #3023


mikhailshilkov:
@danielrbradley All good questions - I had similar thoughts. See my answers below.

> we could apply the transformation for the types generated for just a specific resource

There is no such thing as types scoped to a single resource. Types are contained within a module, and multiple resources of that module can rely on overlapping types. If we wanted to get a list of all types that a given resource depends on, we'd need to traverse its properties recursively and build out a list. And yet, again, it would not be an exclusive list.

I actually think that we need cross-resource and cross-module transformations. We could use them for existing hard-coded generation tweaks like SubResource.ID expansion, User Assigned Identities shape, maintained sub-resource collection, etc.

> it's also not complete as it doesn't allow for transformation of the resource's schema itself

That's straightforward to add later, I don't see why I need to do so in this PR without a use case.

> the order of transformations could be an issue as they could easily interact with each other.

That's fair. Do you have ideas how to overcome this? I could add a test that runs transformations in a few random orders and compares the result. Since we are talking about schema types here, if the order becomes important, we will start getting spurious diffs on schema generation, which is already a test of a kind.

> We're iterating through every generated type which I expect adds quite a bit of redundant processing

It's an iteration through map keys and parsing each key, which is ~O(N) for N in a few thousands. I expect it to take a few ms max. What am I missing?

> We're also using a hard-coded list to try and identify the types associated with the resource. If the schema is updated this will break.

That's fair. Do you have suggestions here? How can we transform the types without relying on their names? The best I can think of is to check the type shape and fail loud and clear if it's not what we expect, forcing a maintainer to resolve it.

If you have a better approach on these in mind, I'm all ears!

danielrbradley:
@mikhailshilkov I could see a few alternatives here...

1. For easier maintainability, we could just write the component completely manually - and skip the resource being generated at all. This would move us away from a half-way house of generating then kudging.
2. Another option would be skip the resource during the main generation, then when creating the custom resource call into the main generation code but from an isolated context so we get a complete list of the types related to just this one resource, then customise this. Then we could merge the resource & associated types into the main schema.
3. Currently our context during generation is for the whole `packageGenerator`, but we could create a new context per resource being generated something like `resourceGenerator` wherein we could load transformations for that specific resource (so resource transformations are registered before generation, keyed by resource) then these transformations would be called while generating all types associated to that resource.
4. A larger change would be to refactor the generation process to avoid mutating any global context - where every resource returns its own resource and associated types, then we use a process to merge each resources types after. This would allow for very simpler intercepting of the values being returned rather than intercepting while they're being generated (just before they're added to the global list).

danielrbradley (commented):
I think this is better, though it still feels a little obscure when reading the custom resource definition as to what it's actually doing. 

The separation of Types* and MetaTypes* is not great as these are almost identical implementations, and could easily introduce bugs if they differed accidentally. It's probably a wider issue to bring these two together into a single model at some point which can be projected into both the schema & metadata as required.

Thinking from the point of view of what we're trying to express in the purpose of this custom resource, perhaps utilizing the idea of the visitor pattern could work nicely here so the custom resource can only interact with types that the original resource referenced, or add new ones. For example:

```go
func portalDashboard() *CustomResource {
	return &CustomResource{
		// Provide a token of the resource to override
		Token: "azure-native:portal:Dashboard",
		// Follow the types referenced from the resource then pass them all in here to be customised
		CustomiseTypes: func (schema map[string]*schema.ComplexTypeSpec, metadata map[string]*resources.AzureAPIType) (map[string]*schema.ComplexTypeSpec, map[string]*resources.AzureAPIType) {
			// Iterate over types & add to output with modifications where required.
			// Extra types can be added here too.
			// Unreferenced types will be dropped from the schema.
		},
		// We could also add a `CustomiseResource` function eaily to mirror this for the root resource's schema too.
	}
}
```

danielrbradley (approved):
Revisiting this now ... let's get it merged.

I'm happy with the public interface change we're making which is the most important part. We can always refactor the internal implementation again later.

PR #3127, opened 2024-03-06
Bugfix for PartialMap.Get: found=false

thomas11:
I came across this trying to use `PartialMap.Get` and noticed that in the current implementation, the second return value is always `err == nil`, which is redundant, instead of indicating whether the key was found.


danielrbradley (approved):

thomas11:
> It would be sweet to add some docs and tests that would define the semantics of the results. Don't block on this, just an observation.

I totally agree but don't have the bandwidth ATM.

Issue #903, opened 2021-06-07
Retry "AnotherOperationInProgress" error messages


sam-cogan:
## Affected feature

Currently if Azure generates a "AnotherOperationInProgress" error Pulumi will fail. This error is often caused by trying to access two sub-resources on a resource at the same time. A good example is creating multiple subnets on a virtual network. Pulumi will try and create them all that the same time, but Azure can only create them one at a time.

Adding a back-off and retry for the "AnotherOperationInProgress" error message would help workaround this problem

infin8x:
@sam-cogan we'll take a look at this. In the meantime, if you create a `dependsOn` chain with these sub-resources, Pulumi won't try to create them all at the same time. See https://www.pulumi.com/docs/intro/concepts/resources/#dependson for details.

sam-cogan:
@infin8x thanks, I have done that and it does indeed work around the issue.

masashi-shib:
> @sam-cogan we'll take a look at this. In the meantime, if you create a `dependsOn` chain with these sub-resources, Pulumi won't try to create them all at the same time. See https://www.pulumi.com/docs/intro/concepts/resources/#dependson for details.

While this workaround works perfectly the chaining of dependencies forces the developer to create the whole VNET and its Subnets in one module which is not optimal especially when trying to split code into different modules.

One example might be that the code for a specific module should create the Subnet in the VNET add the corresponding resources and configure firewall rules.

masashi-shib:
@mikhailshilkov  @infin8x I would also love to hear what is the impact of deleting the first subnet only (e.g. replacing its name) after the chaining is done.

e.g. 
Subnet 1 <--depends_on-- Subnet2 <--depends_on-- Subnet 3

What happens if in the case the Subnet 1 name has been changed?
Does it create a havox delete basically on all resources that have an assignment in any of the subnets?

mikhailshilkov:
> Does it create a havox delete basically on all resources that have an assignment in any of the subnets?

I think the answer is yes, unfortunately. In that case you'd have to remove all the `dependsOn` annotations and run an update before trying to rename the first subnet. I realise it's not ideal and we should solve the original problem.

deepak-goel:
Hi team, are there any plans to fix this? Terraform also had the same kind of race condition, but they already fixed that a while back.

MatthewBettonMaif:
Hello Team, 
[...]

PR #3128, opened 2024-03-06
Initial sketch of using azcore client

mikhailshilkov:
I haven't wired it anywhere, so it's no functional change, but I was playing with `azcore` to see if it solves #903 (it does not) and implemented the minimal version of CRUD. It worked for resource groups/vnets/subnets, but obviously lots of work to do to make it universal and then gradually introduce it to users.

First step towards #1464

@thomas11 @danielrbradley @mjeffryes Do you folks think it's worth pursuing?

### DIFF
--- a/provider/pkg/azure/client_azcore.go
+++ b/provider/pkg/azure/client_azcore.go
@@ -1,0 +1,161 @@
+// Copyright 2016-2024, Pulumi Corporation.
+
+package azure
+
+import (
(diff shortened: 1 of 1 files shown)


thomas11:
Nice! I think it's definitely worth pursuing since we have to do #1464 at some point. The question is, as always, prioritization, i.e., when to do it.

Issue #2455, opened 2023-05-12
Add AzureRM Privileged Identity Management to Provider


phantlantis:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Hello! Looking through the available resources, one set that is missing that could be majorly helpful is the Privileged Identity Management endpoint (the ones specifically for ARM, _not_ the ones for Graph).

These currently sit under the Microsoft.Authorization namespace - the ones I'm specifically hoping for are:
`https://management.azure.com/{scope}/providers/Microsoft.Authorization/roleEligibilityScheduleRequests/{roleEligibilityScheduleRequestName}` as documented [here](https://learn.microsoft.com/en-us/rest/api/authorization/privileged-role-eligibility-rest-sample).

Furthermore, the role management policies that govern PIM activation on each resource scope:
`https://management.azure.com/{scope}/providers/Microsoft.Authorization/roleManagementPolicies/{roleManagementPolicyId}?api-version=2020-10-01` - as documented [here](https://learn.microsoft.com/en-us/rest/api/authorization/privileged-role-policy-rest-sample)

### Affected area/feature

Likely the authorization section of the SDKs as a new resource. 

Thanks!

kpitzen:
Hi @phantlantis - thank you for bringing this to our attention!  This is due to Azure's API specs neither defining an endpoint to `DELETE` this resource nor a reasonable default to set this resource to when Pulumi runs its delete procedures. What we'll want to do to add support here is to add these endpoints and a corresponding default state here:
https://github.com/pulumi/pulumi-azure-native/blob/0b1942064cfcc9086225c91be8d22e0f1411adf2/provider/pkg/openapi/defaultResourcesState.go#L9

We'll get this queued up when we can, but if there are any enterprising souls out there, that would be the desired starting point.

Thank you again!

thomas11:
I just spent a couple of hours looking into this under-documented API and I’m afraid it’s a size L task.

I don’t think our existing "reset to default state" mechanism actually applies here.
  - `/roleManagementPolicies` does have a DELETE method but is missing PUT (_"You do not need to Create role management policies as each role within each resource has a default policy"_)
  - `/roleEligibilityScheduleRequests` is missing DELETE but doesn’t have a default state to be reset to, instead requiring explicit revocation.

The way to support both would be through custom resources.

[...]

PR #3130, opened 2024-03-06
Support Authentication PIM Role Management Policies

thomas11:
## About

This PR adds support for Role Management Policies, part of [Privileged Identity Management (PIM)](https://learn.microsoft.com/en-us/entra/id-governance/privileged-identity-management/pim-configure) in the  Microsoft.Authentication namespace. It's one part of #2455.

Note that this is about the ARM part of PIM; there's also a Microsoft Graph API part which is not covered by this provider.

This resource wasn't automatically included because it supports only GET and PATCH. The policies are singletons that cannot be created or deleted, only modified via PATCH.

## Implementation

Role Management Policies essentially consist of a name which is actually a GUID, and a list of ~20 rules.

Using our existing singleton support `defaults.GetDefaultResourceState` was tricky because
1. there are many policies with many rules, for a total of ~300k lines of JSON for a subscription scope, and there are more scopes, plus
2. I believe the defaults can vary per scope and possibly also per customer.

So instead, I've implemented a custom resource that captures the original state of a policy when it's first "created", i.e., added to Pulumi state. When a rule or the whole policy is removed from Pulumi, we look up the original state and re-apply it.

## Testing

The e2e/integration test for this resource is special because using PIM requires a paid [Entra ID P2 license](https://www.microsoft.com/en-us/security/business/microsoft-entra-pricing). We have one that you can see [here](https://admin.microsoft.com/Adminportal/Home?referrer=entra#/licensedetailpage/84a661c4-e949-4bd2-a560-ed7766fcaf2b).

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -1385,24 +1385,19 @@ func (k *azureNativeProvider) Delete(ctx context.Context, req *rpc.DeleteRequest
 		}
 	case res.Singleton:
 		// Singleton resources can't be deleted (or created), set them to the default state.
-		for _, param := range res.PutParameters {
-			if defaults.SkipDeleteOperation(res.Path) {
-				continue
+		if body, ok := res.BodyParameter(); ok && !defaults.SkipDeleteOperation(res.Path) {
+			requestBody := k.converter.SdkInputsToRequestBody(body.Body.Properties, res.DefaultBody, id)
+
+			queryParams := map[string]interface{}{"api-version": res.APIVersion}
+
+			// Submit the `PUT` or `PATCH` against the ARM endpoint
+			op := k.azureClient.Put
+			if res.UpdateMethod == "PATCH" {
+				op = k.azureClient.Patch
 			}
-			if param.Location == "body" {
-				requestBody := k.converter.SdkInputsToRequestBody(param.Body.Properties, res.DefaultBody, id)
-
-				queryParams := map[string]interface{}{"api-version": res.APIVersion}
-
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/provider.go
```diff
@@ -833,7 +836,16 @@ func (k *azureNativeProvider) Create(ctx context.Context, req *rpc.CreateRequest
 	switch {
 	case isCustom && customRes.Create != nil:
 		// First check if the resource already exists - we want to try our best to avoid updating instead of creating here.
-		_, exists, err := customRes.Read(ctx, id, inputs)
+		var exists bool
+		if customRes.CanCreate != nil {
+			err = customRes.CanCreate(ctx, id)
+			exists = err != nil
```
t0yv0, line 842: Is not existing the only kind of err?

t0yv0:
Glanced over it but the review effort here is formidable, LMK if you need me as a reviewer, I'm happy to try again. Since there's a lot going on it can be helpful to rebase into step-by-step commits.

danielrbradley (commented):
What's the status on this PR? Sorry I didn't get to a review when you put this up. 

Guessing the behaviour change is still needed but we might need to rework the changes that touch the custom resources since more recent updates?

thomas11:
> What's the status on this PR? Sorry I didn't get to a review when you put this up.
> 
> Guessing the behaviour change is still needed but we might need to rework the changes that touch the custom resources since more recent updates?

Yep, exactly that.

thomas11:
superceded by #3663

PR #3154, opened 2024-03-14
Generate properties for primitive string outputs

### LINKED ISSUE #3147: Add a function for the getCustomDomainVerificationId API.
## Hello!

## Issue details

Add a function that executes this API to return the Custom Domain Verification ID.
https://learn.microsoft.com/en-us/rest/api/containerapps/get-custom-domain-verification-id/get-custom-domain-verification-id?view=rest-containerapps-2023-11-02-preview&viewFallbackFrom=rest-containerapps-2023-08-01-preview&tabs=HTTP

Although the verification ID is available as an output from a containerApp resource, it is sometimes needed before creating a containerApp.

### Affected area/feature

From speaking to @thomas11, it appears the function is not being created today due to the fact that it’s missing a response spec. In schema.go, `genFunctions`
```go
	if response == nil || len(response.specs) == 0 {
		// Response is specified empty, do not generate an invoke for it.
		return
	}
```

thomas11:
This is a quick attempt to fix #3147 and include Invokes with primitive string outputs. It's not sufficient, however.

The request part of the invoke seems to work, but the response handling does not. The provider assumes that all POST responses can be unmarshalled into a `map[string]any`, but for this invoke, it's simply a primitive string. Presumably, we could extend metadata to indicate this to unmarshall it correctly at run time.

The change also adds a number of properties to existing resources that I have not verified.

I've used this program to test:
```ts
import * as app from "@pulumi/azure-native/app";

const id = app.getCustomDomainVerificationId();
export const customDomainVerificationId = id;
```

### DIFF
--- a/provider/pkg/azure/client.go
+++ b/provider/pkg/azure/client.go
@@ -6,12 +6,10 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"reflect"
 	"strings"
 
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
-	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/logging"
 )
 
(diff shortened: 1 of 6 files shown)


provider/pkg/gen/properties.go
```diff
@@ -216,6 +216,21 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, isOutput
 			result.requiredProperties.Add(name)
 		}
 	}
+
+	// If the schema has no properties but is a primitive output, we include it as a property so invokes can be generated.
+	if len(names) == 0 && len(result.specs) == 0 && isOutput && len(resolvedSchema.Type) == 1 && resolvedSchema.Type[0] == "string" {
```
danielrbradley, line 221: Could we generalise the specific type to any primative?

sdk/nodejs/azurestackhci/update.ts
```diff
@@ -40,6 +40,8 @@ export class Update extends pulumi.CustomResource {
         return obj['__pulumiType'] === Update.__pulumiType;
     }
 
+    public /*out*/ readonly HealthState!: pulumi.Output<string | undefined>;
+    public /*out*/ readonly RebootRequirement!: pulumi.Output<string | undefined>;
```
danielrbradley, line 44: Are these changes to resources related to the schema gen change somehow or being caused by something else?

danielrbradley (commented):
This looks like a great addition. Would just like to work out where the changes to existing resources is coming from but then this looks like a great gain 👍

thomas11:
> This looks like a great addition. Would just like to work out where the changes to existing resources is coming from but then this looks like a great gain 👍

Thanks but it panics at run time :) When handling the Invoke response, as described above. Need to fix that.

sdk/nodejs/azurestackhci/update.ts
thomas11: These changes are related. I verified by stashing my change and regenerating, which doesn't have them. I haven't investigated yet.

thomas11:
superceded by #3165

PR #3165, opened 2024-03-21
Support invokes with non-object outputs, via wrapping

### LINKED ISSUE #3147: Add a function for the getCustomDomainVerificationId API.

thomas11:
- **Add invokes that return non-object types such as plain strings**
- **AzureClient.Get and .Post can deserialize any response, not just objects**
- **regenerate**

Resolves #3147

If we're generating response properties and the resolved schema has no properties but is a primitive/array output, we include it as a special property so invokes can be generated.

Ideally we'd just represent the response as a primitive type without the wrapping magic property, using  `FunctionSpec.ReturnType`, but pulumi/pulumi#15738 and pulumi/pulumi#15739 prevent this.
    
Unfortunately, this change required adding a third boolean to `genProperties` to limit this new case to function responses only. Happy to incorporate suggestions to model this better.

This is a simple program that demonstrates one of the new invokes.
```ts
import * as app from "@pulumi/azure-native/app";
const id = app.getCustomDomainVerificationIdOutput();
export const customDomainVerificationId = id;
export const customDomainVerificationIdValue = id.value;
```

Output:
```
  + customDomainVerificationId     : {
      + value: "B11..."
    }
  + customDomainVerificationIdValue: "B11..."
```

### DIFF
--- a/provider/pkg/provider/crud/crud.go
+++ b/provider/pkg/provider/crud/crud.go
@@ -354,8 +354,11 @@ func (r *resourceCrudClient) Read(ctx context.Context, id string) (map[string]an
 	if err != nil {
 		return nil, err
 	}
-	// Cast is safe because we're reading a resource, only invokes return non-objects.
-	return resp.(map[string]any), nil
+	// Cast should be safe because we're reading a resource, only invokes return non-objects.
+	if respMap, ok := resp.(map[string]any); ok {
+		return respMap, nil
+	}
+	return nil, errors.Errorf("expected object from Read of '%s', got %T", url, resp)
 }
 
 type propertyPath struct {


provider/pkg/gen/properties.go
```diff
@@ -43,7 +43,7 @@ type propertyBag struct {
 
 type RequiredContainers [][]string
 
-func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, isOutput, isType bool) (*propertyBag, error) {
+func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, isOutput, isType, isResponse bool) (*propertyBag, error) {
```
danielrbradley, line 46: A couple of thoughts on the growing list of bools .. one was fine, two was passable, three is a little rough.

Option 1: We could just group them into a `getPropertyOpts` struct. This would make the call sites a little more obvious without having to rember all the comments on what they are. This would also allow us to describe in the docs string what the pupose/effect of each is.

Option 2: Wrap the `moduleGenerator` context into something like `propertiesGenerator` which includes these contextual details.

Side note: it's a little odd that other other two options just get passed recursively without being modifed, but this new property doesn't get propegated beyond the initial call, so perhaps option 1 makes more sense here.

provider/pkg/provider/provider.go
```diff
@@ -351,13 +351,10 @@ func (k *azureNativeProvider) Invoke(ctx context.Context, req *rpc.InvokeRequest
 		} else {
 			return nil, errors.Errorf("neither GET nor POST is defined for %s", label)
 		}
-
 		if err != nil {
 			return nil, errors.Wrapf(err, "request failed %v", id)
 		}
-
-		// Map the raw response to the shape of outputs that the SDKs expect.
-		outputs = k.converter.ResponseBodyToSdkOutputs(res.Response, response)
```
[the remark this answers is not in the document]
thomas11: `res.Response` is metadata about the response shape. `response` is the actual run time response. It's just that the order of arguments is reversed between `k.converter.ResponseBodyToSdkOutputs` and `k.invokeResponseToOutputs`.

provider/pkg/gen/properties.go
thomas11: Addressed in the latest commit.

danielrbradley (approved):

PR #3180, opened 2024-03-27
Remove defaults for useOidc and useMsi so they don't take precedence over env variables

### LINKED ISSUE #3179: Configuration flags `useMsi` and `useOidc` are optional but default to `false`, overriding the environment variables
## Scenario
- The environment has `ARM_USE_OIDC=true`.
- An explicit provider is created without setting the `useOidc` parameter.

## Expectation
OIDC authentication is enabled and being used (provided the other necessary configurations are set), due to `ARM_USE_OIDC`.

## Actual
OIDC authentication is not used. 

Verbose logs show that when the explicit provider is registered with the engine, `useOidc=false` is passed. This takes precedence over the environment variable.

As a workaround, users can explicitly pass `useOidc=true` when instantiating the provider.

All of the above is also true for `ARM_USE_MSI`/`useMsi`.

thomas11:
Resolves #3179 

See the issue for details.


danielrbradley (approved):
SDK-applied default values are a pain. Assuming the provider also implements the same defaults in Configure, then this looks good 👍

danielrbradley:
Provider defaults set here:
https://github.com/pulumi/pulumi-azure-native/blob/d882dd8ec8091f03d6ded8e7770e7c8c991b7fb2/provider/pkg/provider/auth.go#L84-L88

PR #3181, opened 2024-03-27
Move actionlint to own workflow

danielrbradley:
Only run actionlint workflow for PRs if something in the `.github` folder has been modified.


danielrbradley:
> I assume this requires updating the branch protection rules once it's merged?

I don't think we actually enforce the actionlint check currently because of these issues with large PRs. We could add it if we want though, or just treat it as informational for the time being.

PR #3185, opened 2024-03-28
Bump actions/setup-* to switch from Node 16 to 20, plus Go and Node

thomas11:
**Bump actions/setup-foo to switch from Node 16 to 20**
This gets rid of these warnings in CI:
> [build_test / Build binaries and schema](https://github.com/pulumi/pulumi-azure-> native/actions/runs/8464175900/job/23188371782)
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: ...

**Bump Go to 1.22 and Node to 20 in CI**
Those are the current releases resp. LTS versions.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -12,7 +12,7 @@ runs:
     - name: Install Go
       uses: actions/setup-go@v5
       with:
-        go-version: "1.21.x"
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):
Nice :shipit:

PR #3188, opened 2024-03-28
Fix flaky TestFindOlderVersions due to gofakeit noun collision

thomas11:


provider/pkg/versioning/findOlderVersions.go
```diff
@@ -18,6 +19,9 @@ func findOlderVersions(specVersions ProvidersVersionResources, defaultVersion op
 				continue
 			}
 			olderVersions = append(olderVersions, version)
+			if len(olderVersions) > 1 {
+				log.Println("whaaa")
+			}
```
danielrbradley, line 24: Think this needs to be removed! 😆

provider/pkg/versioning/findOlderVersions_test.go
```diff
@@ -54,7 +54,7 @@ func fakeApiVersion(spec FakeApiVersion) openapi.ApiVersion {
 		min = min.AddDate(0, 0, 1)
 	}
 	if spec.LessThan != "" {
-		max, _ = time.Parse(ApiVersionLayout, spec.GreaterThan)
```
danielrbradley, line 57: Ah nice catch 🤦

danielrbradley (approved):
Looks good once tidied up

Issue #3208, opened 2024-04-10
Pulumi.AzureNative.AppConfiguration.KeyValue failing to create


holtsoftware:
### What happened?

When i go to create a KeyValue object like using AzureNative in C# it will not add the KeyValue to the Azure App configuration instead it errors out with a very generic error `error: autorest/azure: error response cannot be parsed: {"" '\x00' '\x00'} error: EOF` I can provide the same information to AzureClassic and it will create the Key in Azure App Configuration without issue.

### Example

```csharp
new Pulumi.AzureNative.AppConfiguration.KeyValue($"global-{global.Key}", new()
{
    ConfigStoreName = "<The real name of the Config Store>",
    ResourceGroupName = "<The real name of the config store's resource group>",
    KeyValueName = "RootUris:Category:Endpoint$dev/Global",
    Value = "https://someurl.com/",
});
```

when I look at the details of the up it looks like this

<pre>
+ azure-native:appconfiguration:KeyValue: (create)
        [urn=urn:pulumi:dev::...::azure-native:appconfiguration:KeyValue::global-RootUris__Category__Endpoint]
        [provider=urn:pulumi:dev::...::pulumi:providers:azure-native::default_2_36_0::b49d6206-cd02-42d0-828f-be0d0ae7354f]
        configStoreName  : "<The real name of the Config Store>"
        keyValueName     : "RootUris:Category:Endpoint$dev/Global"
        resourceGroupName: "<The real name of the config stores resource group>"
        value            : "https://someurl.com/"
</pre>

After I tell it to up I get this error back
`error: autorest/azure: error response cannot be parsed: {"" '\x00' '\x00'} error: EOF`

### Output of `pulumi about`

CLI
Version      3.112.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
NAME    VERSION
dotnet  unknown

Host
OS       Microsoft Windows 11 Enterprise
Version  10.0.22631 Build 22631
Arch     x86_64

This project is written in dotnet: executable='C:\Program Files\dotnet\dotnet.exe' version='8.0.202'

Current Stack: organization/*/dev

TYPE                           URN
pulumi:pulumi:Stack            urn:pulumi:dev::...::pulumi:pulumi:Stack::...-dev
pulumi:providers:azure-native  urn:pulumi:dev::...::pulumi:providers:azure-native::default_2_36_0

Found no pending operations associated with dev

Backend
Name           aholt-LT
[...]

PR #3210, opened 2024-04-11
Handle non-JSON responses more gracefully

thomas11:
In #3208 we see a case where an Azure API responds with a non-JSON body in an error case. We don't expect that and run `autorest.ByUnmarshallingJSON`, causing a confusing deserialization error:
```
error: autorest/azure: error response cannot be parsed: {"" '\x00' '\x00'} error: EOF
```

After this change, the same program errors with
```
error: unexpected response from the service: '', HTTP status 404 Not Found
```


danielrbradley (approved):
Looks reasonable. Any easy way to unit test this given the new abstraction?

rquitales (approved):

PR #3218, opened 2024-04-17
Update and reactivate postgres test

### LINKED ISSUE #898: Cannot create Postgres Flexible Server Configuration [was: Postgres test is failing]
> The test started failing in master builds on June 3
> 
> `deleting error: Code="InvalidParameterValue" Message="Invalid value given for parameter configurationValue. Specify a valid parameter value."`
> 
> I'll disable the test to unblock the builds but we should investigate.

Update (@thomas11): I wasn't able to create a [Configuration](https://www.pulumi.com/registry/packages/azure-native/api-docs/dbforpostgresql/configuration/) after several attempts. Our REST payload looks exactly like in the docs, though. Details in comment.

thomas11:
I've manually run the nodejs tests twice in addition to the regular CI run.

- Update the postgres test so it builds. Limit to the current Flexible Server API., Single Server is long deprecated by Azure.
- Re-enable the test.

Resolves #898

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -66,7 +66,6 @@ func TestImportTs(t *testing.T) {
 }
 
 func TestPostgresTs(t *testing.T) {
-	t.Skip("disabled due to a failure, see https://github.com/pulumi/pulumi-azure-native/issues/898")
 	skipIfShort(t)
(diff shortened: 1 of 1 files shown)


mikhailshilkov (approved):
LGTM, let's give it a try

PR #3219, opened 2024-04-17
Handle read failure after create failure

### LINKED ISSUE #3200: `504 GatewayTimeout` Error while creating a ContainerApp caused Pulumi to fail and not add it to state despite the resource being created
### What happened?

While attempting to create a container app I encountered a `504 GatewayTimeout` where creating the container apps timed out at `800s` but they ended up actually being created in Azure. A subsequent `pulumi up` results in a resource already exists error.

This requires some manual intervention where I need to `import` the resource or alternatively, delete the resource in the portal (resulting in down time) and recreate the resource again. 

I would consider this to be a transient error and would prefer pulumi to be able to recover from a scenario like this. 

### Error Message

> `+  azure-native:app/v20230501:ContainerApp redacted **creating failed** error: resource partially created but read failed autorest/azure: Service returned an error. Status=504 Code="GatewayTimeout" Message="The gateway did not receive a response from 'Microsoft.App' within the specified time period.": autorest/azure: Service returned an error. Status=504 Code="GatewayTimeout" Message="The gateway did not receive a response from 'Microsoft.App' within the specified time period."`

### Example

<details><summary>
Container App Code
</summary>

```
using Pulumi;
using Pulumi.AzureNative.Resources;
using Pulumi.AzureNative.ContainerRegistry;
using Pulumi.AzureNative.ContainerRegistry.Inputs;
using Pulumi.AzureNative.App;
using Pulumi.AzureNative.App.Inputs;
using Pulumi.Docker;
using System.Collections.Generic;
using Pulumi.Docker.Inputs;
using DockerRegistryArgs = Pulumi.Docker.Inputs.RegistryArgs;
using System;

return await Pulumi.Deployment.RunAsync(async () =>
    {
        var resourceGroupName = "zbuchheitResourceGroup";
        var resourceGroup = new ResourceGroup("resourceGroup", new ResourceGroupArgs
        {
            ResourceGroupName = resourceGroupName,
            Location = "EastUS"
        });
        var config = new Pulumi.Config();
        var importing = config.GetBoolean("importing") ?? false;

        var registry = new Registry("registry", new Pulumi.AzureNative.ContainerRegistry.RegistryArgs
        {
            ResourceGroupName = resourceGroup.Name,
            Sku = new SkuArgs
            {
                Name = "basic"
            },
            AdminUserEnabled = true
        });

        var credentials = Output.Tuple(resourceGroup.Name, registry.Name).Apply(names =>
            ListRegistryCredentials.InvokeAsync(new ListRegistryCredentialsArgs
            {
                ResourceGroupName = names.Item1,
                RegistryName = names.Item2
            })
        );

        var username = credentials.Apply(creds => creds.Username);
        var password = credentials.Apply(creds => creds.Passwords[0].Value);

        var image = new Image("image", new ImageArgs
        {
            ImageName = Output.Format($"{registry.LoginServer}/timeout:latest"),
            Build = new DockerBuildArgs
            {
                Context = ".",
                Platform = "linux/amd64",
            },
            Registry = new DockerRegistryArgs
            {
                Server = registry.LoginServer,
                Username = username!,
                Password = password!
            }
        }, new CustomResourceOptions { IgnoreChanges = new List<string> { "build" } });

        var managedEnvironment = new ManagedEnvironment("myManagedEnvironment", new ManagedEnvironmentArgs
        {
            ResourceGroupName = resourceGroup.Name,
            Location = resourceGroup.Location
        });

        var containerAppName = "zbuchheitcontainerapp";
        var containerApp = new ContainerApp(containerAppName, new ContainerAppArgs
        {
            ContainerAppName = containerAppName,
            ResourceGroupName = resourceGroup.Name,
            Location = resourceGroup.Location,
            EnvironmentId = managedEnvironment.Id,
            ManagedEnvironmentId = managedEnvironment.Id,
            Configuration = new ConfigurationArgs
            {
                MaxInactiveRevisions = 100,
                Registries = new List<RegistryCredentialsArgs>
                {
                    new RegistryCredentialsArgs
                    {
                        Server = registry.LoginServer,
                        Username = username!,
                        PasswordSecretRef = "registry-pwd",
                    }
                },
                Secrets = new List<Pulumi.AzureNative.App.Inputs.SecretArgs>
                {
                    new Pulumi.AzureNative.App.Inputs.SecretArgs
                    {
                        Name = "registry-pwd",
                        Value = password!
                    },
                }
            },
            Template = new TemplateArgs
            {
                Scale = new ScaleArgs
                {
                    MaxReplicas = 10,
                },
                Containers = new List<Pulumi.AzureNative.App.Inputs.ContainerArgs>
                {
                    new Pulumi.AzureNative.App.Inputs.ContainerArgs
                    {
                        Name = "mycontainer",
                        Image = image.ImageName,
                    }
                }
            }
        }, new()
        {
            ImportId = importing
            ? GetContainerApp.InvokeAsync(new GetContainerAppArgs
            {
                ContainerAppName = containerAppName,
                ResourceGroupName = resourceGroupName
            }).Result.Id
            : null,
            // IgnoreChanges = new List<string> {"identity", "configuration.secrets[0]", "configuration.registries[0]", "template.revisionSuffix", "template.containers[0]" }
        });
    });

```
</details>

I don't have a consistent repro as this is a network related error.

### Output of `pulumi about`

```
CLI          
Version      3.112.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
NAME          VERSION
azure-native  2.35.0
command       0.9.2
docker        4.5.3
dotnet        unknown

Host     
OS       darwin
Version  14.2.1
Arch     arm64

This project is written in dotnet: executable='/usr/local/share/dotnet/dotnet' version='6.0.417'

Current Stack: zbuchheit-pulumi-corp/azure-cs/dev

Found no resources associated with dev

Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/zbuchheit-pulumi-corp
User           zbuchheit-pulumi-corp
Organizations  zbuchheit-pulumi-corp
Token type     personal

Dependencies:
NAME                VERSION
Pulumi              3.60.0
Pulumi.AzureNative  2.35.0
Pulumi.Command      0.9.2
Pulumi.Docker       4.5.3
```

### Additional context

In doing discovery in the logs, I see the workflow for creating this resource is as follows

`GET` to check if the resource exists, if not, do a `PUT` on the Azure endpoing to create the resource which will provide a `Azure-Asyncoperation` url to perform a `GET` call against and watch for the operation to complete. It then polls this until it successfully completes. 

I don't have a verbose copy of the logs to verify if the 504 comes from the `PUT` or the subsequent `GET` calls, but I wonder if like there is a possibility for adding some handling in a scenario like this where a 504 is encountered, another GET is fired off with a retry timer to check the resource, and if the resource doesn't exist after a reasonable threshold, bubble up the error.

Much like https://github.com/pulumi/pulumi-azure-native/issues/1266 this causes some friction in use of this provider and causes users to lose confidence in the provider.

### Contributing

danielrbradley:
If a read fails when trying to return a partial result, ensure we still return a partial result with the resource identifier so the operation can be completed using refresh at a later point.

Fixes #3200 - where the partial state failed to be written resulting in the resource being created but lost from state.

### DIFF
--- a/provider/pkg/provider/crud/crud.go
+++ b/provider/pkg/provider/crud/crud.go
@@ -201,7 +201,8 @@ func (r *resourceCrudClient) HandleErrorWithCheckpoint(ctx context.Context, err 
 	// Try reading its state by ID and return a partial error if succeeded.
 	checkpoint, getErr := r.currentResourceStateCheckpoint(ctx, id, inputs)
 	if getErr != nil {
-		return azure.AzureError(errors.Wrapf(err, "resource updated but read failed %s", getErr))
+		// If reading the state failed, combine the errors but still return the partial state.
+		err = azure.AzureError(errors.Wrapf(err, "resource created but read failed %s", getErr))
 	}
 	return partialError(id, err, checkpoint, properties)
 }


thomas11 (approved):

provider/pkg/provider/crud/crud.go
```diff
@@ -201,7 +201,8 @@ func (r *resourceCrudClient) HandleErrorWithCheckpoint(ctx context.Context, err
 	// Try reading its state by ID and return a partial error if succeeded.
 	checkpoint, getErr := r.currentResourceStateCheckpoint(ctx, id, inputs)
 	if getErr != nil {
-		return azure.AzureError(errors.Wrapf(err, "resource updated but read failed %s", getErr))
+		// If reading the state failed, combine the errors but still return the partial state.
+		err = azure.AzureError(errors.Wrapf(err, "resource created but read failed %s", getErr))
```
mikhailshilkov, line 205: This doesn't look right to me... If `getErr` is not nil, then `checkpoint` is nil and that's what we return in the `partialError`. Are we going to save empty state? Will this work at all? Can you please describe how you tested this change?

danielrbradley: Yup, tested this locally and it works fine. Checkpoint being nil is fine - just means we have no inputs or outputs recorded. Most importantly, we have an ID recorded which lets us run refresh if the resource was actually created correctly - otherwise it'll just show up as an update on the next deployment.

danielrbradley: Arguably, we could make this slightly better by still creating some kind of checkpoint object if the read fails - but perhaps missing just the final outputs.

mikhailshilkov: We could also get here from Update operation, right? (In which case I'm not sure that the change in the error message makes it better)
Writing an empty snapshot in this case looks even worse? What will Refresh behavior look like?

PR #3224, opened 2024-04-18
Implement provider cancellation

### LINKED ISSUE #1055: Cooperative cancellation of in-progress operations
Implement the [`Cancel`](https://github.com/pulumi/pulumi-azure-native/blob/a2a36970861bce67b6e11cd322b023c4a5b936b4/provider/pkg/provider/provider.go#L1201) operation in the provider so that users could Ctrl-C to cancel operations that are currently awaiting for the result.

Kubernetes provider has this implemented.

danielrbradley:
Attempt to gracefully shut down the provider if the engine indicates that we need to halt operations.

Fixes #1055

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -65,6 +65,9 @@ type azureNativeProvider struct {
 	rgLocationMap    map[string]string
 	lookupType       resources.TypeLookupFunc
 	skipReadOnUpdate bool
+
+	context  context.Context
+	shutdown context.CancelFunc
 }
 
 func makeProvider(host *provider.HostClient, name, version string, schemaBytes []byte,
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/provider.go
```diff
@@ -1007,6 +1016,8 @@ func (k *azureNativeProvider) findUnsetPropertiesToMaintain(res *resources.Azure
 
 // Read the current live state associated with a resource.
 func (k *azureNativeProvider) Read(ctx context.Context, req *rpc.ReadRequest) (*rpc.ReadResponse, error) {
+	// Use the global context to handle provider shutdown.
+	ctx = k.context
```
thomas11, line 1020: Is it really correct to disregard the passed-in context? Can you explain why?

danielrbradley: This is exactly what aws-native and kubernetes do. I think by the time the request gets cancelled, we're way past being able to actually do anything to shut down cleanly. This is why the cancellation exists externally.

My first pass of this was to do something fancy with combining the contexts to make it an OR between the two cancellations, but I think it's significantly harder to reason about in practice and this appears sufficient from other providers.

thomas11 (approved):

Issue #2937, opened 2023-12-07
Using Entra ID Only Auth and Updating causes a failure when Tag is incorrectly set


xenon8:
### What happened?

We have turned Entra ID  Only Auth on and then changed a Tag on the Azure SQL Server from the value that Pulumi sets.

This caused this error:

"AadOnlyAuthenticationIsEnabled" Message="Azure Active Directory Only Authentication is enabled. Please contact your system administrator."

Once the stack is in this state we are unable to get it refresh the value back, even if we manually update the tag to the correct value.

We are using version v2.14 of Azure Native and Pulumi v3.74.0

### Example

``` csharp
            var databaseServer = new Server(databasePulumiName, new ServerArgs
            {
                ServerName = args.Name,
                Version = args.Version,
                Location = args.Location,
                ResourceGroupName = args.ResourceGroupName,
                MinimalTlsVersion = "1.2",
                PublicNetworkAccess = args.IsPrivateLinkEnabled ? ServerNetworkAccessFlag.Disabled : ServerNetworkAccessFlag.Enabled,
                AdministratorLogin = args.AdminUserName,
                AdministratorLoginPassword = adminUserPassword,
                Administrators = new ServerExternalAdministratorArgs
                {
                    AdministratorType = AdministratorType.ActiveDirectory,
                    PrincipalType = PrincipalType.Group,
                    AzureADOnlyAuthentication = args.EntraIdOnlyAuth,
                    Login = administratorArgs.Apply(x => x.AdministratorGroupName),
                    Sid = administratorArgs.Apply(x => x.AdministratorSid),
                    TenantId = administratorArgs.Apply(x => x.AdministratorTenantId)
                },
                Identity = new ResourceIdentityArgs { Type = IdentityType.SystemAssigned },
                RestrictOutboundNetworkAccess = args.RestrictOutboundNetworkAccess ? ServerNetworkAccessFlag.Enabled : ServerNetworkAccessFlag.Disabled,
                Tags = args.Tags,
            }, sqlOptions);
```

### Output of `pulumi about`

We are using version v2.14 of Azure Native and Pulumi v3.74.0, Pulumi Automation 3.55

CLI
Version      3.91.0
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME          VERSION
azure         5.41.0
azure-native  2.14.0
azuread       5.38.0
command       0.5.2
dotnet        unknown
kubernetes    3.30.0
random        4.13.2

### Contributing

[...]

PR #3242, opened 2024-04-26
Document the special behavior of updating sql.Server passwords

thomas11:
This documentation intends to help users when encountering this surprising behavior and addresses #2937 at least in the short-medium term.

I'm not a native speaker so please feel free to suggest changes.


danielrbradley:
Does this not update the schema because we don't commit the one with the docs included?

thomas11:
> Does this not update the schema because we don't commit the one with the docs included?

Exactly. Here's the [registry preview for the change](http://registry--origin-pr-4414-f31e61e6.s3-website.us-west-2.amazonaws.com/registry/packages/azure-native/api-docs/sql/server/).

PR #3247, opened 2024-04-29
Use new reusable actions

danielrbradley:
1. Use version action for all workflows - the new action works for both dev and release versions.
2. Use new check-git-status action.


thomas11 (approved):

danielrbradley:
It appears the changes being flagged are from https://github.com/pulumi/pulumi-azure-native/commit/9b47251f40ea19d1b015194d6d378001b778fdae which didn't re-generate the SDKs

danielrbradley:
Rebased to fix outdated files

PR #3257, opened 2024-05-03
Bump azure/login, clean up unused environment from nightly-sdk-generation

thomas11:
Housekeeping. The environment variables were never used in this workflow.


danielrbradley (approved):
Nice!

PR #3260, opened 2024-05-06
Manual pulumi/pulumi upgrade to 3.115

thomas11:
The manual upgrade was necessary due to an API change in `dotnetgen.GeneratePackage`. Confirmed with @Frassle that `nil` is ok here.


VenelinMartinov (approved):
lgtm

PR #3266, opened 2024-05-07
Add communication.Domain docs

### LINKED ISSUE #3263: [Docs] `communication.Domain`: docs should indicate `domainName` must be set to `AzureManagedDomain` if `domainMainagement` set to `AzureManaged`
## Hello!

## Issue details

The docs page for `communication.Domain` (https://www.pulumi.com/registry/packages/azure-native/api-docs/communication/domain/#inputs) indicates that the `domainName` property is optional - which is usually true.
However, if the `domainManagement` property is set o `AzureManaged` then `domainName` has to be set to `AzureManagedDomain`

FWIW, the following code represents this:
```
const emailDomain = new azure_native.communication.Domain(`${baseName}-dom`, {
        domainName: "AzureManagedDomain", // THIS is the required name if using azure managed. If you don't set it with azure managed you'll get an error message about it.
        domainManagement: "AzureManaged", // This is the string equivalent of the enum as shown here: https://www.pulumi.com/registry/packages/azure-native/api-docs/communication/domain/#domainmanagement
        emailServiceName: emailService.name, 
        location: "global",
        resourceGroupName: resourceGroup.name,
    });
```

danielrbradley:
Fixes #3263

Add details on how to contribute documentation additions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,6 +34,11 @@ yarn link @pulumi/azure-native
 pulumi up
 ```
 
+## Adding Resource Documentation
+
+Documentation is mostly generated automatically from the specifications. Additional documentation can be added for specific resources by adding markdown files in the [docs/resources](./docs/resources/) folder. Each filename is the combination of the module and resource name as it appears in the registry: `[module]-[Resource].md` e.g. `sql-Server.md`.
+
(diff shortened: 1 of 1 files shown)


thomas11 (approved):
Nice, thanks for documenting this!

PR #3281, opened 2024-05-13
Fail early on mismatched environments/clouds

### LINKED ISSUE #3274: When using CLI auth, the provider should warn about mismatched Azure environments
Prompted by #3268. 

Azure has different clouds or "environments" like public and usgovernment. The environment of a Pulumi program cab be configured via `azure-native:environment`. However, when using the Azure CLI to authenticate, the provider will always use the environment we're logged in to, via `az login`.

This is somewhat unavoidable, since otherwise, Pulumi would need to change the CLI configuration. Both the subscription and the auth token are, to my knowledge, separate per cloud. Our authentication library uses `az account show` to determine the currently active subscription (from `az login`), and then takes the cloud from the subscription as well.

```json
❯ az account show
{
  "environmentName": "AzureCloud",    <--- regardless of pulumi config
}
```

Therefore, `azure-native:environment` has no effect when using the CLI to authenticate. For all other auth methods, this problem should not exist since there's no local state lookup like `az account show`, everything is explicitly configured.

We should warn or error when the configured environment is different from the one `az` is logged into.

thomas11:
Quoting the code comment: CLI authentication can only use the environment that the CLI is configured for. Fail early if that's not the one from config, to avoid obscure errors about missing subscriptions or endpoints later.

Resolves #3274 

Sample shell session:
```
❯ pulumi config -s gov | rg environ
azure-native:environment  usgovernment

❯ az account show | rg environ
  "environmentName": "AzureCloud",

❯ PATH=$HOME/pulumi/pan/bin:$PATH pulumi up -s gov
Previewing update (gov)

     Type                                     Name              Plan       Info
 +   pulumi:pulumi:Stack                      azn-3268-env-gov  create     3 warnings
     └─ azure-native:resources:ResourceGroup  resourceGroup                1 error

Diagnostics:
    error: The configured Azure environment 'usgovernment' does not match the determined environment 'public'.
    When authenticating using the Azure CLI, the configured environment needs to match the one shown by 'az account show'.
    You can change environments using 'az cloud set --name <environment>'.
```

### DIFF
--- a/provider/pkg/provider/auth.go
+++ b/provider/pkg/provider/auth.go
@@ -139,6 +139,15 @@ func (k *azureNativeProvider) getAuthConfig() (*authConfig, error) {
 		return nil, err
 	}
 
+	// CLI authentication can only use the environment that the CLI is configured for.
+	// Fail early if that's not the one from config, to avoid obscure errors about missing subscriptions or endpoints later.
+	if c.Environment != envName {
+		return nil, errors.Errorf(`The configured Azure environment '%s' does not match the determined environment '%s'.
+When authenticating using the Azure CLI, the configured environment needs to match the one shown by 'az account show'.
+You can change environments using 'az cloud set --name <environment>'.`,
+			envName, c.Environment)
+	}
+
 	return &authConfig{c, needCli}, nil
 }
 


danielrbradley (approved):

provider/pkg/provider/auth.go
```diff
@@ -139,6 +139,15 @@ func (k *azureNativeProvider) getAuthConfig() (*authConfig, error) {
 		return nil, err
 	}
 
+	// CLI authentication can only use the environment that the CLI is configured for.
+	// Fail early if that's not the one from config, to avoid obscure errors about missing subscriptions or endpoints later.
+	if c.Environment != envName {
```
mikhailshilkov, line 144: Can one of them be empty, while the other one is "public"? I'm a bit nervous we could break folks in this case. Or are both guaranteed to have a value at this point?

mikhailshilkov (commented):
I'm missing a test to go with this change.

provider/pkg/provider/auth.go
thomas11: They're guaranteed to have a value. Will add a test, though.

thomas11:
> I'm missing a test to go with this change.

Fair enough. The whole area isn't covered much.

https://github.com/pulumi/pulumi-azure-native/pull/3285 adds a bunch.

Issue #3243, opened 2024-04-26
AzureNative.App.ContainerApp failure to add referenced Azure Key Vault secret.


mentallabyrinth:
### What happened?

We're trying to add a secret which references Azure Key Vault into our Azure Container App. However, attempts to update fail with this error message:

```
azure-native:app:ContainerApp (dev-spa-eus2-aca0001):
    error: Code="ContainerAppSecretInvalid" Message="Invalid Request: Container app secret(s) with name(s) 'testing123' are invalid: value or keyVaultUrl and identity should be provided."
```

### Example

Here's the code that adds the secret into the container app.

```csharp
new AzureNative.App.Inputs.SecretArgs
{
      Name = "testing123",
      KeyVaultUrl = "https://dev-api-kv01.vault.azure.net/secrets/AZUREADCLIENT--CLIENTSECRET",
      Identity = "/subscriptions/<redacted>/resourcegroups/dev-app-eus2-rg0001fdfde73a/providers/Microsoft.ManagedIdentity/userAssignedIdentities/dev-api-eus2-mi0001a15f69b1s"
}
```

The failure occurs after running `pulumi -s dev up`

We were successful in adding a key vault referenced secret manually using the same configuration as above in the Azure portal. Inspecting the resource JSON we see the following:

```json
{
      "name": "testing",
      "keyVaultUrl": "https://dev-api-kv01.vault.azure.net/secrets/AZUREADCLIENT--CLIENTSECRET",
      "identity": "/subscriptions/<redacted>/resourcegroups/dev-app-eus2-rg0001fdfde73a/providers/Microsoft.ManagedIdentity/userAssignedIdentities/dev-api-eus2-mi0001a15f69b1"
}
```

### Output of `pulumi about`

CLI          
Version      3.90.1
Go Version   go1.21.9
Go Compiler  gc

Plugins
NAME          VERSION
azure-native  2.38.0
azuread       5.48.0
azuredevops   3.0.0
dotnet        unknown
random        4.16.0
twingate      0.0.48

Host     
OS       darwin
Version  13.5.2
Arch     arm64

This project is written in dotnet: executable='/nix/store/d38z3dpvqpaq8az30nnlp2lq37fjprbi-dotnet-sdk-8.0.101/bin/dotnet' version='8.0.101'

Dependencies:
NAME                                VERSION
Azure.Containers.ContainerRegistry  1.1.1
Pulumi                              3.63.0
[...]

PR #3284, opened 2024-05-14
Add note on app.ContainerApp API versions and KV secrets

thomas11:
Prompted by #3243. When creating a ContainerApp referencing a Key Vault secret via URL, Azure can at least sometimes respond with "_invalid: value or keyVaultUrl and identity should be provided_". This is a bug in the particular API version we use as default.

This PR adds a warning and a pointer later API versions to the registry docs.


danielrbradley (approved):
This looks helpful.

Out of interest, have we tried bumping the default version just to see if it's compatible?

thomas11:
> Out of interest, have we tried bumping the default version just to see if it's compatible?

I ran this, indicating that it's not.

```
❯ schema-tools squeeze -s bin/schema-full.json --old azure-native:app:ContainerApp --new azure-native:app/v20230501:ContainerApp
Found 2 breaking changes:
Resource "azure-native:app/v20230501:ContainerApp" missing input "workloadProfileType"
Resource "azure-native:app/v20230501:ContainerApp" missing output "workloadProfileType"
```

danielrbradley:
> > Out of interest, have we tried bumping the default version just to see if it's compatible?
> 
> I ran this, indicating that it's not.
> 
> ```
> ❯ schema-tools squeeze -s bin/schema-full.json --old azure-native:app:ContainerApp --new azure-native:app/v20230501:ContainerApp
> Found 2 breaking changes:
> Resource "azure-native:app/v20230501:ContainerApp" missing input "workloadProfileType"
> Resource "azure-native:app/v20230501:ContainerApp" missing output "workloadProfileType"
> ```

Great, thanks for confirming!

mikhailshilkov:
If we find that the default version is broken is some meaningful way, I would support us bumping it up to a better version even with minor breaking changes. It's a trade-off, but if we can take a closer look at what happened with `workloadProfileType` and confirm it's not reducing the functionality, I think a minor break is worth to fix the defaults.

Side note: please add some description to the PR. I know it feels like busywork, but I'd really want to encourage us to keep a high bar on PR text quality.

thomas11:
> If we find that the default version is broken in some meaningful way, I would support us bumping it up to a better version even with minor breaking changes. It's a trade-off, but if we can take a closer look at what happened with `workloadProfileType` and confirm it's not reducing the functionality, I think a minor break is worth to fix the defaults.

The difference in the spec is that `workloadProfileType` was replaced by `workloadProfileName`. Both are simple strings, but there sets of allowed values seem to be different based on the examples: `"workloadProfileType": "GeneralPurpose"` vs `"workloadProfileName": "My-GP-01"`.

So, seems like changing the API version would require everyone to update their code, not only changing the property name but also looking up the correct value. On the positive side, it's limited to that one property.

thomas11:
Addendum: v3 of the provider is also not too far away. Overall, we've decided to leave the API version as-is.

PR #3285, opened 2024-05-14
Test coverage for auth configuration

thomas11:
This is a core part of the provider that lacked coverage. Also includes the new #3281.

### DIFF
--- a/provider/pkg/provider/auth_test.go
+++ b/provider/pkg/provider/auth_test.go
@@ -3,6 +3,7 @@
 package provider
 
 import (
+	"os"
 	"testing"
 
 	goversion "github.com/hashicorp/go-version"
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):

mikhailshilkov:
It looks like some of the tests are failing

thomas11:
> It looks like some of the tests are failing

Hmm, it's because we're not logged in to Azure for the unit tests. As it should be, really. But without logging in, we can't test CLI functionality. Hmm... thoughts, anyone (@danielrbradley)?

mikhailshilkov:
Would it be hard to refactor the method to allow mocking of the Azure CLI dependency to turn these tests into pure unit tests of our code?

thomas11:
> Would it be hard to refactor the method to allow mocking of the Azure CLI dependency to turn these tests into pure unit tests of our code?

The call out to `az` is unfortunately not in our code but in a dependency, and it's in the same method that handles the other auth mechanisms as well.

mikhailshilkov:
> The call out to az is unfortunately not in our code but in a dependency, and it's in the same method that handles the other auth mechanisms as well.

Could we mock it out? Define an interface, pass the real implementation in the provider, pass a stub in the test, and only test our code (like the error logic that you added)?

thomas11:
> > The call out to az is unfortunately not in our code but in a dependency, and it's in the same method that handles the other auth mechanisms as well.
> 
> Could we mock it out? Define an interface, pass the real implementation in the provider, pass a stub in the test, and only test our code (like the error logic that you added)?

I don't see mocking as a good fit here. Like I said, most of the logic is in the dependency and the `az` tool, so if you mock it, the test just tests the mock. And that `"public" != "usgovernment"`, which we already know. I'm in camp "mocking is overused" anyway.

Also, this whole issue is an aside I picked up from triage and I'm not planning to spend time for larger refactorings on it.

Issue #3282, opened 2024-05-13
Import Azure-native.keyvault.AccessPolicy


welkin18fe:
### What happened?

Hello, I'm trying to import multiple access policies using the import command available [azure-native.keyvault.AccessPolicy](https://www.pulumi.com/registry/packages/azure-native/api-docs/keyvault/accesspolicy/#import), but it seems that something is missing. I tried with the object id like the Azure provider and principal name in case something is missing in the doc, but no luck.

### Example

I removed the names and ids to simplify.

Using :
`pulumi import azure-native:keyvault:AccessPolicy myresource1 /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}/accessPolicy`

```
panic: interface conversion: interface {} is nil, not resource.PropertyMap
    goroutine 77 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/resource.PropertyValue.ObjectValue(...)
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.107.0/go/common/resource/properties.go:451
    github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources/customresources.(*accessPolicyClient).read(0xc0006fe168, {0xb1e20e0, 0xc0006a7560}, {0xc0001641e0?, 0x0?}, 0xc0007caa90?)
        /home/runner/work/pulumi-azure-native/pulumi-azure-native/provider/pkg/resources/customresources/custom_keyvault_accesspolicy.go:100 +0x5fc
    github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider.(*azureNativeProvider).Read(0xc000961c00, {0xb1e20e0, 0xc0006a7560}, 0xc0002be480)
        /home/runner/work/pulumi-azure-native/pulumi-azure-native/provider/pkg/provider/provider.go:1068 +0x34a
    github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Read_Handler.func1({0xb1e20e0, 0xc0006a7560}, {0x1f85fc0?, 0xc0002be480})
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.107.0/proto/go/provider_grpc.pb.go:611 +0x75
[...]

PR #3295, opened 2024-05-17
Allow keyvault accesspolicy to have object id in its id

danielrbradley:
Fixes https://github.com/pulumi/pulumi-azure-native/issues/3282

For import to work, we need to be able to identify which access policy we're addressing and therefore need the object id to be part of the id.

We're unable to update the `id` of existing resources with the current provider implementation structure so need to be able to continue to parse ids in both the old and new format. It might be possible to refactor the provider to allow custom resources to return a new `id` but this should be undertaken in as follow-up work.

The Path property of the custom resource is used as an internal identifier, but this is looked up from the metadata which is generated by the resource type which is not changing, so this is backwards compatible.

The `crudClient.PrepareAzureRESTIdAndQuery` is that part which builds the `id`. `objectId` doesn't exist as a top-level property so we set it from within the `policy` instead (`policy.objectId`) - for which we also have to allow paths to contain references to nested properties.

## Testing

The added keyvault-accesspolicy test-program can be used to test the two scenarios for this change: id compatibility and import.

### 1) Upgrading

1. Deploy the program with a previous version of the provider (e.g. 2.43.0)
2. Upgrade the provider to include this change (e.g. via attaching to the running dev version)
3. Perform another `up` and `refresh` to ensure no changes are made.

Note: this can be targeted using `pulumi refresh -t urn:pulumi:dev::keyvault-accesspolicy::azure-native:keyvault:AccessPolicy::ap1`

### 2) Importing

Using the new version of the provider, import the access policy back in with a different name (e.g. `ap2`):

```
pulumi import azure-native:keyvault:AccessPolicy ap2 "$(pulumi stack output ap1Id)/$(pulumi stack output ap1ObjectId)"
```

### Automation

Ideally, these tests would be run via automation. However, there's no current test infrastructure written for handling upgrade testing or import testing so this will have to be done as follow-up work.

### DIFF
--- a/provider/pkg/provider/crud/crud_test.go
+++ b/provider/pkg/provider/crud/crud_test.go
@@ -44,6 +44,21 @@ func TestPathParamsErrorHandling(t *testing.T) {
 		}
 	})
 
+	t.Run("Path param from props", func(t *testing.T) {
+		id, _, err := PrepareAzureRESTIdAndQuery("/path/{p1}",
+			[]resources.AzureAPIParameter{
+				{
+					Name:     "p1",
+					Location: "path",
+					Value:    &resources.AzureAPIProperty{Type: "string"},
+				},
+			}, map[string]any{
+				"p1": "val",
+			}, nil)
+		require.NoError(t, err)
+		assert.Equal(t, "/path/val", id)
+	})
+
 	t.Run("Nested path param lookup from props", func(t *testing.T) {
 		id, _, err := PrepareAzureRESTIdAndQuery("/path/{container.p1}",
 			[]resources.AzureAPIParameter{
@@ -59,5 +74,24 @@ func TestPathParamsErrorHandling(t *testing.T) {
 			}, nil)
 		require.NoError(t, err)
 		assert.Equal(t, "/path/innerVal", id)
+	})
+
+	t.Run("Deeply nested path param", func(t *testing.T) {
+		id, _, err := PrepareAzureRESTIdAndQuery("/path/{container.inner.p1}",
+			[]resources.AzureAPIParameter{
+				{
+					Name:     "container.inner.p1",
+					Location: "path",
+					Value:    &resources.AzureAPIProperty{Type: "string"}, // correct, but value is not
+				},
+			}, map[string]any{
+				"container": map[string]any{
+					"inner": map[string]any{
+						"p1": "deepVal",
+					},
+				},
+			}, nil)
+		require.NoError(t, err)
+		assert.Equal(t, "/path/deepVal", id)
 	})
 }


provider/pkg/resources/customresources/custom_keyvault_accesspolicy.go
```diff
@@ -22,7 +22,10 @@ const (
 	policy    = "policy"
 )
 
-const path = "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}/accessPolicy"
+// This used to be "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}/accessPolicy"
+// This was updated to include the objectId in the path so it accurately represents the logical resource so we can perform imports by ID.
+// However, it's not possible to update an existing id, so we have to ensure we can still work with the old id format - where we fetch the objectId from the properties.
+const path = "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}/accessPolicy/{policy.objectId}"
```
mikhailshilkov, line 28: Consider adding a comment on why it has to be `policy.objectId`

provider/pkg/resources/customresources/custom_keyvault_accesspolicy.go
```diff
@@ -233,17 +243,25 @@ func (c *accessPolicyClient) modify(ctx context.Context, properties resource.Pro
 type vaultPathParams struct {
 	ResourceGroup string
 	VaultName     string
+	ObjectId      *string
 }
 
 func parseKeyVaultPathParams(id string) (vaultPathParams, error) {
```
mikhailshilkov, line 249: I'd love to see a test for this func

provider/pkg/provider/test-programs/keyvault-accesspolicy/Pulumi.yaml
```diff
@@ -0,0 +1,44 @@
...
+          name: standard
+        tenantId: ${config.tenantId}
+        enabledForDeployment: false
+      
+  ap1:
+    type: azure-native:keyvault:AccessPolicy
+    properties:
+      resourceGroupName: ${rg.name}
+      vaultName: ${vault.name}
+      policy:
```
mikhailshilkov, line 28: I previously missed that we have this unexpected design where we added this extra `policy` container to contain all policy properties. I guess it's biting us now with the requirement to work around `policy.objectId` identifier handling.

Why was this done? Couldn't we have `objectId`, `permissions`, `tenantId` etc on the top level?

provider/pkg/provider/crud/crud.go
```diff
@@ -130,7 +130,9 @@ func PrepareAzureRESTIdAndQuery(path string, parameters []resources.AzureAPIPara
 			sdkName = param.Value.SdkName
 		}
 		// Look in both `method` and `client` inputs with `method` first
-		val, has = methodInputs[sdkName]
+		// Navigate into each level of the path to find the value.
+
+		val, has = findInContainer(methodInputs, strings.Split(sdkName, "."))
```
mikhailshilkov, line 135: Probably unlikely - but have you double-checked that we have no existing properties with a dot in the name somehow?

provider/pkg/provider/crud/crud_test.go
```diff
@@ -42,4 +43,21 @@ func TestPathParamsErrorHandling(t *testing.T) {
 			assert.Equal(t, "expected string value for path parameter 'p1', got int", err.Error())
 		}
 	})
+
+	t.Run("Nested path param lookup from props", func(t *testing.T) {
+		id, _, err := PrepareAzureRESTIdAndQuery("/path/{container.p1}",
+			[]resources.AzureAPIParameter{
+				{
+					Name:     "container.p1",
```
mikhailshilkov, line 51: Add test cases for 1 segment and 3 segments to prove it works for arbitrary depth?

mikhailshilkov (commented):
Overall makes sense, asked a few tactical questions

provider/pkg/provider/test-programs/keyvault-accesspolicy/Pulumi.yaml
danielrbradley: Not 100% sure why this was done, but I would guess that this mirrors the shape of the original resource where each policy has it's objectId within it.

danielrbradley: It's a part of the keyvault resource - accessPolicies. We just created this resource to manage one at a time rather than having to set all of them on the original resource straight away. This also avoids circular dependencies.

Here's the original type within the Vault resource: https://www.pulumi.com/registry/packages/azure-native/api-docs/keyvault/vault/#accesspolicyentry

danielrbradley: Agreed, it might be nice to tidy up at some point, but would require a breaking change.

provider/pkg/provider/crud/crud.go
danielrbradley: I'm pretty sure that's pysically impossible as we're using the sdkName here.

provider/pkg/resources/customresources/custom_keyvault_accesspolicy.go
danielrbradley: 👍 updated the existing test and broken it out into sub-tests.

provider/pkg/provider/crud/crud_test.go
danielrbradley: Done 👍

provider/pkg/provider/crud/crud.go
mikhailshilkov: And yet we are splitting it on a `.`. It's any `param.Name` that is defined anywhere in the API specs, it could in theory be any string.
Probably fine, since our SDK gen would likely fail for an invalid identifier. Let's maybe add yet another comment explaining why exactly we are doing the split on an `sdkName` that is supposed to be a valid identifier.
I'd really love to clean this up later on, there are too many special cases here due to this `policy` inlining.

mikhailshilkov (approved):
LGTM, let's add a follow-up issue on flattening the `policy` property

danielrbradley:
> LGTM, let's add a follow-up issue on flattening the `policy` property

Thanks! Done: https://github.com/pulumi/pulumi-azure-native/issues/3333

Issue #1554, opened 2022-03-08
golang SDK eventgrid EventSubscriptionArgs Destination examples do not work


tjholm:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Trying to create a new eventgrid `EventSubscription` using golang and azure-native `v1.60.0` following Webhook example at:
https://www.pulumi.com/registry/packages/azure-native/api-docs/eventgrid/eventsubscription/

### Steps to reproduce

1. Attempt to create new eventgrid EventSubscription in golang using the webhook example https://www.pulumi.com/registry/packages/azure-native/api-docs/eventgrid/eventsubscription/ using azure-native `v1.60.0`.

Expected: Destination field accepts `eventgrid.WebHookEventSubscriptionDestination`
Actual: typing is not correct (program cannot be compiled, `eventgrid.WebHookEventSubscriptionDestination` does not implement `pulumi.Input`)

viveklak:
Thanks for the above. I believe using `WebHookEventSubscriptionDestinationArgs` should help getting unblocked here.

tjholm:
@viveklak thanks for your reply. I'm probably missing something but I don't think such a type exists, at least not in `v1.60.0` https://pkg.go.dev/github.com/pulumi/pulumi-azure-native/sdk@v1.60.0/go/azure/eventgrid

mikhailshilkov:
@Zaid-Ajaj The provider is now on pulumi 3.109.0 but the docs issue is still present. Could you please take another look?

Zaid-Ajaj:
Looking into this again, it remains a bug unfortunately. With https://github.com/pulumi/pulumi/pull/14679 merged, this was supposed to be fixed. In fact, generating the azure-native SDK locally _does_ correctly generate the `Args` version of `WebHookEventSubscriptionDestination`:

Repro with:
```bash
pulumi version # 3.116.1

# get schema
pulumi package get-schema azure-native > schema.json

# generate sdk locally
pulumi package gen-sdk --language go --out sdk schema.json
```
Will give you these types under `./sdk/go/eventgrid/pulumiTypes.go`

<details>
<summary>See generated types:</summary>

```go
// Information about the webhook destination for an event subscription.
type WebHookEventSubscriptionDestination struct {
	// The Azure Active Directory Application ID or URI to get the access token that will be included as the bearer token in delivery requests.
[...]

PR #3300, opened 2024-05-21
Enable generateExtraInputTypes for Go

thomas11:
GenerateExtraInputTypes makes the code generator generate input and output types for all plain types, instead of for only types that are used as input/output types.

We need this here because the Go SDK is special in that it's generated as separate modules, from a sub-schema per module. This causes SDK-gen to not infer the type used from a union as input.

Unfortunately, this change increases the Go SDK size by about 19%.

Addresses #1554


danielrbradley:
Once this is green, let run the [master workflow](https://github.com/pulumi/pulumi-azure-native/actions/workflows/master.yml) to create a 2.43-alpha branch in https://github.com/pulumi/pulumi-azure-native-sdk then link the diff back into this PR manually for completeness.

thomas11:
We have since determined that the apparently missing input types are actually present. There was a misunderstanding when investigating the issue. The remaining issue is limited to Go codegen for the examples, so `generateExtraInputTypes` is not required.

Issue #2990, opened 2023-12-18
Incorrect/missing code formatting on some examples


scottslowe:
### What happened?

While browsing the API docs for the Azure Native provider, I noted that numerous examples were incorrectly formatted or were missing formatting. While this doesn't make it impossible to use the example for guidance, it does significantly raise the level of effort required. Based on what I have observed so far, this only seems to affect examples in Go, and not any other languages.

URLs of specific examples:
https://www.pulumi.com/registry/packages/azure-native/api-docs/storage/storageaccount/ (see the second example in Go)
https://www.pulumi.com/registry/packages/azure-native/api-docs/compute/virtualmachine/ (see the first example in Go)

I strongly suspect there are more, but these are the ones I can for sure confirm.

I viewed these using Firefox, and have not checked any other browsers.

### Example

https://www.pulumi.com/registry/packages/azure-native/api-docs/storage/storageaccount/ (see the second example in Go)
https://www.pulumi.com/registry/packages/azure-native/api-docs/compute/virtualmachine/ (see the first example in Go)

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mjeffryes:
Thanks for reporting this @scottslowe; we'll take a look.

danielrbradley:
It seems that we've managed to reduce the number of incorrectly formated examples, but quite a few remain.

It's possible to identify the affected resources by searching for a newline directly followed by `pulumi.Run` - with no whitespace between e.g. using the regex `\npulumi\.Run`.

Some examples are always misformatted and some examples are sometimes misformatted.

Here's a list of resources which at least sometimes contain mis-formatted Go examples:
- `appplatform/ContainerRegistry`
- `appplatform/Job`
- `azurestackhci/UpdateRun`
- `datafactory/Dataset`
- `dataprotection/BackupInstance`
- `datareplication/Fabric`
- `datareplication/Policy`
- `datareplication/ProtectedItem`
- `datareplication/ReplicationExtension`
- `eventhub/AppicationGroup`
- `machinelearning/WebService`
- `machinelearningservices/Compute`
- `machinelearningservices/FeaturesetVersion`
- `machinelearningservices/Job`
- `machinelearningservices/LabelingJob`
- `media/Job`
- `media/MediaGraph`
- `migrate/WorkloadDeployment`
[...]

PR #3335, opened 2024-06-06
Fix generated Go import paths

### LINKED ISSUE #3308: pulumi convert emits incorrect import paths in generated Go programs due to azure-native v2 module structure
### What happened?

`pulumi convert` gives imcorrect import paths for generated Go programs in azure-native v2 and is unable to correctly install dependencies for these programs.

### Example

Put the following code in `main.pp` file:
```
resource "example" "azure-native:eventgrid:EventSubscription" {
    expirationTimeUtc = "example"
}
```
then 
```
pulumi convert --from pcl --language go --out example
```
You get:
```go
package main

import (
	"github.com/pulumi/pulumi-azure-native-sdk/v2/eventgrid"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		_, err := eventgrid.NewEventSubscription(ctx, "example", &eventgrid.EventSubscriptionArgs{
			ExpirationTimeUtc: pulumi.String("example"),
		})
		if err != nil {
			return err
		}
		return nil
	})
}
```
Installing the dependencies fails with:
```
Converting from pcl...
Converting to go...
Installing dependencies...

go: finding module for package github.com/pulumi/pulumi-azure-native-sdk/v2/eventgrid
go: azex imports
        github.com/pulumi/pulumi-azure-native-sdk/v2/eventgrid: module github.com/pulumi/pulumi-azure-native-sdk/v2@latest found (v2.42.1), but does not contain package github.com/pulumi/pulumi-azure-native-sdk/v2/eventgrid
error: installing dependencies failed; rerun manually to try again, then run `pulumi up` to perform an initial deployment: `go mod tidy` failed to install dependencies: exit status 1
```

Notice the path `github.com/pulumi/pulumi-azure-native-sdk/v2/eventgrid` is incorrect and it should be 
```go
import (
  eventgrid "github.com/pulumi/pulumi-azure-native-sdk/eventgrid/v2"
)
```
This is a combination of the base import path in the schema being set to `github.com/pulumi/pulumi-azure-native-sdk/v2` and in Go program-gen, we use the pattern `{baseImportPath}/{module}` when  computing imports in the programs. This convention will not work for azure-native v2 since we need a suffix `/v2` 

### Output of `pulumi about`

```
CLI          
Version      3.116.1
Go Version   go1.22.3
Go Compiler  gc

Host     
OS       darwin
Version  13.2.1
Arch     arm64
```

### Proposed solution

program-gen needs to be extended to allow emitting correct import paths for azure-native v2 programs. 

My idea is to implement a `go` language option `importPathPattern` which contains placeholders such as `{baseImportPath}` and `{module}`, then we replace those placeholders with actual values. 

Following that feature, the azure-native v2 would require a few changes:
 - Adding `"importPathPattern": "github.com/pulumi/pulumi-azure-native-sdk/{module}/v2"` to the Go language options
 - _fix_ the `packageImportAliases` language option such that keys that look like "github.com/pulumi/pulumi-azure-native-sdk/v2/aad/v2" become "github.com/pulumi/pulumi-azure-native-sdk/aad/v2" (notice the removed `/v2` before the module)

Note how I didn't actually use the `{baseImportPath}` placeholder in the pattern, this is because we don't the `/v2` from the base import path but we still want to keep in the base import path field to maintain the top-level import paths (when `module` == `index`)

### Contributing

danielrbradley:
Fixes #3308

- Set `importPathPattern` so convert can create module names accurately.
- Fix `golangImportAliases` so they don't have `v2` in twice.
- Update Pulumi dependencies & regenerate docs schema.

This updates the paths in all go examples. Exploded schema diff:
image

Each file has the following example change:

```diff
-  "github.com/pulumi/pulumi-azure-native-sdk/azurestackhci/v2"
+  azurestackhci "github.com/pulumi/pulumi-azure-native-sdk/azurestackhci/v2"
```

This also related to #2990 though appears it doesn't solve all formatting yet.


thomas11 (commented):
```diff
-  "github.com/pulumi/pulumi-azure-native-sdk/azurestackhci/v2"
+  azurestackhci "github.com/pulumi/pulumi-azure-native-sdk/azurestackhci/v2"
```

Why is this an improvement? Using the first form, Go will handle the `/v2` and you can just use `azurestackhci.Foo`.

danielrbradley:
> Why is this an improvement? Using the first form, Go will handle the `/v2` and you can just use `azurestackhci.Foo`.

I _think_ this change is inconsequential, but adding the metadata to the schema allows pulumi convert to function correctly. @Zaid-Ajaj is that correct?

Zaid-Ajaj:
> I think this change is inconsequential, but adding the metadata to the schema allows pulumi convert to function correctly. @Zaid-Ajaj is that correct?

That's right, I think we now get the an alias because it actually hit one of keys in `packageImportAliases` which might be redundant in this case but that what we want to happen when we generate the programs

Zaid-Ajaj (approved):

Issue #3347, opened 2024-06-12
Auth throws non-descriptive error if the `az` CLI is not installed


tomweston:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

### Affected area/feature

Azure Native Provider - Subscription Feature Registration [Docs](https://www.pulumi.com/registry/packages/azure-native/api-docs/features/subscriptionfeatureregistration/#azure-native-features-subscriptionfeatureregistration)

### Problem Description

Currently, enabling preview features in Azure (e.g., `NodeOSUpgradeChannelPreview` in the `Microsoft.ContainerService` namespace) requires the use of the Azure CLI. The current implementation expects the Azure CLI (`az`) to be available.

The equivalent Azure CLI commands are:

```
az feature register --namespace Microsoft.ContainerService --name NodeOSUpgradeChannelPreview
az feature list -o table --query "[?contains(name, 'Microsoft.ContainerService/NodeOSUpgradeChannelPreview')].{Name:name,State:properties.state}"
az provider register --namespace Microsoft.ContainerService
```

**Example:**

```
package main

import (
	"github.com/pulumi/pulumi-azure-native-sdk/features/v2"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		_, err := features.NewSubscriptionFeatureRegistration(ctx, "subscriptionFeatureRegistration", &features.SubscriptionFeatureRegistrationArgs{
			FeatureName:       pulumi.String("NodeOSUpgradeChannelPreview"),
			ProviderNamespace: pulumi.String("Microsoft.ContainerService"),
		})
		if err != nil {
			return err
		}

		return nil
	})
}
```

**Results in:**

```
     Type                                                      Name                             Plan       Info
 +   pulumi:pulumi:Stack                          example     
     └─ azure-native:features:SubscriptionFeatureRegistration  subscriptionFeatureRegistration             1 error

Diagnostics:
  azure-native:features:SubscriptionFeatureRegistration (subscriptionFeatureRegistration):
    error: could not find `az`: exec: "az": executable file not found in $PATH
```

### Suggested Solution

[...]

PR #3349, opened 2024-06-13
Improve auth CLI missing error

danielrbradley:
Add context for why the CLI is expected and what operation is being attempted.

Fixes https://github.com/pulumi/pulumi-azure-native/issues/3347

This will now print a message like:

```
Diagnostics:
  azure-native:features:SubscriptionFeatureRegistration (subscriptionFeatureRegistration):
    error: checking az cli version: could not find `az`: exec: "az": executable file not found in $PATH: MSI, OIDC, 
      client secret and client certificate authentication are not configured so we require the AZ CLI for authentication
```

### DIFF
--- a/provider/pkg/provider/auth.go
+++ b/provider/pkg/provider/auth.go
@@ -123,11 +123,12 @@ func (k *azureNativeProvider) getAuthConfig() (*authConfig, error) {
 
 	needCli := needAzCli(builder)
 	if needCli {
+		requireCliHint := "MSI, OIDC, client secret and client certificate authentication are not configured so we require the AZ CLI for authentication"
 		// Check that we have a good version of the cli (#1565). The check needs to happen before
 		// builder.Build(), or we return the less fitting error message from go-azure-helpers.
 		v, err := getAzVersion()
 		if err != nil {
-			return nil, err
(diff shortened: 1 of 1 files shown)


thomas11:
The message is missing a period, semicolon or the like after `$PATH`. I know the linter says not to use full sentences in errors, but maybe we should here?

danielrbradley:
> The message is missing a period, semicolon or the like after `$PATH`. I know the linter says not to use full sentences in errors, but maybe we should here?

Oh that was just me mocking up the diagnostics. It actually has a colon between. Updated the example.

thomas11 (approved):

provider/pkg/provider/auth.go
```diff
@@ -123,11 +123,12 @@ func (k *azureNativeProvider) getAuthConfig() (*authConfig, error) {
 
 	needCli := needAzCli(builder)
 	if needCli {
+		requireCliHint := "MSI, OIDC, client secret and client certificate authentication are not configured so we require the AZ CLI for authentication"
```
mikhailshilkov, line 126: > AZ CLI

It's called Azure CLI in all official product docs. If you mean the command, then it's probably lowercase `az`.

PR #3360, opened 2024-06-18
Allow diffs to unblock release

danielrbradley:
Fixes https://github.com/pulumi/pulumi-azure-native/issues/3354

Follow up issue: https://github.com/pulumi/pulumi-azure-native/issues/3361


thomas11 (approved):
Would be nice to add a TODO comment with the issue number to the added lines.

Issue #3361, opened 2024-06-18
Diff after refresh for privateEndpointConnections


danielrbradley:
Consistently failing tests:
- TestMessagingTs
- TestServicebusRecreateSdk

These tests need to be asserting no diff on refresh. We will allow the diff on these two tests as a temporary fix to unblock the release - implemented in https://github.com/pulumi/pulumi-azure-native/pull/3360

These each have a diff on their `privateEndpointConnections` method - where the diff is reporting that the property is being removed. It appears that the property was originally just an empty array.

This property on each resource is marked with `MaintainSubResourceIfUnset` which could be related to the issue.

All responses of the providers/Microsoft.EventHub/namespaces/{ns} are identical - after the initial create and the refresh.

PR #3374, opened 2024-06-21
Disable RequireEmptyPreviewAfterRefresh for 2 tests

danielrbradley:
Related to #3361

CI run including all tests: https://github.com/pulumi/pulumi-azure-native/actions/runs/9610436970


thomas11 (commented):
We'll need a green master build to prove it works since the PR workflow doesn't run all tests.

thomas11 (approved):

PR #3376, opened 2024-06-21
Pin pulumi CLI version

danielrbradley:


danielrbradley:
This does not fix the refresh issues in TestMessagingTs and TestServicebusRecreateSdk

https://github.com/pulumi/pulumi-azure-native/actions/runs/9615036572

PR #3378, opened 2024-06-24
Upgrade to p/p 3.121

thomas11:
Doing it manually because the automated job also upgrades p-java which we cannot do at the moment due to pulumi/pulumi-java#1363.


danielrbradley (approved):
If you could bump the version installed in the github workflow too that'd be great!

thomas11:
> If you could bump the version installed in the github workflow too that'd be great!

Done, could you double-check?

danielrbradley (approved):

PR #3382, opened 2024-06-24
Support OIDC file token authentication

### LINKED ISSUE #3373: Support using `ARM_OIDC_TOKEN_FILE_PATH ` to Configure OIDC
### What happened?

I followed the configuration guide(https://www.pulumi.com/registry/packages/azure-native/installation-configuration/#configuration-options) to set up CI runners in AKS cluster.

[Azure workload identity](https://azure.github.io/azure-workload-identity/docs/quick-start.html) injects an environment variable named `AZURE_FEDERATED_TOKEN_FILE` and its value is the file path mounted to Pod(/var/run/secrets/azure/tokens/azure-identity-token).

This means I cannot set up the Pulumi Azure native provider to auth with Azure using the required `ARM_OIDC_REQUEST_TOKEN` directly.

It is a common pattern to support `ARM_OIDC_TOKEN_FILE_PATH`. I also find that `ARM_OIDC_REQUEST_TOKEN ` exists in the source code. https://github.com/pulumi/pulumi-azure-native/blob/d1a236663f9224ccf09a04c6892e38db7da8a7a0/provider/pkg/provider/auth.go#L112

### Example

If I set up the provider like below

```
            export ARM_CLIENT_ID=$AZURE_CLIENT_ID
            export ARM_TENANT_ID=$AZURE_TENANT_ID
            export ARM_OIDC_TOKEN_FILE_PATH=$AZURE_FEDERATED_TOKEN_FILE
            export ARM_OIDC_REQUEST_URL=$AZURE_AUTHORITY_HOST
```

the error is `OIDC authentication was requested via useOidc/ARM_USE_OIDC but no token and/or request URL were configured. See
    https://www.pulumi.com/registry/packages/azure-native/installation-configuration/#credentials for more information.`

the workaround is to retrieve the token from file system.
```
            export ARM_CLIENT_ID=$AZURE_CLIENT_ID
            export ARM_TENANT_ID=$AZURE_TENANT_ID
            export ARM_OIDC_TOKEN=$(cat /var/run/secrets/azure/tokens/azure-identity-token)
            export ARM_OIDC_REQUEST_URL=$AZURE_AUTHORITY_HOST
```

### Output of `pulumi about`

CLI
Version      3.117.0
Go Version   go1.22.3
Go Compiler  gc

Host
OS       darwin
Version  14.4.1
Arch     arm64

### Additional context

I feel there could be some code missing, but I haven't invested more time.

Terraform Azure provider supports `oidc_token_file_path`, so the issue might also apply to Pulumi Azure classic provider.

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/service_principal_oidc

### Contributing

thomas11:
One typical way of authenticating via OIDC is that the identity provider supplies the OIDC token in a file. This provider already had support for this built in to the auth stack, but due to an oversight, validation wouldn't consider it and the provider would error before attempting to authenticate. This PR fixed this oversight and resolves #3373.

pulumi/registry#4790 updates the docs accordingly.


provider/pkg/provider/auth.go
```diff
@@ -109,7 +118,7 @@ func (k *azureNativeProvider) getAuthConfig() (*authConfig, error) {
 		IDTokenRequestToken: oidcConf.oidcRequestToken,
 		IDTokenRequestURL:   oidcConf.oidcRequestUrl,
 		IDToken:             oidcConf.oidcToken,
-		IDTokenFilePath:     k.getConfig("oidcTokenFilePath", "ARM_OIDC_TOKEN_FILE_PATH"),
+		IDTokenFilePath:     oidcConf.oidcTokenFilePath,
```
danielrbradley, line 121: oidcTokenFilePath doesn't appear to be available in the provider object arguments. I.e. I think we should also add this here: 
https://github.com/pulumi/pulumi-azure-native/blob/b7d82161560bc42c1ba0269f4738626d514e6848/provider/pkg/gen/schema.go#L131-L147

danielrbradley (commented):
Core change looks okay, but I think we need to update the provider config in the schem, and update the docs too: https://www.pulumi.com/registry/packages/azure-native/installation-configuration/

provider/pkg/provider/auth.go
thomas11: Ah, good point.

thomas11:
> Core change looks okay, but I think we need to update the provider config in the schem, and update the docs too: https://www.pulumi.com/registry/packages/azure-native/installation-configuration/

Registry PR is already open, see PR description

provider/pkg/provider/auth.go
thomas11: done

danielrbradley (approved):

PR #3400, opened 2024-07-01
Enable TypedDict input types for the Python SDK

julienp:
Enable TypedDict input types for the Python SDK. The only manual change is to set the InputTypes flag in the python language features of the schema. This is done here https://github.com/pulumi/pulumi-azure-native/commit/5a99a477ea930cf8e2728f51ce7e4de9d15479a7. This flag is currently opt in, but will default to classes-and-dicts in the future.

Epic https://github.com/pulumi/pulumi/issues/12689


thomas11:
I don't know enough Python to properly review this but it LGTM, I guess.

How much larger is the new Python SDK? SDK size is already an issue for this provider.

julienp:
> I don't know enough Python to properly review this but it LGTM, I guess.
> 
> How much larger is the new Python SDK? SDK size is already an issue for this provider.

Good question! 

The  release from PyPi measures 73 MB and the locally generated package with this change is 76 MB.

We've released this typing change to AWS and Kubernetes, so I feel fairly confident about it.

thomas11 (approved):

thomas11:
Had to approve via the CLI due to the number of files :)

PR #3456, opened 2024-07-22
Fall back to property name when an enum has no defined name

thomas11:
Azure introduced [an x-ms-enum without a name attribute](https://github.com/Azure/azure-rest-api-specs/commit/83f13c5688a9bccd29569a9404ff4f96dd01598b#diff-81adef30b508d8dcb6b3878969b13a2cb45b5f790f7f55cb928b4ce68f1e4c7bR499). This fails our nightly SDK generation with

> panic: failed to generate ‘azure-native:azurefleet:Fleet’: request type: name key missing from enum metadata

[Their docs](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/creating-swagger.md#x-ms-enum-extension-structure-) don’t explicitly say whether name is required or not.

The property looks like this:

```
        "passName": {
           "type": "string",
           "description": "The pass name. Currently, the only allowable value is OobeSystem.",
           "enum": [
             "OobeSystem"
           ],
           "x-ms-enum": {
             "modelAsString": false
           }
         },
```

This PR adds a fall back to the property name as the enum name. Otherwise, we’d have to hard-code a name for such enums.


danielrbradley (approved):
LGTM

PR #3464, opened 2024-07-25
Custom Read for WebApp to get the full siteConfig

### LINKED ISSUE #1468: Refresh doesn't detect changes in WebApp.SiteConfig
Adventure 1:

following code will make the program to hang out for several minutes with not descriptive error of  **Code="Failed" Message="The async operation failed." AdditionalInfo=[{"Message":"An error has occurred."}]:**, with debug or verbose modes not helping too much or even worsening the situation pointing on random earlier invocations:
image

Notice that this code, strict null in specific, I simplified it to the root cause , my code has some abstractions (Template Method) that due to unfinished implementation ,resulted in this null. So despite I had the problem with my code I think the error result should be better described somehow (if this is possible) and the program should fail fast, not hang out.

So it doesn't have a significant impact but it may be related to my second adventure:

Adventure 2:
if I put the following code for IP restrictions:
```
new IpSecurityRestrictionArgs
            {
                Action = "Allow",
                Priority = 1,
                Name = "Allow xxx",
                IpAddress = "xxx"
            };
```
it provisions restriction nicely.
However, when I'm importing WebApp with such SiteConfig, Pulumi doesn't see it (ipsecurityrestriction) exists and wants to add it again. I know there was/is a lot of issues with SiteConfig, hope this one is not a one of them and can be solved.

thomas11:
WebApp.SiteConfig is created and updated via a PUT to the WebApp itself, but a GET of the WebApp does not return the SiteConfig. We need to make a separate GET request to `/config/web` and merge the results. 

I verified that the integration test fails on `master`.

Fixes #1468

Also contains:
- **Small corrections to existing error messages**
- **Use custom Read even in default Create and Update**
- **Extract repeated read-after-write logic**

### DIFF
--- a/provider/pkg/provider/provider_test.go
+++ b/provider/pkg/provider/provider_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/convert"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider/crud"
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources"
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources/customresources"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
@@ -341,5 +342,85 @@ func TestInvokeResponseToOutputs(t *testing.T) {
 }
 
 func TestReader(t *testing.T) {
-	var customRes *customresources.CustomResource
+	t.Run("custom Read", func(t *testing.T) {
+		var customReads []string
+		customRes := &customresources.CustomResource{
+			Read: func(ctx context.Context, id string, inputs resource.PropertyMap) (map[string]any, bool, error) {
+				customReads = append(customReads, id)
+				return map[string]any{}, true, nil
+			},
+		}
+
+		azureClient := &mockAzureClient{}
+		crudClient := crud.NewResourceCrudClient(azureClient, nil, nil, "123", nil)
+
+		r := reader(customRes, crudClient)
+		_, err := r(context.Background(), "id1", nil)
+		require.NoError(t, err)
+		assert.Equal(t, []string{"id1"}, customReads)
+		assert.Empty(t, azureClient.getIds)
+	})
+
+	t.Run("no custom Read", func(t *testing.T) {
+		resource := &resources.AzureAPIResource{
+			Response: map[string]resources.AzureAPIProperty{},
+		}
+
+		for _, otherCustomRes := range []*customresources.CustomResource{nil, {} /* custom resource that doesn't implement Read */} {
+			azureClient := &mockAzureClient{}
+			crudClient := crud.NewResourceCrudClient(azureClient, nil, nil, "123", resource)
+
+			r := reader(otherCustomRes, crudClient)
+			_, err := r(context.Background(), "id2", nil)
+			require.NoError(t, err)
+			assert.Contains(t, azureClient.getIds, "id2")
+		}
+	})
+}
+
+func TestReadAfterWrite(t *testing.T) {
+	read := false
+	var reader readFunc = func(ctx context.Context, id string, inputs resource.PropertyMap) (map[string]any, error) {
+		read = true
+		return nil, nil
+	}
+
+	for _, skipReadOnUpdate := range []bool{true, false} {
+		p := azureNativeProvider{
+			skipReadOnUpdate: skipReadOnUpdate,
+		}
+		p.readAfterWrite(context.Background(), "id", "urn", "create", resource.PropertyMap{}, reader)
+		assert.Equal(t, !skipReadOnUpdate, read)
+	}
+
+}
+
+type mockAzureClient struct {
+	getIds []string
+}
+
+func (m *mockAzureClient) Delete(ctx context.Context, id, apiVersion, asyncStyle string, queryParams map[string]any) error {
+	return nil
+}
+func (m *mockAzureClient) CanCreate(ctx context.Context, id, path, apiVersion, readMethod string, isSingletonResource, hasDefaultBody bool, isDefaultResponse func(map[string]any) bool) error {
+	return nil
+}
+func (m *mockAzureClient) Get(ctx context.Context, id string, apiVersion string) (any, error) {
+	m.getIds = append(m.getIds, id)
+	return map[string]any{}, nil
+}
+func (m *mockAzureClient) Head(ctx context.Context, id string, apiVersion string) error {
+	return nil
+}
+func (m *mockAzureClient) Patch(ctx context.Context, id string, bodyProps map[string]interface{}, queryParameters map[string]interface{}, asyncStyle string) (map[string]interface{}, bool, error) {
+	return nil, false, nil
+}
+func (m *mockAzureClient) Post(ctx context.Context, id string, bodyProps map[string]interface{}, queryParameters map[string]interface{}) (any, error) {
+	return nil, nil
+}
+func (m *mockAzureClient) Put(ctx context.Context, id string, bodyProps map[string]interface{}, queryParameters map[string]interface{}, asyncStyle string) (map[string]interface{}, bool, error) {
+	return nil, false, nil
+}
+func (m *mockAzureClient) IsNotFound(err error) bool {
+	return false
 }


provider/pkg/resources/customresources/customresources.go
```diff
@@ -82,9 +91,10 @@ func BuildCustomResources(env *azureEnv.Environment,
 		newStorageAccountStaticWebsite(env, &storageAccountsClient),
 		newBlob(env, &storageAccountsClient),
 		// Customization of regular resources
-		customWebAppDelete(lookupResource, azureClient),
+		// customWebAppDelete(lookupResource, azureClient),
```
danielrbradley, line 94: Do we want to remove this commented line?

danielrbradley: Also, should we be deleting the `customWebAppDelete` function too now?

danielrbradley (approved):
Looks good 👍  One possible clean up mentioned inline.

provider/pkg/resources/customresources/custom_webapp.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return &CustomResource{
+		path: "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}",
+		tok:  webAppResourceType,
+
+		// WebApp.SiteConfig is created and updated via a PUT to the WebApp itself, but a GET of
+		// the WebApp does not return the SiteConfig. We need to make a separate GET request to
+		// /config/web and merge the results. #1468
+		Read: func(ctx context.Context, id string, inputs resource.PropertyMap) (map[string]any, bool, error) {
+			apiVersion, ok := versionLookup.GetDefaultApiVersionForResource("Storage", "BlobContainer")
```
mikhailshilkov, line 40: Why Storage/BlobContainer?

provider/pkg/resources/customresources/custom_webapp.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	return &CustomResource{
+		path: "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}",
+		tok:  webAppResourceType,
+
+		// WebApp.SiteConfig is created and updated via a PUT to the WebApp itself, but a GET of
+		// the WebApp does not return the SiteConfig. We need to make a separate GET request to
+		// /config/web and merge the results. #1468
+		Read: func(ctx context.Context, id string, inputs resource.PropertyMap) (map[string]any, bool, error) {
```
mikhailshilkov, line 39: I think it would be cleaner to refactor Read and Delete implementations into their separate functions and call them from here.

provider/pkg/resources/customresources/customresources.go
```diff
@@ -42,7 +45,7 @@ type CustomResource struct {
 	Create func(ctx context.Context, id string, inputs resource.PropertyMap) (map[string]interface{}, error)
 	// Read the state of an existing resource. Constructs the resource ID based on input values. Returns a map of
 	// resource outputs. If the requested resource does not exist, the second result is false.
-	Read func(ctx context.Context, id string, inputs resource.PropertyMap) (map[string]interface{}, bool, error)
+	Read CustomReadFunc
```
mikhailshilkov, line 48: Why do this for `Read` but not other functions?

provider/pkg/resources/customresources/custom_webapp.go
```diff
@@ -0,0 +1,105 @@
...
+			siteConfigResponse, ok := siteConfigResponseRaw.(map[string]any)
+			if !ok {
+				return nil, true, fmt.Errorf("unexpected response type %T", siteConfigResponseRaw)
+			}
+
+			if err := mergeWebAppSiteConfig(webAppResponse, siteConfigResponse); err != nil {
+				return nil, true, err
+			}
+
+			return crudClient.ResponseBodyToSdkOutputs(webAppResponse), true, nil
```
mikhailshilkov, line 64: It's a bit odd to have the conversion methods on crud client... But I guess that's not introduced in this PR.

provider/pkg/resources/customresources/custom_webapp.go
```diff
@@ -0,0 +1,105 @@
...
+
+func mergeWebAppSiteConfig(webApp, siteConfig map[string]any) error {
+	if outerProperties, ok := util.GetInnerMap(webApp, "properties"); ok {
+		outerProperties["siteConfig"] = siteConfig["properties"]
+		return nil
+	}
+	return fmt.Errorf("'properties' not found in response, cannot update siteConfig")
+}
+
+func createCrudClient(crudClientFactory crud.ResourceCrudClientFactory, lookupResource resources.ResourceLookupFunc) (crud.ResourceCrudClient, error) {
```
mikhailshilkov, line 94: Probably this function can be made universal and reused in other custom resources?

provider/pkg/resources/customresources/customresources.go
thomas11: Correct, done

examples/py-webapp/__main__.py
```diff
@@ -0,0 +1,29 @@
+from pulumi_azure_native import resources
```
mikhailshilkov, line 1: Nit: shouldn't we be creating all new examples in YAML?

provider/pkg/provider/provider.go
```diff
@@ -910,26 +919,48 @@ func (k *azureNativeProvider) defaultCreate(ctx context.Context, req *rpc.Create
...
+		// Map the raw response to the shape of outputs that the SDKs expect.
+		response = crudClient.ResponseBodyToSdkOutputs(response)
+	}
+	return id, response, nil
+}
+
+// Read the state of the resource from its Read endpoint. While we already have resource state from
+// the PUT reponse, it may not match precisely the shape of Read reponses that we also use for Refresh
+// operations. That discrepancy may cause noisy diffs that are hard for users to reconcile.
+func (p *azureNativeProvider) readAfterWrite(ctx context.Context, id, urn, opForLogging string,
```
mikhailshilkov, line 937: Are we missing tests for all the changes in this file?

provider/pkg/provider/provider.go
```diff
@@ -910,26 +919,48 @@ func (k *azureNativeProvider) defaultCreate(ctx context.Context, req *rpc.Create
...
+	inputs resource.PropertyMap, crudClient crud.ResourceCrudClient, customRead customresources.CustomReadFunc,
```
mikhailshilkov, line 938: Could we pass a single `read` function instead of passing both crudClient and customRead?

provider/pkg/provider/provider.go
```diff
@@ -910,26 +919,48 @@ func (k *azureNativeProvider) defaultCreate(ctx context.Context, req *rpc.Create
...
-		}
+	if readResponse := k.readAfterWrite(ctx, id, req.GetUrn(), "Create", inputs, crudClient, customRead); readResponse != nil {
+		response = readResponse
 	}
 
-	// Map the raw response to the shape of outputs that the SDKs expect.
-	outputs := crudClient.ResponseBodyToSdkOutputs(response)
-	return id, outputs, nil
+	// Custom resources take care of SDK conversion themselves.
+	if !(k.skipReadOnUpdate && customRead != nil) {
```
mikhailshilkov, line 927: Kind of inconcrete, but this whole method feels very fragile with conditions like this one scattered around. Is there a better abstraction we can design here?

mikhailshilkov (commented):
Took the first pass with a few suggestions

provider/pkg/resources/customresources/custom_webapp.go
thomas11: Oops, copy-paste error. Test passed anyway - versions probably match.

provider/pkg/resources/customresources/custom_webapp.go
thomas11: This crud client is resource-aware, i.e., there's an instance per resource type and it uses the resource internally. In this case, using the resource's response shape.

I agree that factoring is not quite right yet in that this method should probably be internal.

examples/py-webapp/__main__.py
thomas11: Is that an official guideline now? I know we've been talking about it but there isn't a test harness to YAML yet and I wanted to focus on this PR. I picked python because the python suite has the shortest run time.

mikhailshilkov: That was the goal some time ago, but I don't know where we stand. Feel free to ignore if the infrastructure is not there, definitely no need to impose unrelated work on you here.

provider/pkg/provider/provider.go
thomas11: I agree. I just pushed a refactoring that goes in this direction. The main underlying problem is still that custom resources were designed to be completely separate code paths, but actually in many cases they're just variations of the regular paths.

provider/pkg/provider/provider.go
thomas11: added

PR #3467, opened 2024-07-26
Re-enable postgres test, take 2

### LINKED ISSUE #898: Cannot create Postgres Flexible Server Configuration [was: Postgres test is failing]

thomas11:
Previously: #3218, #3227

After a bit more experimentation, I believe the `Configuration` was the problem. Without it, I didn't get the "_An unexpected error occurred while processing the request_" error from Azure.

I don't know what the problem with `Configuration` is but I propose to not spend any more time here and at least have a test for `Server`.

[Three successful master runs in a row](https://github.com/pulumi/pulumi-azure-native/actions/runs/10091454161)

Resolves #898

### DIFF
--- a/examples/postgres/index.ts
+++ b/examples/postgres/index.ts
@@ -27,13 +27,14 @@ const flexibleServer = new postgresql.Server("server", {
     },
 });
 
-new postgresql.Configuration("ssl", {
-    resourceGroupName: resourceGroup.name,
-    serverName: flexibleServer.name,
-    configurationName: "ssl",
-    value: "on",
-    source: "user-override",
-});
+// Blocked tu to #898
+// new postgresql.Configuration("ssl", {
+//     resourceGroupName: resourceGroup.name,
+//     serverName: flexibleServer.name,
+//     configurationName: "ssl",
+//     value: "on",
+//     source: "user-override",
+// });
 
 export const resourceGroupName = resourceGroup.name;
 export const serverName = flexibleServer.name;


examples/postgres/index.ts
```diff
@@ -17,19 +17,12 @@ const flexibleServer = new postgresql.Server("server", {
...
     administratorLoginPassword: `pa$$w0rd`,
-    version: "12",
+    version: "14",
     storage: {
-        storageSizeGB: 32,
+        storageSizeGB: 64,
     },
 });
-
-new postgresql.Configuration("backslash_quote", {
```
mikhailshilkov, line 30: I think I may have created this test to test the `Configuration` resource specifically. Maybe keep it commented out with a note why? This probably means our users are facing the same issue, potentially.

examples/postgres/index.ts
```diff
@@ -4,9 +4,9 @@ import * as postgresql from "@pulumi/azure-native/dbforpostgresql";
 import * as resources from "@pulumi/azure-native/resources";
 
 const resourceGroup = new resources.ResourceGroup("rg", {
-    // westus2 was not available for Postgres Flexible Server at the time of writing, see
-    // https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/overview#azure-regions
-    location: "uksouth",
+    // // westus2 was not available for Postgres Flexible Server at the time of writing, see
+    // // https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/overview#azure-regions
```
mikhailshilkov, line 8: Nit: double comment

mikhailshilkov (approved):

examples/postgres/index.ts
thomas11: I've tried to create a `Configuration` a few more times with different configurations and variations, without success. I added my data to #898, and changed the title to show that the problem is not limited to the test.

PR #3480, opened 2024-08-01
Filter REDACTED user name from SiteConfig to avoid meaningless diffs

thomas11:
A follow-up to #3464. The WebApp API has the quirk that it accepts or auto-creates a "publishing user name" on creation, but when the WebApp's siteConfig is read, it shows as "REDACTED". This guarantees a meaningless diff, so filter it out.


VenelinMartinov (approved):
lgtm

PR #3484, opened 2024-08-02
Add an exception for broken VMSS.diskController property type: should be string

### LINKED ISSUE #2639: Type Mismatch: Virtual Machine Scaleset - VM Profile - Storage Profile - Disk Controller Type
### What happened?

When migrating to a newer version of the compute provider, virtual machine scale sets get a new option: `diskControllerType`, which should be a `string`.

This field is defined as a string here: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2023-03-01/virtualMachineScaleSet.json#L4091

The Azure REST API returns a string on read, and expects a string - it returns an error when providing a string array.

However, MS's documentation (https://learn.microsoft.com/en-us/rest/api/compute/virtual-machine-scale-sets/create-or-update?tabs=HTTP#virtualmachinescalesetstorageprofile) defines the field as a string array, and pulumi also defines it as such.

We can no longer refresh, preview, or deploy our stacks, due to read errors from this type mismatch. Pulumi expects a string array when reading the resource, but receives a string, and so errors out.

```
{
    "sequence": 112,
    "timestamp": 16914XXXXX,
    "diagnosticEvent": {
        "urn": "urn:pulumi:XXXX::XXXX::azure-native:compute/v20230301:VirtualMachineScaleSet::XXXX",
        "prefix": "error: ",
        "message": "azure-native:compute/v20230301:VirtualMachineScaleSet resource 'XXXX' has a problem: 'virtualMachineProfile.storageProfile.diskControllerType' should be of type 'array' but got a string\n",
        "color": "never",
        "severity": "error"
    }
}
```

### Expected Behavior

We expect pulumi to accept string values for this resource option.

### Steps to reproduce

1. Manually deploy a virtual machine scaleset
2. Note the `vmProfile.storageProfile.diskControllerType` value in the JSON view
3. Attempt to import the resource with pulumi

### Output of `pulumi about`

We use the pulumi automation api.

```
CLI          
Version      3.75.0
Go Version   go1.20.5
Go Compiler  gc

Host     
OS       arch
Version  "rolling"
Arch     x86_64

Pulumi locates its logs in /tmp by default
warning: Failed to read project: no Pulumi.yaml project file found (searching upwards from /home/XXXX/projects/XXXX/XXXX). If you have not created a project yet, use `pulumi new` to do so: no project file found
```

### Additional context

The root cause is likely a documentation mistake upstream / in MS's court. I will create an issue there shortly; the primary purpose of this issue is to document the impact and to use as a reference.

### Contributing

thomas11:
Special case for [VirtualMachineScaleSet.diskControllerType](https://www.pulumi.com/registry/packages/azure-native/api-docs/compute/virtualmachinescaleset/#diskcontrollertype_go). This property is erroneously defined as a string _and_ an array in the spec. Primitive string is correct. See #2639, upstream issue Azure/azure-rest-api-specs#25202. Fixes #2639.

Note: I tried going with the primitive type over the array for all such properties but it caused a large diff in the azurestackhci SDK, so I opted not to do that.

There's no test since there's no test coverage for the whole complex `genTypeSpec` method. That would have made it necessary to refactor and invent a proper testing harness which was beyond the scope of this quick fix. The SDK output is pretty obvious to review. I also tested it in a local program.

### DIFF
--- a/sdk/dotnet/Compute/Inputs/VirtualMachineScaleSetStorageProfileArgs.cs
+++ b/sdk/dotnet/Compute/Inputs/VirtualMachineScaleSetStorageProfileArgs.cs
@@ -28,12 +28,7 @@ namespace Pulumi.AzureNative.Compute.Inputs
         }
 
         [Input("diskControllerType")]
-        private InputList<Union<string, Pulumi.AzureNative.Compute.DiskControllerTypes>>? _diskControllerType;
-        public InputList<Union<string, Pulumi.AzureNative.Compute.DiskControllerTypes>> DiskControllerType
-        {
-            get => _diskControllerType ?? (_diskControllerType = new InputList<Union<string, Pulumi.AzureNative.Compute.DiskControllerTypes>>());
-            set => _diskControllerType = value;
-        }
+        public Input<string>? DiskControllerType { get; set; }
 
         /// <summary>
         /// Specifies information about the image to use. You can specify information about platform images, marketplace images, or virtual machine images. This element is required when you want to use a platform image, marketplace image, or virtual machine image, but is not used in other creation operations.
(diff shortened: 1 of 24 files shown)


mjeffryes (approved):
I noticed that this field shows up in several versions of the API. Are we reasonably confident that it's a string not an array in all versions?

Issue #3472, opened 2024-07-30
Unable to delete `defenderForStorage` resources


stooj:
### What happened?

When trying to delete a [azure-native.security.DefenderForStorage](https://www.pulumi.com/registry/packages/azure-native/api-docs/security/defenderforstorage/) resource with azure-native versions >= 2.49.0, the azure api returns an error:

```
error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="Properties can't be null"
```

In Azure Native versions < 2.49.0, the REST call is:

```
PUT https://management.azure.com/%2Fsubscriptions%2F***************%2FresourceGroups%2FDefenderStorageTeste8525af1%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Fdefendertest048a72fd/providers/Microsoft.Security/defenderForStorageSettings/current?api-version=2022-12-01-preview

Body:
{"properties":{"isEnabled":false,"overrideSubscriptionLevelSettings":true}}
```

But versions >= 2.49.0 are sent as the following:
```
PUT https://management.azure.com/subscriptions/*************/resourceGroups/DefenderStorageTestd0e1996e/providers/Microsoft.Storage/storageAccounts/defendertest5f49c189/providers/Microsoft.Security/defenderForStorageSettings/current?api-version=2022-12-01-preview

Body:
{}
```

### Example

```python
import getpass

import pulumi
from pulumi_azure_native import eventgrid, resources, security, storage

username = getpass.getuser()
case_id = "3472"

config = pulumi.Config()
location = config.get("location", "uksouth")

resource_group = resources.ResourceGroup(f"rg{username.title()}{case_id}")

account = storage.StorageAccount(
    f"sa{username}{case_id}",
    resource_group_name=resource_group.name,
    sku=storage.SkuArgs(
        name=storage.SkuName.STANDARD_LRS,
    ),
    kind=storage.Kind.STORAGE_V2,
)

topic = eventgrid.Topic(
    f"topic{username.title()}{case_id}",
    public_network_access=eventgrid.PublicNetworkAccess.ENABLED,
    resource_group_name=resource_group.name,
)

subscription_id = resource_group.id.apply(lambda id: id.split("/")[2])

defender_for_storage = security.DefenderForStorage(
    f"defenderForStorage{username.title()}{case_id}",
    properties={
        "is_enabled": True,
        "malware_scanning": {
            "on_upload": {
                "cap_gb_per_month": -1,
[...]

PR #3488, opened 2024-08-05
Test default state conversion

danielrbradley:
Fixes https://github.com/pulumi/pulumi-azure-native/issues/3472

- Add test to validate that all default states match the expected shape of the API for every version we include in the SDK.
- During conversion, report on fields which are passed by the user but not used in the request.
- Add version-specific default resource states as some properties are renamed in specific versions.

### DIFF
--- a/provider/pkg/convert/sdkInputsToRequestBody.go
+++ b/provider/pkg/convert/sdkInputsToRequestBody.go
@@ -70,6 +70,8 @@ func (k *SdkShapeConverter) convertSdkPropToRequestBodyPropValue(id string, prop
 
 		// Otherwise, delegate to the normal map conversion flow.
 		converted, _ := k.SdkInputsToRequestBody(props, values, id)
+		// We ignore the error here as we haven't previously handled these errors recursively through convertTypedSdkInputObjectsToRequestBody.
+		// A difficulty is that the error is treated as a warning if we got a valid result, so should only be thrown if `converted` is nil.
 		return converted
 	})
 }


provider/pkg/convert/sdkInputsToRequestBody.go
```diff
@@ -47,8 +68,9 @@ func (k *SdkShapeConverter) convertSdkPropToRequestBodyPropValue(id string, prop
 			}
 		}
 
-		// Otherwise, delegate to the normal map convertion flow.
-		return k.SdkInputsToRequestBody(props, values, id)
+		// Otherwise, delegate to the normal map conversion flow.
+		converted, _ := k.SdkInputsToRequestBody(props, values, id)
```
thomas11, line 72: We probably want a comment and annotation here to silence the "error ignored" warning

provider/pkg/provider/crud/crud.go
```diff
@@ -178,14 +184,15 @@ func findInContainer(container map[string]any, path []string) (interface{}, bool
...
+	methodInputs map[string]any, converter *convert.SdkShapeConverter) (map[string]any, error) {
 	// Build the body JSON.
 	var body map[string]interface{}
+	var err error
 	for _, param := range parameters {
 		if param.Location != "body" {
 			continue
 		}
-		body = converter.SdkInputsToRequestBody(param.Body.Properties, methodInputs, id)
+		body, err = converter.SdkInputsToRequestBody(param.Body.Properties, methodInputs, id)
```
thomas11, line 195: Should the following loop on line 200 happen if `err != nil`?

provider/pkg/provider/provider.go
```diff
@@ -308,7 +309,14 @@ func (k *azureNativeProvider) Invoke(ctx context.Context, req *rpc.InvokeRequest
 		if err != nil {
 			return nil, err
 		}
-		body := crud.PrepareAzureRESTBody(id, parameters, nil, args.Mappable(), k.converter)
+		body, err := crud.PrepareAzureRESTBody(id, parameters, nil, args.Mappable(), k.converter)
+		if err != nil {
+			if body == nil {
+				return nil, fmt.Errorf("error preparing body for %s: %v", label, err)
+			}
+			// Log the error and continue with whatever was convertible.
```
thomas11, line 317: Why continue instead of failing? Doesn't this risk creating hard-to-debug and reproduce errors?

provider/pkg/provider/crud/crud.go
danielrbradley: We're breaking the loop so we'll immediately return the error. Looking at this now, I think it would be better to replace the `break` with an early return from within the loop for clarity.

provider/pkg/provider/provider.go
danielrbradley: We don't currently throw any errors from this method, so there might be cases where this is being hit right now but it silently continues. This just adds visibility of the issue for now, but leaves the behaviour unchanged (apart from if there's no body to continue with).

provider/pkg/provider/crud/crud.go
danielrbradley: Oh actually, can't return early because we still need to perform the required container adding.

thomas11 (approved):

PR #3495, opened 2024-08-06
Extend the custom handling of SiteConfig and deleteEmptyServerFarm to WebAppSlot

thomas11:
This PR extends the changes of both #3464 and #3004 from the WebApp to the WebAppSlot resource. Web apps and web apps slot are virtually identical and share much of their Azure spec as well, therefore, the fixes apply to them as well.

The only net-new code is in adjusting the tests and some refactoring, since both resources share the actual CRUD implementation.

The integration test is extended to create and verify both resource types.

### DIFF
--- a/provider/pkg/openapi/defaults/defaultResourcesState.go
+++ b/provider/pkg/openapi/defaults/defaultResourcesState.go
@@ -27,7 +27,9 @@ var defaultResourcesStateRaw = map[string]map[string]interface{}{
 	"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DBForMySQL/servers/{serverName}/configurations/{configurationName}": {
 		"source": "system-default",
 	},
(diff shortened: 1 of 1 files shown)


provider/pkg/openapi/defaults/defaultResourcesState.go
```diff
@@ -27,9 +27,7 @@ var defaultResourcesStateRaw = map[string]map[string]interface{}{
 	"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DBForMySQL/servers/{serverName}/configurations/{configurationName}": {
 		"source": "system-default",
 	},
-	"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DBForPostgreSql/flexibleServers/{serverName}/configurations/{configurationName}": {
-		"source": "system-default",
-	},
+	"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DBForPostgreSql/flexibleServers/{serverName}/configurations/{configurationName}": {},
```
danielrbradley, line 30: Why does the default resource state change here?

danielrbradley (approved):
All makes sense apart from the `DBForPostgreSql` default resource state change.

PR #3500, opened 2024-08-07
Upgrade specs & remove manual portal.Dashboard implementation

### LINKED ISSUE #3489: Workflow failure: nightly-sdk-generation
## Workflow Failure

- [2024-08-06T04:52:19.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/10260391164)

- [2024-08-06T06:32:24.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/10260391164)

- [2024-08-07T04:52:17.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/10278244747)

- [2024-08-08T04:52:50.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/10296057367)

danielrbradley:
Fixes #3489 

Revert #3123 which hard coded types for the dashboard resource. The specifications have been updated and improved which means the mixins caused the schema to be invalid. Our manual mixins are no longer needed as the specification appears to include the same level of detail. The structure of the types have changed slightly which will be a minor breaking change for existing usage which will require users to update their code.

## Migrating existing dashboard code

1. The top-level `lenses` and `metadata` properties are now wrapped inside a `properties` object.
2. The only supported "Part Type" is `Extension/HubsExtension/PartType/MarkdownPart`.
3. The `properties.lenses.parts.metadata.settings` property has now got concrete types instead of being a dictionary.
4. The names of some sub-types have changed.

### DIFF
--- a/examples/cs-dashboard/Program.cs
+++ b/examples/cs-dashboard/Program.cs
@@ -16,7 +16,7 @@ return await Pulumi.Deployment.RunAsync(() =>
     var dashboard = new Dashboard("dash", new()
     {
         ResourceGroupName = resourceGroup.Name,
-        Properties = new()
+        Properties = new DashboardPropertiesWithProvisioningStateArgs
         {
             Lenses = new[]
             {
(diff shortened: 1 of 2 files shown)


thomas11 (commented):
Some examples need fixing up. Which is also an example of the breaking changes.


Is there a way to fix the P1 in a non-breaking way, even if it's hacky, seeing that v3 is around the corner anyway where we can do the proper change?

danielrbradley:
> Some examples need fixing up. Which is also an example of the breaking changes.

Yup, looking at that now 👍 

> Is there a way to fix the P1 in a non-breaking way, even if it's hacky, seeing that v3 is around the corner anyway where we can do the proper change?

I think we'd have to replace the entire schema dashboard resource and invoke. I'm not even sure if we've got the mechanism in place right now to do that either - so would need to build more hacks into the generation process. I'm looking into the usage to understand how widely used this is.

danielrbradley:
The usage of the portal dashboard resource appears to be very low so we'll allow the breaking change in this instance.

PR #3503, opened 2024-08-08
Any addition or removal of properties to an object marked as ForceNew should cause its replacement

thomas11:
This PR was prompted by making a schema change that added `willReplaceOnChanges` to an existing property `parameters`.
```diff
--- a/provider/cmd/pulumi-resource-azure-native/schema.json
+++ b/provider/cmd/pulumi-resource-azure-native/schema.json
@@ -491175,11 +491175,12 @@
                 "parameters": {
                     "type": "object",
                     "additionalProperties": {
                         "type": "string"
                     },
-                    "description": "Gets or sets a list of job properties."
+                    "description": "Gets or sets a list of job properties.",
+                    "willReplaceOnChanges": true
                 },
```

The change had no effect when adding a new parameter. This is because our diff algorithm [only looks at changes of existing object members](https://github.com/pulumi/pulumi-azure-native/blob/tkappler/jobschedule-params/provider/pkg/provider/diff.go#L349).

### DIFF
--- a/provider/pkg/provider/diff_test.go
+++ b/provider/pkg/provider/diff_test.go
@@ -5,7 +5,6 @@ package provider
 import (
 	"encoding/json"
 	"fmt"
-	"strings"
 	"testing"
 	"unicode"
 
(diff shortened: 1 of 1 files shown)


danielrbradley (commented):
I _think_ this change looks about right, but the tests are pretty incomprehensible.

Perhaps rewriting the tests would be worthwhile to break it up into small feature-focused cases.

thomas11:
> I _think_ this change looks about right, but the tests are pretty incomprehensible.
> 
> Perhaps rewriting the tests would be worthwhile to break it up into small feature-focused cases.

Good call, done in the latest commit.

danielrbradley (approved):
Thank you for the test refactor - it looks much better now and I can see your exact scenario in there!

This also looks like a potentially valuable place to spend some more time checking coverage to highlight missing cases at some point.

Issue #2432, opened 2023-05-04
Invalid audience for Key Vault secret operations when using MSI


thomas11:
```
error: keyvault.BaseClient#DeleteSecret: Failure responding to request: StatusCode=401 -- Original Error: autorest/azure: Service returned an error. Status=401 Code="Unauthorized" Message="AKV10022: Invalid audience. Expected https://vault.azure.net, found: https://management.azure.com/."
```

When using Managed Identity (ARM_USE_MSI) with the Pulumi Automation API. See #2374 for the original report. That issue is about the same error but using az CLI authentication, which was fixed.

brucetan-displayr:
Having the same error, woudl be good if this is fixed

thomas11:
I took a more in-depth look at this and unfortunately, it's due to a shortcoming in a dependency that we use that cannot be upgraded quickly. Related to #1464. We will try to schedule this workstream soon.

thecodetinker:
Hello - we have just tried this with v2.75.0 and unfortunately got a different error. 

`error: keyvault.BaseClient#DeleteSecret: Failure responding to request: StatusCode=401 -- Original Error: autorest/azure: Service returned an error. Status=401 Code="Unauthorized" Message="AKV10022: Invalid audience. Expected cfa8b339-82a2-471a-a3c9-0fc0be7a4093, found: https://management.azure.com/."`

The GUID represents the Key Vault Service Principal - it is registered as the same in all tenants (as per [this doc](https://learn.microsoft.com/en-us/azure/key-vault/secrets/overview-storage-keys#service-principal-application-id)). Not sure if I need to open a new issue for this.

thomas11:
Hi @thecodetinker, sorry for the additional trouble here. This is curious, as we have a CI test for secrets that passes, and I also just verified it locally using `az` user authentication. Maybe it's another MSI-specific issue. I'll look into testing that as soon as I can.

In the meantime, would you open a new issue please? It's helpful for tracking.

Here's my minimal attempted repro.

```yaml
name: azn-2374-delete-secret
runtime: yaml

config:
  tenantId:
    type: string
  objectId:
    type: string

resources:
  resourceGroup:
    type: azure-native:resources:ResourceGroup

  keyVault:
    type: azure-native:keyvault:Vault
    properties:
      resourceGroupName: ${resourceGroup.name}
      properties:
        sku:
          name: standard
          family: A
        tenantId: ${tenantId}
        accessPolicies:
[...]

Issue #957, opened 2021-06-29
Allow use of Service Principal with az login


ianmccaul:
Part of our workflow in github does a number of things in Azure. All using a Service Principal. Prior to anything in the workflow running we run the azure/login action, which is effectively az login, but we use the service principal credentials to login there. If we do this and then run the pulumi/actions up it will fail due to it not honoring that auth context, despite it being completely valid. So now we have to extract the password/secret from the SP and store that as an additional secret because the azure/login takes the whole json object to auth. So now we have multiple secrets doing the same thing, and if we need to reset the secret we have to do it to all these secrets now.

## Affected feature

Allow the pulumi/actions or the pulumi cli to honor the az login context regardless if its a user or SP.

mikhailshilkov:
Thank you for opening this issue @ianmccaul 

The last note that I see is this:

> Unfortunately when authenticating as a Service Principal not all required information is available
As such instead we support authenticating as a Service Principal using the separate auth methods
for Client Certificate/Secret authentication.

I don't know which exact piece of information is missing. Do you mind running `az account show -o=json` while authenticated as a SP and pasting an obfuscated result here?

ianmccaul:
So I have to run the SP version inside the github workflow. The output is a bit odd because I think its trying to make some text a secret. But here it is:

![SP](https://user-images.githubusercontent.com/2401270/123965375-00f18d80-d97a-11eb-9a75-93cf3831518e.PNG)
 
Also, I ran this as me locally and got this:

![me](https://user-images.githubusercontent.com/2401270/123965422-0d75e600-d97a-11eb-9107-f21aab161fd5.PNG)

There is not much difference imo. I would assume tenantId is one of the things github is trying to make a secret. If that's the case the outputs are the same property wise, just user.type is different naturally. It really shouldnt matter what the type is, so Im confused what the difference is on the pulumi side of things.

EronWright:
The ability to use service principals with az login is fixed in Azure Native v3.8; enjoy!

PR #3536, opened 2024-08-22
Replace Autorest with azcore

thomas11:
This PR introduces a new implementation of `AzureClient` that's based on [azcore](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.14.0) instead of the deprecated autorest. It's part of [[Epic] Replace deprecated REST and auth packages in Azure Native #3576](https://github.com/pulumi/home/issues/3576). The initial commit is by @mikhailshilkov, thanks!

The PR is rather large. I've attempted to keep logically meaningful individual commits, so reviewing commit by commit might be helpful.

The high-level view of the changes is:
- Implement a new `AzureClient` that uses azcore. The `AzureClient` interface remains unchanged.
  - azcore implements polling for long-running requests internally, using the `Location` header. That isn't always a qualified URL, as I've observed, so there's new code to make the `Location` qualified.
- azcore wants to authenticate via the new azidentity package which we're not using yet. We do already have a bridge, however, that gets tokens from the current auth stack and serves them via azidentity's interface. We pass that bridge into the new Azure client.
- Using the new client is a small change in provider.go. Since regressions of this change could be very impactful and blocking, there's an escape hatch `PULUMI_USE_AUTOREST` environment variable to revert to the old client.
- "**Fix older bug in CanCreate**" and "**TokenCredential bridge: use correct environment**" are fixes for already existing bugs that I've found.

Integration testing is essentially provided by the existing test suite since all tests now use the new Azure client.

Unit testing is still rudimentary (although already better than the old client's) and will be extended.

### DIFF
--- a/provider/pkg/azure/client_azcore.go
+++ b/provider/pkg/azure/client_azcore.go
@@ -91,7 +91,19 @@ func (c *azCoreClient) initRequest(ctx context.Context, method, id string, query
 	c.setHeaders(req, method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch)
 
 	urlValues := MapToValues(queryParams)
-	// TODO,tkappler go-autorest WithQueryParameters in preparer.go does an additional query-unescape for each value - why?
+	// URL-unescape each value before encoding the URL (which encodes all values). Presumably, this
+	// prevents double-encoding. It's not clear to me that this is necessary but Autorest does it and
+	// we want to match behavior here as closely as possible.
+	for key, value := range urlValues {
+		for i := range value {
+			d, err := url.QueryUnescape(value[i])
+			if err != nil {
+				return nil, err
+			}
+			value[i] = d
+		}
+		urlValues[key] = value
+	}
 	req.Raw().URL.RawQuery = urlValues.Encode()
 
 	return req, nil
--- a/provider/pkg/azure/client_azcore_test.go
+++ b/provider/pkg/azure/client_azcore_test.go
@@ -65,10 +65,12 @@ func TestInitRequestQueryParams(t *testing.T) {
 	client := c.(*azCoreClient)
 
 	queryParams := map[string]any{
-		"api-version": "2022-09-01",
-		"bool":        true,
-		"int":         42,
-		"slice":       []string{"a", "b"},
+		"api-version":                 "2022-09-01",
+		"string_with_unescaped_slash": "a/path",
+		"string_with_escaped_slash":   "a%2Fpath",
+		"bool":                        true,
+		"int":                         42,
+		"slice":                       []string{"a", "b"},
 	}
 	req, err := client.initRequest(context.Background(), http.MethodGet, "/subscriptions/123", queryParams)
 	require.NoError(t, err)
@@ -76,6 +78,8 @@ func TestInitRequestQueryParams(t *testing.T) {
 	query := req.Raw().URL.Query()
 	assert.Len(t, query, len(queryParams))
 	assert.Equal(t, "2022-09-01", query.Get("api-version"))
+	assert.Equal(t, "a/path", query.Get("string_with_unescaped_slash"))
+	assert.Equal(t, "a/path", query.Get("string_with_escaped_slash"))
 	assert.Equal(t, "true", query.Get("bool"))
 	assert.Equal(t, "42", query.Get("int"))
 	assert.Equal(t, "a", query.Get("slice"))


provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func (c *azCoreClient) putOrPatch(ctx context.Context, method string, id string, bodyProps map[string]any,
+	queryParameters map[string]any, asyncStyle string,
+) (map[string]any, bool, error) {
+	if method != http.MethodPut && method != http.MethodPatch {
+		return nil, false, fmt.Errorf("method must be PUT or PATCH, got %s. Please report this issue", method)
+	}
+
+	apiVersion := queryParameters["api-version"].(string)
```
mikhailshilkov, line 142: Do we ignore all other query parameters?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+		q.Add("api-version", apiVersion)
+		locUrl.RawQuery = q.Encode()
+	}
+
+	headers.Set("Location", locUrl.String())
+}
+
+func (c *azCoreClient) Post(ctx context.Context, id string, bodyProps map[string]interface{},
+	queryParameters map[string]interface{}) (any, error) {
+	req, err := c.initRequest(ctx, http.MethodPost, id, queryParameters["api-version"].(string))
```
mikhailshilkov, line 237: Do we ignore all other queryParameters?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+
+	return &azCoreClient{
+		host:      azureCloud.Services[cloud.ResourceManager].Endpoint,
+		pipeline:  pipeline,
+		userAgent: userAgent,
+	}
+}
+
+func (c *azCoreClient) setHeaders(req *policy.Request) {
+	req.Raw().Header.Set("Accept", "application/json")
```
mikhailshilkov, line 45: Do we need to set content type to `application/json; charset=utf-8` as the old implementation does?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	if bodyProps != nil {
+		err = runtime.MarshalAsJSON(req, bodyProps)
+		if err != nil {
+			return nil, false, err
+		}
+	}
+
+	resp, err := c.pipeline.Do(req)
```
mikhailshilkov, line 156: Do we need something analogous to the old `DoRetryWithRegistration` decorator?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func NewAzCoreClient(tokenCredential azcore.TokenCredential, userAgent string, azureCloud cloud.Configuration, opts *arm.ClientOptions) AzureClient {
+	pipeline, err := armruntime.NewPipeline("pulumi-azure-native", version.Version, tokenCredential,
+		runtime.PipelineOptions{}, opts)
+	if err != nil {
+		panic(err)
+	}
+
+	return &azCoreClient{
```
mikhailshilkov, line 37: Do we have a mechanism to log requests analogous to `RequestInspector` in the old client?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+		return nil, err
+	}
+
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated) {
+		return nil, runtime.NewResponseError(resp)
+	}
+
+	var responseBody map[string]interface{}
+	err = runtime.UnmarshalAsJSON(resp, &responseBody)
+	return responseBody, err
```
mikhailshilkov, line 258: Should we do something in line with `handleResponseError` in the old client? E.g. when Azure returns a plain string with an error - it could be useful to show that error to the user, even though it's not JSON?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	return outputs, true, nil
+}
+
+// The azcore location header Poller validates that the location header is a valid absolute URL.
+// In some cases, like user-assigned identities, the location header in the PUT response is a
+// relative URL "/subscriptions/...", which we try to make absolute.
+// Also, add the api version query param if missing.
+func normalizeLocationHeader(host, apiVersion string, headers http.Header) {
```
mikhailshilkov, line 202: I'm a bit anxious about this part. Is there no 1st- or 3rd-party implementation that we could reuse? Based on my experience with autorest, LROs are a minefield and I'd really love to avoid maintaining any parts of logic for them. Why do we have to do normalization and others don't?

mikhailshilkov (commented):
This looks promising, also the code is quite easy to read! It's also nice how it's almost a drop-in replacement for the old client.

Some comments are below. I'm slightly concerned some of my findings don't result in test failures - it's hard to trust the test suite to catch all errors, it seems.

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+	locUrlStr := headers.Get("Location")
+	if locUrlStr == "" {
+		return
+	}
+
+	locUrl, err := url.Parse(locUrlStr)
+	if err != nil {
+		logging.V(3).Infof("Location header '%s' is not a valid URL: %v", locUrlStr, err)
```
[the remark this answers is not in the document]
thomas11: They won't. I picked 3 somewhat arbitrarily, as something not seen by default but lower than the 9 we usually use, to express something like a warning. We probably need to decide on a unified approach to this.

provider/pkg/azure/client_azcore.go
thomas11: I don't know why we need to do normalization. Without it, the polling requests fail with an error I unfortunately haven't captured.

It seems like an oversight in the azure-sdk-for-go.

[The RFC](https://httpwg.org/specs/rfc9110.html#field.location) shows that Location URLs can be relative:

> For example, a GET request generated for the URI reference "http://www.example.org/~tim" might result in a 303 (See Other) response containing the header field:
> Location: /People.html#tim
> which suggests that the user agent redirect to "http://www.example.org/People.html#tim"

Where as the poller implementation explicitly [asserts that the URL is absolute](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/internal/poller/util.go#L55).

provider/pkg/azure/client_azcore.go
thomas11: I'm not sure since integration tests pass without but I've added it now.

provider/pkg/azure/client_azcore.go
mikhailshilkov: Interesting... Worth opening an issue at https://github.com/Azure/azure-sdk-for-go/issues asking why, and maybe we can eventually get rid of it on our side?

provider/pkg/provider/provider.go
```diff
@@ -231,6 +234,24 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 	}, nil
 }
 
+func (k *azureNativeProvider) newAzureClient(armAuth autorest.Authorizer, tokenCred azCoreTokenCredential, userAgent string) azure.AzureClient {
+	if os.Getenv("PULUMI_USE_AUTOREST") == "true" {
```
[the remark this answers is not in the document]
thomas11: Interesting. Not in this provider - the existing variables like `ARM_USE_OIDC` are also checked against "true" only, and [the docs](https://www.pulumi.com/registry/packages/azure-native/installation-configuration/) say to set them to "true".

I'll punt on this one in this PR.

provider/pkg/provider/crud/crud.go
```diff
@@ -350,7 +350,7 @@ func (r *resourceCrudClient) ResponseBodyToSdkOutputs(response map[string]any) m
 }
 
 func (r *resourceCrudClient) CanCreate(ctx context.Context, id string) error {
-	return r.azureClient.CanCreate(ctx, id, r.res.ReadMethod, r.res.APIVersion, r.res.ReadMethod, r.res.Singleton, r.res.DefaultBody != nil, func(outputs map[string]any) bool {
+	return r.azureClient.CanCreate(ctx, id, r.res.ReadPath, r.res.APIVersion, r.res.ReadMethod, r.res.Singleton, r.res.DefaultBody != nil, func(outputs map[string]any) bool {
```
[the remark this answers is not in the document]
thomas11: That was a bug I spotted. The client actually expects `ReadPath` here. It only worked until now, mostly, because `ReadMethod` is almost always the empty string (defaulting to GET), so appending it to the URL path is a no-op. Similarly, `ReadPath` is almost always empty so we didn't miss it being appended.

I'm adding test coverage for this now.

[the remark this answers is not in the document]
mikhailshilkov: Ah! Now I finally see it, thank you.

mikhailshilkov:
On the rollout strategy, I suggest we should start with the opt-in strategy: use the old auth by default unless the flag is supplied. We can then seek opportunities for deliberate testing of the feature internally and externally with users that benefit from this change (what are those scenarios?)

provider/pkg/azure/client_azcore.go
thomas11: I was going to do that after reproducing the issue again to capture all the details, and now I cannot repro it anymore. Weird.

For now, I opened [an issue to ask about relative Location URLs in general](https://github.com/Azure/azure-sdk-for-go/issues/23385), and will leave the normalization code in as it won't do anything for absolute URLs.

provider/pkg/azure/client_azcore.go
thomas11: Both retries and [RP registration](https://github.com/Azure/azure-sdk-for-go/blob/sdk/azcore/v1.14.0/sdk/azcore/arm/runtime/pipeline.go#L42) are already part of the default azcore request pipeline.

I've just discovered that the retry backoffs are configured very differently, though, so I've adjusted them to be more like in Autorest and added a comment.

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,380 @@
...
+	// The new backoff sequence is       20s, 60s, 120s.
+	// It's notable, however, that the azcore default backoff sequence is only 1s, 3s, 7s. We might
+	// consider moving a bit towards a faster sequence, e.g. 10s, 30s, 70s.
+	opts.Retry.RetryDelay = 20 * time.Second
+	opts.Retry.MaxRetryDelay = 120 * time.Second
+
+	pipeline, err := armruntime.NewPipeline("pulumi-azure-native", version.Version, tokenCredential,
+		runtime.PipelineOptions{}, opts)
+	if err != nil {
+		panic(err)
```
danielrbradley, line 58: I think this should be fairly easy to return the error back from this fuction all the way back through the call stack rather than panicing. Might be useful for testing invalid configurations being caught.

provider/pkg/provider/auth.go
```diff
@@ -339,7 +339,7 @@ func (cred azCoreTokenCredential) GetToken(ctx context.Context, options policy.T
 	if err != nil {
 		return azcore.AccessToken{}, err
 	}
-	token, err := cred.p.getOAuthToken(ctx, authConfig, string(environments.ResourceManagerPublic.Endpoint))
+	token, err := cred.p.getOAuthToken(ctx, authConfig, cred.p.environment.ResourceManagerEndpoint)
```
danielrbradley, line 342: Why was the other environment wrong?

provider/pkg/provider/provider.go
```diff
@@ -84,8 +86,8 @@ func makeProvider(host *provider.HostClient, name, version string, schemaBytes [
 
 	// Return the new provider
 	p := &azureNativeProvider{
-		// This client will be regnerated with correct environment and authorizer in Configure.
-		azureClient:   azure.NewAzureClient(azureEnv.PublicCloud, nil, azure.BuildUserAgent(PulumiPartnerID)),
+		// This client will be regenerated with correct environment and authorizer in Configure.
+		azureClient:   azure.NewAzCoreClient(nil, azure.BuildUserAgent(PulumiPartnerID), cloud.AzurePublic, nil),
```
danielrbradley, line 90: Not for this PR, but perhaps `azureClient` should be a pointer and should not be created until we've called configure? Is there some reason we need a default client implementation for use before we've configured the provider that you know of?

provider/pkg/provider/crud/crud_test.go
```diff
@@ -95,3 +102,105 @@ func TestPathParamsErrorHandling(t *testing.T) {
 		assert.Equal(t, "/path/deepVal", id)
 	})
 }
+
+func TestCanCreate_RequestUrls(t *testing.T) {
```
danielrbradley, line 106: Running these tests locally was quite slow. Would it be safe to mark these with `t.Parallel()` to reduce the overall wait time?

provider/pkg/azure/client_azcore_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	t.Run("NoLocationHeader", func(t *testing.T) {
+		headers := http.Header{}
+		normalizeLocationHeader(host, apiVersion, headers)
+		assert.Empty(t, headers.Get("Location"))
+	})
+}
+
+func TestPOST(t *testing.T) {
```
danielrbradley, line 61: These are neat for potentially reproducing issues in the future. Shall we add similar for Head, Get, Delete, Put & Patch?

danielrbradley (approved):
Overall looks great to me. Couple of minor questions inline.

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,337 @@
...
+
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusNoContent) {
+		return runtime.NewResponseError(resp)
+	}
+	return nil
+}
+
+// CanCreate asserts that a resource with a given ID and API version can be created
+// or returns an error otherwise.
+func (c *azCoreClient) CanCreate(ctx context.Context, id, path, apiVersion, readMethod string,
```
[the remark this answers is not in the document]
thomas11: Not sure if it wasn't used in a previous iteration but it's used now.

provider/pkg/provider/crud/crud_test.go
thomas11: I found and fixed the slowness (retries).

provider/pkg/provider/auth.go
thomas11: `environments.ResourceManagerPublic` is the resource manager endpoint of the public Azure cloud. When a user configured their cloud/environment to be Government or China, it would have been ignored here.

provider/pkg/provider/provider.go
thomas11: The only reason I know is that `azureClient` would then be a pointer to an interface which is annoying to use: `(*k.azureClient).Get`.

provider/pkg/azure/client_azcore_test.go
thomas11: Done, although covering only error handling based on the initial HTTP status code. There's still room for adding coverage of polling.

thomas11:
> We can then seek opportunities for deliberate testing of the feature internally and externally with users that benefit from this change (what are those scenarios?)

It might be a hard sell as the benefits are all internal - get off an unsupported dependency, receive security updates and bug fixes, etc.

The next step, basing the auth stack on the new and supported azidentity, should close some user-visible bugs: #2432, #2734, maybe #957.

mikhailshilkov:
> It might be a hard sell as the benefits are all internal - get off an unsupported dependency, receive security updates and bug fixes, etc.

We'll have to keep the dependency for a while regardless of the default value. It there's little immediate user value (upside), we should probably be even more careful not to break anyone (downside).

> The next step, basing the auth stack on the new and supported azidentity, should close some user-visible bugs: https://github.com/pulumi/pulumi-azure-native/issues/2432, https://github.com/pulumi/pulumi-azure-native/issues/2734, maybe https://github.com/pulumi/pulumi-azure-native/issues/957.

Sounds like a great motivation to get somebody to try the new auth after those are shipped.

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,426 @@
...
+	if err != nil {
+		return err
+	}
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusAccepted, http.StatusNoContent, http.StatusNotFound) {
+		return runtime.NewResponseError(resp)
+	}
+
+	// Some APIs are explicitly marked `x-ms-long-running-operation` and we should only do the
+	// poll for the deletion result in that case.
+	if asyncStyle != "" {
```
mikhailshilkov, line 140: I think I don't see test coverage for this branch - am I missing it?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,426 @@
...
+
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusNoContent) {
+		return runtime.NewResponseError(resp)
+	}
+	return nil
+}
+
+// CanCreate asserts that a resource with a given ID and API version can be created
+// or returns an error otherwise.
+func (c *azCoreClient) CanCreate(ctx context.Context, id, path, apiVersion, readMethod string,
```
mikhailshilkov, line 320: Similarly, do we need tests for CanCreate?

provider/pkg/azure/client_azcore.go
thomas11: added some

provider/pkg/azure/client_azcore.go
thomas11: added

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,432 @@
...
+	return responseBody, nil
+}
+
+func (c *azCoreClient) Delete(ctx context.Context, id, apiVersion, asyncStyle string, queryParams map[string]any,
+) error {
+	queryParameters := map[string]interface{}{
+		"api-version": apiVersion,
+	}
+	for k, v := range queryParams {
+		queryParameters[k] = v
```
mikhailshilkov, line 128: It looks like we never pass those in tests

provider/pkg/azure/client_azcore.go
mikhailshilkov: I don't see it? There is a test for Put but I don't see one for Delete.

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,432 @@
...
+	return err
+}
+
+func (c *azCoreClient) Put(ctx context.Context, id string, bodyProps map[string]interface{},
+	queryParameters map[string]interface{}, asyncStyle string,
+) (map[string]interface{}, bool, error) {
+	return c.putOrPatch(ctx, http.MethodPut, id, bodyProps, queryParameters, asyncStyle)
+}
+
+func (c *azCoreClient) Patch(ctx context.Context, id string, bodyProps map[string]interface{},
```
mikhailshilkov, line 173: No tests call Patch?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,432 @@
...
+	queryParameters map[string]interface{}, asyncStyle string,
+) (map[string]interface{}, bool, error) {
+	return c.putOrPatch(ctx, http.MethodPatch, id, bodyProps, queryParameters, asyncStyle)
+}
+
+func (c *azCoreClient) putOrPatch(ctx context.Context, method string, id string, bodyProps map[string]any,
```
mikhailshilkov, line 179: `bodyProps` never passed in tests?

provider/pkg/azure/client_azcore.go
```diff
@@ -0,0 +1,432 @@
...
+func (c *azCoreClient) initRequest(ctx context.Context, method, id string, queryParams map[string]any) (*policy.Request, error) {
+	req, err := runtime.NewRequest(ctx, method, runtime.JoinPaths(c.host, id))
+	if err != nil {
+		return nil, err
+	}
+
+	c.setHeaders(req, method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch)
+
+	urlValues := MapToValues(queryParams)
+	// TODO,tkappler go-autorest WithQueryParameters in preparer.go does an additional query-unescape for each value - why?
```
mikhailshilkov, line 94: What's the plan about this TODO? Resolve before merging or open an issue?

mikhailshilkov (approved):
Looks great! Just a few nits here and there.

What is the rollout plan? Something like
```
1. Merge, ship, make sure everything is quiet.
2. Set PULUMI_USE_AUTOREST=false in all CIs, make sure everything works.
3. Advertize PULUMI_USE_AUTOREST=false in case somebody wants to give it a try.
4. Flip the flag in 3.0.
```
?

provider/pkg/azure/client_azcore.go
thomas11: you're right, added, thanks for the close look!

provider/pkg/azure/client_azcore.go
thomas11: Not directly, but it's 99% identical to PUT. I added one as a basic sanity check.

provider/pkg/azure/client_azcore.go
thomas11: Is hoping that one of the clever reviewers can answer the question a valid plan? :)

thomas11: The `Encode()` method called just after query-escapes the values. So I assume un-escaping them first is to prevent double-escaping, and is a no-op otherwise.

danielrbradley (approved):

PR #3559, opened 2024-09-05
Add exception for CIAMResourceSKU tier schemagen

thomas11:
[CIAMResourceSKU](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cpim/resource-manager/Microsoft.AzureActiveDirectory/preview/2023-05-17-preview/externalIdentities.json#L1506) has two properties `name` and `tier`. The latter is incorrectly annotated with `x-ms-client-flatten`, causing it to be missing from `CIAMResourceSKU` in the Pulumi schema. See upstream issue https://github.com/Azure/azure-rest-api-specs/issues/30443 for more detail.

This PR adds an unfortunate hard-coded exception to unblock the `CIAMResource` resource.

Fixes #3556


danielrbradley (approved):
I think the workaround is reasonable - pretty self-contained and documented which is the best we can hope for here. Is there any way to also scope this to the module/version at all or is that not available in this context?

PR #3600, opened 2024-09-30
Unblock SDK gen with required properties edits

### LINKED ISSUE #3598: Workflow failure: nightly-sdk-generation
## Workflow Failure

- [2024-09-27T04:52:51.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/11064649692)

- [2024-09-30T04:48:14.000Z](https://github.com/pulumi/pulumi-azure-native/actions/runs/11099503074)

thomas11:
- **Add two more required property edits for conflicting types**
- **Update spec and regenerate**
- **Add section to playbook for this kind of conflict**

Fixes #3598


danielrbradley (approved):
This type conflict looks okay to just remove the required properties from 👍 

The alternative approach here would have been to exclude the new version of network from the current default version so they'd only be available via the explicit version until the next major release - which then avoids the conflict altogether.

Issue #3493, opened 2024-08-06
Replace go-azure-helpers with azidentity


thomas11:
See the [Epic](https://github.com/pulumi/home/issues/3576) for background.

- [x] Support oidcTokenFilePath
- [x] Test explicit user-assigned identity. This probably requires updating the go-azure-in-azure test to use the provider build under test instead of the latest release.
- [x] investigate multi-tenant auth

mjeffryes:
I think this is done now

PR #3630, opened 2024-10-07
Replace go-azure-helpers with azidentity

thomas11:
# Overview

This PR implements #3493 which is part of Epic https://github.com/pulumi/home/issues/3576, Replace deprecated REST and auth packages in Azure Native.

The goal is to replace outdated and deprecated libraries that receive no bug fixes, block us from fixing some issues, and might pose security risks.

# Parts

The legacy authentication setup is in auth.go. This PR adds auth_azidentity.go in parallel, with the same purpose: read all auth-related configuration, decide on the correct authentication method, and initialize it.

The core library used is Azure's official [azidentity](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity). It has various `FooCredential` types that all return an `azcore.TokenCredential`, abstracting the authentication method being used.

Thanks to the existing AzureClient abstraction, which has an azcore implementation, we can simply pass the new TokenCredential from auth_azidentity.go there without further changes.

One place where we have to plug in the new auth backend manually is the `getClientToken` RPC method in provider.go.

# Rollout

For the sake of caution, the new authentication backend is off by default, behind feature flag `PULUMI_USE_LEGACY_AUTH` (defaulting to true). It's not 100% clear yet how and when we'll decide to turn it on.

# Testing

New unit tests should be self-explanatory.

Existing integration tests inherently cover authentication, as long as they're run with the feature flag. For that purpose, there's a new GH workflow [azcore-scheduled.yml](https://github.com/pulumi/pulumi-azure-native/actions/workflows/azcore-scheduled.yml). It runs every night and can be run on demand via workflow dispatch as well. **Note that this means that the tests that are part of this PR's checks do not run on the new backend, only azcore_scheduled does.**

I expanded the existing go-azure-in-azure test to use the provider binary under test even on the remote VM, and to create two user-managed identities. That forces the test program to configure which one should be used, increasing test coverage of user-managed identities.

A missing chunk of test coverage I'm aware of is that there's no integration test using the `az` CLI for authentication. It's quite tricky as the only interesting authentication method here is interactive user authentication via the Azure portal. The CLI can also do service principal authentication but we already test that. 

# Missing parts

1. The CLI integration test mentioned above.
2. Our two custom Azure storage resources, StorageAccountStaticWebsite and Blob, are built using an obsolete Azure SDK that only works with the legacy authentication stack. I've started on updating and converting it but kept it out of this PR since it can be committed separately. Unless that's completed, the legacy auth stack cannot be removed, though.

# Review guide

1. Start at provider(_test).go. The changes are the introduction of a new feature flag, and the related logic to choose the correct authentication package.
2. auth_azidentity(_test).go are the meat of the PR. That's the new azidentity-based authentication stack. `func (k *azureNativeProvider) newTokenCredential()` is the main entry point. (Would be nice if authentication was in its own package but with the removal of the legacy stack it'll shrink.)
3. `examples/azure-native-sdk-v2/go-azure-in-azure/main.go` is the integration test that runs Pulumi on an Azure VM to test managed identity authentication. It was extended to properly test user-assigned identities.
4. The changes in `.github` go together with the ones in `examples/examples_nodejs_keyvault_test.go`. They constitute a new integration test that uses `az` CLI authentication.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -22,14 +22,10 @@ on:
         type: number
         description: The number of days for which we retain assets
         default: 90
-      use_autorest:
+      use_azcore:
         type: boolean
-        description: Whether to use autorest, not azcore
-        default: true
-      use_legacy_auth:
-        type: boolean
-        description: Whether to use go-azure-helpers, not azidentity
-        default: true
+        description: Whether to use the newer azcore+azidentity backend for REST and auth, or the older autorest
+        default: false
 
 env:
   GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
@@ -49,8 +45,8 @@ env:
   ARM_SUBSCRIPTION_ID: 0282681f-7a9e-424b-80b2-96babd57a8a1
   ARM_TENANT_ID: 706143bc-e1d4-4593-aee2-c9dc60ab9be7
   PULUMI_API: https://api.pulumi-staging.io
-  PULUMI_USE_AUTOREST: ${{ inputs.use_autorest }}
-  PULUMI_USE_LEGACY_AUTH: ${{ inputs.use_legacy_auth }}
+  # Feature toggle that's read in provider.go enableAzcoreBackend()
+  PULUMI_ENABLE_AZCORE_BACKEND: ${{ inputs.use_azcore }}
   # This is the content of a ~/.azure/ folder, zipped and base64-encoded, for CLI auth.
   # If/when the contained refresh token expires, someone with access to our subscription needs to
   # `az login` on their own computer and repeat the steps below.
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -217,14 +217,14 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 	userAgent := k.getUserAgent()
 
 	var credential azcore.TokenCredential
-	if useLegacyAuth() {
-		logging.V(9).Infof("Using legacy authentication")
-		credential = azCoreTokenCredential{p: k}
-	} else {
+	if enableAzcoreBackend() {
 		credential, err = k.newTokenCredential()
 		if err != nil {
 			return nil, fmt.Errorf("creating Pulumi auth credential: %w", err)
 		}
+	} else {
+		logging.V(9).Infof("Using legacy authentication")
+		credential = azCoreTokenCredential{p: k}
 	}
 
 	k.azureClient, err = k.newAzureClient(resourceManagerAuth, credential, userAgent)
@@ -246,18 +246,12 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 }
 
 func (k *azureNativeProvider) newAzureClient(armAuth autorest.Authorizer, tokenCred azcore.TokenCredential, userAgent string) (azure.AzureClient, error) {
-	useAutorest := os.Getenv("PULUMI_USE_AUTOREST") != "false"
-
-	if !useAutorest && useLegacyAuth() {
-		return nil, errors.New("PULUMI_USE_LEGACY_AUTH=true requires PULUMI_USE_AUTOREST=true")
+	if enableAzcoreBackend() {
+		logging.V(9).Infof("AzureClient: using azcore and azidentity")
+		return azure.NewAzCoreClient(tokenCred, userAgent, azure.GetCloudByName(k.environment.Name), nil)
 	}
-
-	if useAutorest {
-		logging.V(9).Infof("AzureClient: using autorest")
-		return azure.NewAzureClient(k.environment, armAuth, userAgent), nil
-	}
-	logging.V(9).Infof("AzureClient: using azCore")
-	return azure.NewAzCoreClient(tokenCred, userAgent, azure.GetCloudByName(k.environment.Name), nil)
+	logging.V(9).Infof("AzureClient: using autorest")
+	return azure.NewAzureClient(k.environment, armAuth, userAgent), nil
 }
 
 // Invoke dynamically executes a built-in function in the provider.
@@ -372,19 +366,20 @@ func (k *azureNativeProvider) Invoke(ctx context.Context, req *rpc.InvokeRequest
 func (k *azureNativeProvider) getClientToken(ctx context.Context, authConfig *authConfig, endpointArg resource.PropertyValue) (string, error) {
 	endpoint := k.tokenEndpoint(endpointArg)
 
-	if useLegacyAuth() {
-		return k.getOAuthToken(ctx, authConfig, endpoint)
+	if enableAzcoreBackend() {
+		cred, err := k.newTokenCredential()
+		if err != nil {
+			return "", err
+		}
+		t, err := cred.GetToken(ctx, tokenRequestOpts(endpoint))
+		if err != nil {
+			return "", err
+		}
+		return t.Token, nil
 	}
 
-	cred, err := k.newTokenCredential()
-	if err != nil {
-		return "", err
-	}
-	t, err := cred.GetToken(ctx, tokenRequestOpts(endpoint))
-	if err != nil {
-		return "", err
-	}
-	return t.Token, nil
+	// legacy autorest/go-azure-helpers auth
+	return k.getOAuthToken(ctx, authConfig, endpoint)
 }
 
 // Returns the Azure endpoint where tokens can be requested. If the argument is not null or empty,
@@ -1617,6 +1612,13 @@ func (k *azureNativeProvider) autorestEnvToHamiltonEnv() environments.Environmen
 	}
 }
 
-func useLegacyAuth() bool {
-	return os.Getenv("PULUMI_USE_LEGACY_AUTH") != "false"
+// enableAzcoreBackend is a feature toggle that returns true if the newer backend using azcore and
+// azidentity for REST and authentication should be used. Otherwise, the previous autorest backend
+// is used.
+// Tracked in epic #3576, the new backend was added to upgrade from unmaintained libraries that
+// don't receive security and other updates. It uses the latest official Azure packages.
+// The new backend is gated behind this feature toggle to allow enabling it selectively,
+// limiting the blast radius of regressions. It's enabled in the daily CI workflow azcore-scheduled.
+func enableAzcoreBackend() bool {
+	return os.Getenv("PULUMI_ENABLE_AZCORE_BACKEND") == "true"
 }
--- a/provider/pkg/provider/provider_test.go
+++ b/provider/pkg/provider/provider_test.go
@@ -401,40 +401,24 @@ func TestUsesCorrectAzureClient(t *testing.T) {
 	p := azureNativeProvider{}
 
 	t.Run("default", func(t *testing.T) {
-		t.Setenv("PULUMI_USE_AUTOREST", "")
-		t.Setenv("PULUMI_USE_LEGACY_AUTH", "")
+		t.Setenv("PULUMI_ENABLE_AZCORE_BACKEND", "")
 		client, err := p.newAzureClient(nil, &fake.TokenCredential{}, "pulumi")
 		require.NoError(t, err)
 		assert.Equal(t, "azureClientImpl", reflect.TypeOf(client).Elem().Name())
 	})
 
-	t.Run("Autorest enabled", func(t *testing.T) {
-		t.Setenv("PULUMI_USE_AUTOREST", "true")
+	t.Run("Autorest and legacy auth disabled explicitly", func(t *testing.T) {
+		t.Setenv("PULUMI_ENABLE_AZCORE_BACKEND", "false")
 		client, err := p.newAzureClient(nil, &fake.TokenCredential{}, "pulumi")
 		require.NoError(t, err)
 		assert.Equal(t, "azureClientImpl", reflect.TypeOf(client).Elem().Name())
 	})
 
-	t.Run("Autorest and legacy auth disabled", func(t *testing.T) {
-		t.Setenv("PULUMI_USE_AUTOREST", "false")
-		t.Setenv("PULUMI_USE_LEGACY_AUTH", "false")
+	t.Run("Azcore enabled", func(t *testing.T) {
+		t.Setenv("PULUMI_ENABLE_AZCORE_BACKEND", "true")
 		client, err := p.newAzureClient(nil, &fake.TokenCredential{}, "pulumi")
 		require.NoError(t, err)
 		assert.Equal(t, "azCoreClient", reflect.TypeOf(client).Elem().Name())
-	})
-
-	t.Run("Legacy auth without autorest is an error", func(t *testing.T) {
-		t.Setenv("PULUMI_USE_AUTOREST", "false")
-		t.Setenv("PULUMI_USE_LEGACY_AUTH", "")
-		_, err := p.newAzureClient(nil, &fake.TokenCredential{}, "pulumi")
-		require.Error(t, err)
-	})
-
-	t.Run("Legacy auth without autorest is an error, explicit", func(t *testing.T) {
-		t.Setenv("PULUMI_USE_AUTOREST", "false")
-		t.Setenv("PULUMI_USE_LEGACY_AUTH", "true")
-		_, err := p.newAzureClient(nil, &fake.TokenCredential{}, "pulumi")
-		require.Error(t, err)
 	})
 }
 
--- a/.github/workflows/azcore-scheduled.yml
+++ b/.github/workflows/azcore-scheduled.yml
@@ -19,5 +19,4 @@ jobs:
       version: ${{ needs.version.outputs.version }}
       short_test: false
       retention_days: 7
-      use_autorest: false
-      use_legacy_auth: false
+      use_azcore: true


examples/azure-native-sdk-v2/go-azure-in-azure/main.go
```diff
@@ -247,30 +260,62 @@ func main() {
 			return err
 		}
 
+		// Copy the provider binary under test (the one on PATH) to the VM.
```
thomas11, line 263: This works only if the test machine and the remote VM have the same arch and OS. That's the case in CI.

danielrbradley (approved):
At an overview the approach looks good. With the feature switch and existing test coverage, it looks quite safe to go in. Could probably go through the net-new code with a finer toothcomb, but overall the logic looks to mirror the existing code.

It might be worth adding some extra provider e2e tests which the new auth being used.

thomas11:
> It might be worth adding some extra provider e2e tests which the new auth being used.

The azcore-scheduled workflow runs _all_ e2e tests on the new auth. Do you have any scenarios in mind that are not covered?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+	clientSecret       string
+	clientCertPath     string
+	clientCertPassword string
+
+	useMsi      bool
+	msiEndpoint string
+}
+
+// getAuthConfig collects auth-related configuration from Pulumi config and environment variables
+func (k *azureNativeProvider) getAuthConfig3() (*authConfiguration, error) {
```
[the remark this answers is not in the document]
thomas11: Oops, that's an artifact from when I was trying out different things. Will rename.

provider/pkg/provider/provider_test.go
```diff
@@ -413,12 +414,27 @@ func TestUsesCorrectAzureClient(t *testing.T) {
...
 		client, err := p.newAzureClient(nil, &fake.TokenCredential{}, "pulumi")
 		require.NoError(t, err)
 		assert.Equal(t, "azCoreClient", reflect.TypeOf(client).Elem().Name())
 	})
+
+	t.Run("Legacy auth without autorest is an error", func(t *testing.T) {
+		t.Setenv("PULUMI_USE_AUTOREST", "false")
+		t.Setenv("PULUMI_USE_LEGACY_AUTH", "")
+		_, err := p.newAzureClient(nil, &fake.TokenCredential{}, "pulumi")
+		require.Error(t, err)
```
mikhailshilkov, line 429: Isn't this problematic? Say, I set `PULUMI_USE_AUTOREST=false` today before this PR merges. It works with Az Core and I'm happy. Then I upgrade to the new version with the new auth method and it breaks because I don't have `PULUMI_USE_LEGACY_AUTH` set?

provider/pkg/provider/provider.go
```diff
@@ -237,12 +247,18 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 }
 
 func (k *azureNativeProvider) newAzureClient(armAuth autorest.Authorizer, tokenCred azcore.TokenCredential, userAgent string) (azure.AzureClient, error) {
-	if os.Getenv("PULUMI_USE_AUTOREST") == "false" {
-		logging.V(9).Infof("AzureClient: using azCore")
-		return azure.NewAzCoreClient(tokenCred, userAgent, k.getAzureCloud(), nil)
+	useAutorest := os.Getenv("PULUMI_USE_AUTOREST") != "false"
+
+	if !useAutorest && useLegacyAuth() {
```
mikhailshilkov, line 252: Will this combination not work, or why do we want to disallow it?

provider/pkg/provider/provider_test.go
```diff
@@ -413,12 +414,27 @@ func TestUsesCorrectAzureClient(t *testing.T) {
...
+	})
+
+	t.Run("Legacy auth without autorest is an error, explicit", func(t *testing.T) {
+		t.Setenv("PULUMI_USE_AUTOREST", "false")
+		t.Setenv("PULUMI_USE_LEGACY_AUTH", "true")
+		_, err := p.newAzureClient(nil, &fake.TokenCredential{}, "pulumi")
+		require.Error(t, err)
+	})
```
mikhailshilkov, line 437: Does true/false work fine? Will there be reasonable use cases for it? Do we need a test for it? Will we have integration tests for that combination? If any answer is "no", should we just have a single env var that switches both client and auth?

examples/examples_nodejs_keyvault_test.go
```diff
@@ -80,9 +82,39 @@ func TestAccKeyVaultTs_ClientCert(t *testing.T) {
...
+				"ARM_CLIENT_SECRET=",
+				"ARM_CLIENT_CERTIFICATE_PATH=",
+				"ARM_USE_MSI=false",
+				"ARM_USE_OIDC=false",
+			},
+		})
+
+	integration.ProgramTest(t, &test)
+
+	_ = os.Rename(filepath.Join(usr.HomeDir, ".azure"), filepath.Join(usr.HomeDir, ".azure.tmp"))
```
mikhailshilkov, line 119: Why is this important? Should it be in `defer()`?

.github/workflows/build-test.yml
```diff
@@ -202,6 +221,15 @@ jobs:
         run: |
           echo "${{ secrets.ARM_CLIENT_CERTIFICATE }}" | base64 -d > "${{ runner.temp }}/azure-client-certificate.pfx"
 
+      - name: Write .azure.tmp folder
```
mikhailshilkov, line 224: Why can't it be `.azure`?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+	return newSingleMethodAuthCredential(authConf)
+}
+
+// newSingleMethodAuthCredential creates an azcore.TokenCredential. Depending on the given authConfiguration, it is
+// backed by one of several authentication methods such as Service Principal, OIDC, Managed Identity, or Azure CLI.
+//
+// Note: this function's behavior is written to match the behavior of
+// "github.com/hashicorp/go-azure-helpers/authentication".Builder.Build() in some ways, to minimize changes in provider
+// behavior when upgrading authentication dependencies from go-azure-helpers to azidentity. Namely:
+//   - The order in which the the different authentication methods are attempted is the same.
```
mikhailshilkov, line 39: Let's document all authentication methods in order here in the comment?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+		AdditionallyAllowedTenants: authConf.auxTenants, // usually empty which is fine
+	}
+	cli, err := azidentity.NewAzureCLICredential(options)
+	if err == nil {
+		return cli, nil
+	}
+	return nil, errors.Errorf("Failed to find any valid credentials")
+}
+
+func newOidcCredential(authConf *authConfiguration) (azcore.TokenCredential, error) {
```
mikhailshilkov, line 94: I haven't reviewed this function... just have no idea what is going on here. Please shout if you think we must do that.

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	var pwBytes []byte
+	if certPassword != "" {
+		pwBytes = []byte(certPassword)
+	}
+
+	certs, key, err := azidentity.ParseCertificates(cert, pwBytes)
+	if err != nil || len(certs) == 0 {
+		return nil, nil, errors.Wrapf(err, "failed to parse certificate from %s", certPath)
```
mikhailshilkov, line 182: if `err` is `nil` but `certs` is empty, `errors.Wrapf` will return `nil`, so we will return three nils. Sounds wrong?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+	auxTenantsString := k.getConfig("auxiliaryTenantIds", "ARM_AUXILIARY_TENANT_IDS")
+	var auxTenants []string
+	if auxTenantsString != "" {
+		err := json.Unmarshal([]byte(auxTenantsString), &auxTenants)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to unmarshal '%s' as Auxiliary Tenants", auxTenantsString)
+		}
+	}
+
+	envName := k.getConfig("environment", "ARM_ENVIRONMENT")
```
mikhailshilkov, line 221: Nit: call it `cloud` from the get-go?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+	if envName == "" {
+		envName = "public"
+	}
+
+	return &authConfiguration{
+		subscriptionId:     k.getConfig("subscriptionId", "ARM_SUBSCRIPTION_ID"),
+		clientId:           k.getConfig("clientId", "ARM_CLIENT_ID"),
+		tenantId:           k.getConfig("tenantId", "ARM_TENANT_ID"),
+		auxTenants:         auxTenants,
+		cloud:              envName,
```
mikhailshilkov, line 231: Where is `cloud` used?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+			return nil, errors.Wrapf(err, "failed to unmarshal '%s' as Auxiliary Tenants", auxTenantsString)
+		}
+	}
+
+	envName := k.getConfig("environment", "ARM_ENVIRONMENT")
+	if envName == "" {
+		envName = "public"
+	}
+
+	return &authConfiguration{
```
mikhailshilkov, line 226: I have no idea what `ARM_METADATA_HOSTNAME` from the old implementation is - but do we need to support it?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+		subscriptionId:     k.getConfig("subscriptionId", "ARM_SUBSCRIPTION_ID"),
```
mikhailshilkov, line 227: Where is `subscriptionId` used?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+		clientId:           k.getConfig("clientId", "ARM_CLIENT_ID"),
+		tenantId:           k.getConfig("tenantId", "ARM_TENANT_ID"),
+		auxTenants:         auxTenants,
+		cloud:              envName,
+		clientSecret:       k.getConfig("clientSecret", "ARM_CLIENT_SECRET"),
+		clientCertPath:     k.getConfig("clientCertificatePath", "ARM_CLIENT_CERTIFICATE_PATH"),
+		clientCertPassword: k.getConfig("clientCertificatePassword", "ARM_CLIENT_CERTIFICATE_PASSWORD"),
+
+		useMsi:      k.getConfig("useMsi", "ARM_USE_MSI") == "true",
+		msiEndpoint: k.getConfig("msiEndpoint", "ARM_MSI_ENDPOINT"),
```
mikhailshilkov, line 237: Where is `msiEndpoint` used?

provider/pkg/provider/auth_azidentity_test.go
```diff
@@ -0,0 +1,270 @@
...
+			},
+		} {
+			_, err := newSingleMethodAuthCredential(conf)
+			require.Error(t, err)
+		}
+	})
+
+	t.Run("MSI", func(t *testing.T) {
+		conf := &authConfiguration{
+			useMsi: true,
```
mikhailshilkov, line 264: Do we want test cases with clientId and MSI endpoint?

provider/pkg/provider/auth_azidentity_test.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+		cred, err := newSingleMethodAuthCredential(conf)
+		require.NoError(t, err)
+		require.IsType(t, &azidentity.ManagedIdentityCredential{}, cred)
+	})
```
mikhailshilkov, line 269: Will Azure CLI auth fail before we can inspect the result, is that why that test is missing?

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		return oidcTokenCredential(string(token))
+	}
+
+	// In this case, we need to obtain the OIDC token first from the configured endpoint.
+	if authConf.oidcTokenRequestUrl != "" && authConf.oidcTokenRequestToken != "" {
+		return azidentity.NewClientAssertionCredential(
+			authConf.tenantId,
+			authConf.clientId,
+			func(ctx context.Context) (string, error) {
```
mikhailshilkov, line 122: I think there are no tests for this function. Can probably break it down and have partially tested?

provider/pkg/provider/provider.go
```diff
@@ -301,9 +317,25 @@ func (k *azureNativeProvider) Invoke(ctx context.Context, req *rpc.InvokeRequest
 		if endpointArg := args["endpoint"]; endpointArg.HasValue() && endpointArg.IsString() {
 			endpoint = endpointArg.StringValue()
 		}
-		token, err := k.getOAuthToken(ctx, auth, endpoint)
-		if err != nil {
-			return nil, err
+
+		var token string
```
mikhailshilkov, line 321: Time to add some test coverage for this code?

mikhailshilkov (commented):
The approach makes sense to me, nice separation of two auth method. I tried to review all the details except OIDC peculiarities.

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+//     the next method.
+//   - Auxiliary or additional tenants are supported for SP with client secret and CLI authentication, not for others.
+func newSingleMethodAuthCredential(authConf *authConfiguration) (azcore.TokenCredential, error) {
+	if authConf.clientCertPath != "" {
+		logging.V(9).Infof("[auth] Using SP with client certificate credential")
+		certs, key, err := readCertificate(authConf.clientCertPath, authConf.clientCertPassword)
+		if err != nil {
+			return nil, err
+		}
+		return azidentity.NewClientCertificateCredential(authConf.tenantId, authConf.clientId, certs, key, nil)
```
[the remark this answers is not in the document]
thomas11: The azidentity package checks these kind of validations internally, e.g. `func validTenantID(tenantID string) bool` in azidentity@v1.7.0/azidentity.go.

provider/pkg/provider/provider.go
```diff
@@ -301,9 +317,25 @@ func (k *azureNativeProvider) Invoke(ctx context.Context, req *rpc.InvokeRequest
...
+		} else {
+			cred, err := k.newTokenCredential()
+			if err != nil {
+				return nil, err
+			}
+			t, err := cred.GetToken(ctx, policy.TokenRequestOptions{
+				Scopes: []string{endpoint + "/.default"},
+			})
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
thomas11: I read through the source and the only thing I could find that it contains is the tenant id. Which is fine IMO.

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+	} else {
+		logging.V(9).Infof("SP with client certificate credential is not enabled, skipping")
+	}
+
+	if authConf.clientSecret != "" {
+		logging.V(9).Infof("[auth] Using SP with client secret credential")
+		options := &azidentity.ClientSecretCredentialOptions{
+			AdditionallyAllowedTenants: authConf.auxTenants, // usually empty which is fine
```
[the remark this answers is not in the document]
thomas11: Good catch!

provider/pkg/provider/auth_azidentity.go
```diff
@@ -0,0 +1,245 @@
...
+// newSingleMethodAuthCredential creates an azcore.TokenCredential. Depending on the given authConfiguration, it is
+// backed by one of several authentication methods such as Service Principal, OIDC, Managed Identity, or Azure CLI.
+//
+// Note: this function's behavior is written to match the behavior of
+// "github.com/hashicorp/go-azure-helpers/authentication".Builder.Build() in some ways, to minimize changes in provider
+// behavior when upgrading authentication dependencies from go-azure-helpers to azidentity. Namely:
+//   - The order in which the the different authentication methods are attempted is the same.
+//   - When a method is configured but instantiating the credential fails, we return an error and do not fall through to
+//     the next method.
+//   - Auxiliary or additional tenants are supported for SP with client secret and CLI authentication, not for others.
```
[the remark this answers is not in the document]
thomas11: Initially, I wanted to mimic the behavior of the old implementation as closely as possible. However, adding further capabilities is backwards-compatible, so I added the aux tenants where possible.

provider/pkg/provider/auth_azidentity_test.go
thomas11: I can add a test that a configuration with both also yields an MSI token credential, but we can't check for the client id without reflection. go-azure-in-azure is an integration test that tests both system and user-assigned managed identity.

provider/pkg/provider/auth_azidentity.go
thomas11: I've added a comment to hopefully elucidate a bit.

provider/pkg/provider/auth_azidentity.go
thomas11: It's not supported anymore. I'm tracking updates to the registry docs separately, and have added a clarifying comment here.

provider/pkg/provider/auth_azidentity.go
thomas11: It's unused currently. Auth doesn't need it. However, when we fully replace the previous auth with the new one, the provider (provider.go) will need it. Currently, it reads it from the old auth config.

thomas11: On second thought, that's weird factoring. I'll change it so the provider reads it directly as it's the only user.

provider/pkg/provider/provider.go
thomas11: Added some. It's tricky to test the core logic because it gets actual Azure tokens and isn't abstracted behind an interface. We could do that but I'm punting for now. The smaller functions are hopefully easier to understand.

provider/pkg/provider/provider.go
```diff
@@ -237,12 +247,18 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 }
 
 func (k *azureNativeProvider) newAzureClient(armAuth autorest.Authorizer, tokenCred azcore.TokenCredential, userAgent string) (azure.AzureClient, error) {
-	if os.Getenv("PULUMI_USE_AUTOREST") == "false" {
-		logging.V(9).Infof("AzureClient: using azCore")
-		return azure.NewAzCoreClient(tokenCred, userAgent, k.getAzureCloud(), nil)
+	useAutorest := os.Getenv("PULUMI_USE_AUTOREST") != "false"
+
+	if !useAutorest && useLegacyAuth() {
+		return nil, errors.New("PULUMI_USE_LEGACY_AUTH=true requires PULUMI_USE_AUTOREST=true")
```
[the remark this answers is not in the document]
thomas11: > saying one true requires another true sounds wrong?

Why? It's only in one direction: useAutorest requires useLegacyAuth.

```
autorest | legacyAuth | valid?
-----------------------------
0        | 0          | yes
0        | 1          | yes
1        | 0          | NO
1        | 1          | yes
```

That being said, I think I'll switch to single feature flag. Will make everything much easier.

provider/pkg/azure/azure.go
```diff
@@ -61,3 +64,21 @@ func AzureError(err error) error {
...
+func GetCloudByName(cloudName string) azcloud.Configuration {
+	switch cloudName {
+	case "china":
+		return azcloud.AzureChina
+	case "usgov":
+		return azcloud.AzureGovernment
+	case "usgovernment":
+		return azcloud.AzureGovernment
+	}
+	return azcloud.AzurePublic
```
mikhailshilkov, line 83: Should it fail if we get some random "brazil" in here?

provider/pkg/provider/provider.go
```diff
@@ -216,15 +216,24 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 
 	userAgent := k.getUserAgent()
 
-	azCoreTokenCredential := azCoreTokenCredential{p: k}
+	var credential azcore.TokenCredential
+	if enableAzcoreBackend() {
+		credential, err = k.newTokenCredential()
```
mikhailshilkov, line 221: Add a similar log message "Using azidentity authentication"?

provider/pkg/azure/azure.go
mikhailshilkov: Ah, now I can see it's the same as it used to be, it's okay then

mikhailshilkov (approved):

PR #3651, opened 2024-10-17
Maintenance: bump some GH actions versions

thomas11:


.github/actions/install/action.yml
```diff
@@ -12,13 +12,13 @@ runs:
     - name: Install Go
       uses: actions/setup-go@v5
       with:
-        go-version: "1.22.x"
+        go-version: "1.23.x"
```
danielrbradley, line 15: We could get this value automatically from the go.mod ...

```suggestion
        go-version-file: "providergo.mod"
```

danielrbradley (approved):
LGMT

.github/actions/install/action.yml
thomas11: Oh nice, didn't know that!

PR #3653, opened 2024-10-18
Retry Conflicts for AnotherOperationInProgress

### LINKED ISSUE #903: Retry "AnotherOperationInProgress" error messages

danielrbradley:
The new azclient allows us to pass a callback function to decide if we want to retry a request. The default behaviour is to only retry a specific set of status codes. This extends that behaviour to also check Conflict (409) and retry if the "error.code" field is "AnotherOperationInProgress".

We're avoiding retrying all conflict status codes because some won't be fixed with a retry, so we're targeting the specific case instead.

Tested this both manually stepping through the code with a debugger and also created an integration test which previously failed but now passes when configured to use the new azclient instead of autorest client.

To access this improved behaviour right now, set the environment variable `PULUMI_ENABLE_AZCORE_BACKEND=true`. This will become the default in the future, but is currently in testing behind a feature flag.

This should address #903 which we will close once we've defaulted to using the new azcore implementation:
- https://github.com/pulumi/pulumi-azure-native/pull/3654

### DIFF
--- a/provider/pkg/azure/client_azcore.go
+++ b/provider/pkg/azure/client_azcore.go
@@ -61,6 +61,32 @@ func NewAzCoreClient(tokenCredential azcore.TokenCredential, userAgent string, a
 	opts.Retry.RetryDelay = 20 * time.Second
 	opts.Retry.MaxRetryDelay = 120 * time.Second
 
+	// These are the same as the default values in sdk/azcore/runtime/policy_retry.go setDefaults
+	retryableStatusCodes := []int{
+		http.StatusRequestTimeout,      // 408
+		http.StatusTooManyRequests,     // 429
+		http.StatusInternalServerError, // 500
+		http.StatusBadGateway,          // 502
+		http.StatusServiceUnavailable,  // 503
+		http.StatusGatewayTimeout,      // 504
+	}
+
+	opts.Retry.ShouldRetry = func(resp *http.Response, err error) bool {
+		if err != nil {
+			return true
+		}
+		// Replicate default retry behaviour first.
+		if runtime.HasStatusCode(resp, retryableStatusCodes...) {
+			return true
+		}
+
+		if shouldRetryConflict(resp) {
+			return true
+		}
+
+		return false
+	}
+
 	pipeline, err := armruntime.NewPipeline("pulumi-azure-native", version.Version, tokenCredential,
 		runtime.PipelineOptions{}, opts)
 	if err != nil {
@@ -74,6 +100,18 @@ func NewAzCoreClient(tokenCredential azcore.TokenCredential, userAgent string, a
 		deletePollingIntervalSeconds: 30, // same as autorest.DefaultPollingDelay
 		updatePollingIntervalSeconds: 10,
 	}, nil
+}
+
+func shouldRetryConflict(resp *http.Response) bool {
+	if runtime.HasStatusCode(resp, http.StatusConflict) {
+		err := runtime.NewResponseError(resp)
+		if responseErr, ok := err.(*azcore.ResponseError); ok {
+			if responseErr.ErrorCode == "AnotherOperationInProgress" {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 func (c *azCoreClient) setHeaders(req *policy.Request, contentTypeJson bool) {
--- a/provider/pkg/azure/client_azcore_test.go
+++ b/provider/pkg/azure/client_azcore_test.go
@@ -630,3 +630,50 @@ func TestPutPatchRequestBody(t *testing.T) {
 			qp, "")
 	})
 }
+
+func TestShouldRetryConflict(t *testing.T) {
+	shouldRetry := func(status int, body string) bool {
+		resp := http.Response{
+			StatusCode: status,
+			Body:       io.NopCloser(strings.NewReader(body)),
+		}
+		return shouldRetryConflict(&resp)
+	}
+	t.Run("Not a conflict", func(t *testing.T) {
+		assert.False(t, shouldRetry(200, ""))
+	})
+	t.Run("Blank body", func(t *testing.T) {
+		assert.False(t, shouldRetry(409, ""))
+	})
+	t.Run("Unreadable body", func(t *testing.T) {
+		resp := http.Response{
+			StatusCode: 409,
+			Body:       errorReadCloser{},
+		}
+		assert.False(t, shouldRetryConflict(&resp))
+	})
+	t.Run("empty json", func(t *testing.T) {
+		assert.False(t, shouldRetry(409, "{}"))
+	})
+	t.Run("wrong error type", func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


thomas11:
Note that the feature toggle variable has been renamed to `PULUMI_ENABLE_AZCORE_BACKEND`.

danielrbradley:
> Note that the feature toggle variable has been renamed to `PULUMI_ENABLE_AZCORE_BACKEND`.

This code path still seems to use the variable I'm referencing: https://github.com/pulumi/pulumi-azure-native/blob/6a41396ffd60f1d2647fe14e76dbb291baa18258/provider/pkg/provider/provider.go#L240

provider/pkg/azure/client_azcore.go
```diff
@@ -61,6 +62,41 @@ func NewAzCoreClient(tokenCredential azcore.TokenCredential, userAgent string, a
...
+			// Check body for {"error":{"code":"AnotherOperationInProgress"}}
+			body, err := runtime.Payload(resp)
+			if err != nil {
+				return false
+			}
+			var errorBody map[string]interface{}
+			if err := json.Unmarshal(body, &errorBody); err != nil {
+				return false
+			}
+			if error, ok := errorBody["error"].(map[string]interface{}); ok {
```
thomas11, line 90: I didn't know this was possible in Go. The type conversion doesn't panic even when there's no "error" key in the map?

provider/pkg/azure/client_azcore.go
```diff
@@ -61,6 +62,41 @@ func NewAzCoreClient(tokenCredential azcore.TokenCredential, userAgent string, a
...
+	// These are the same as the default values in sdk/azcore/runtime/policy_retry.go setDefaults
+	retryableStatusCodes := []int{
+		http.StatusRequestTimeout,      // 408
+		http.StatusTooManyRequests,     // 429
+		http.StatusInternalServerError, // 500
+		http.StatusBadGateway,          // 502
+		http.StatusServiceUnavailable,  // 503
+		http.StatusGatewayTimeout,      // 504
+	}
+	opts.Retry.ShouldRetry = func(resp *http.Response, err error) bool {
```
thomas11, line 74: A unit test for ShouldRetry would be good.

thomas11:
> > Note that the feature toggle variable has been renamed to `PULUMI_ENABLE_AZCORE_BACKEND`.
> 
> This code path still seems to use the variable I'm referencing:
> 
> https://github.com/pulumi/pulumi-azure-native/blob/6a41396ffd60f1d2647fe14e76dbb291baa18258/provider/pkg/provider/provider.go#L240

Yes, the rename is in a PR #3630 that will be merged shortly (I hope).

danielrbradley:
> Yes, the rename is in a PR #3630 that will be merged shortly (I hope).

Ok, happy to wait for that to go in, update this, then merge.

provider/pkg/azure/client_azcore.go
danielrbradley: Gone with extracting the complex bit and unit testing that.

provider/pkg/azure/client_azcore.go
danielrbradley: Reworked this to use structs which are easier to reason about.

provider/pkg/azure/client_azcore.go
```diff
@@ -61,6 +62,32 @@ func NewAzCoreClient(tokenCredential azcore.TokenCredential, userAgent string, a
...
+		http.StatusServiceUnavailable,  // 503
+		http.StatusGatewayTimeout,      // 504
+	}
+
+	opts.Retry.ShouldRetry = func(resp *http.Response, err error) bool {
+		if err != nil {
+			return true
+		}
+		// Replicate default retry behaviour first.
+		if runtime.HasStatusCode(resp, retryableStatusCodes...) {
```
mikhailshilkov, line 80: Is there no way to literally call the old retry function instead of duplicating the logic?

provider/pkg/azure/client_azcore.go
```diff
@@ -76,6 +103,24 @@ func NewAzCoreClient(tokenCredential azcore.TokenCredential, userAgent string, a
 	}, nil
 }
 
+func shouldRetryConflict(resp *http.Response) bool {
+	if runtime.HasStatusCode(resp, http.StatusConflict) {
+		// Check body for {"error":{"code":"AnotherOperationInProgress"}}
+		body, err := runtime.Payload(resp)
```
mikhailshilkov, line 109: Could we use https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/exported/response_error.go#L23 instead? It looks like it's doing a similar parsing?

provider/pkg/azure/client_azcore.go
danielrbradley: Nope, it's not actually a standalone function. It's just inline within the main retry implementation: https://github.com/Azure/azure-sdk-for-go/blob/068ef3d103a28dbd9657573091ccab7b5caca775/sdk/azcore/runtime/policy_retry.go#L179-L190

And the default status codes are defined here, but the setDefaults method isn't available either. The defaults are also only set on the internal copy of the struct so we can't just reference them from the opts object:
https://github.com/Azure/azure-sdk-for-go/blob/068ef3d103a28dbd9657573091ccab7b5caca775/sdk/azcore/runtime/policy_retry.go#L48-L58

provider/pkg/azure/client_azcore.go
danielrbradley: Guess we could, but the we get back an untyped error which we have to try-cast into a `ResponseError` to access the `ErrorCode` field. I think what's here is a little easier to follow and has less magic.

mikhailshilkov: What's magical about casting to `ResponseError`? I feel like deserializing into an arbitrary structure is more bespoke.

thomas11 (approved):
I like the ResponseError suggestion and would try to use it. But approving since the current version also LGTM.

provider/pkg/azure/client_azcore.go
danielrbradley: I guess it being called `NewResponseError` indicates it _really_ should be of type ResponseError .. and they shouldn't change the concrete type beneath us in the future. So guess it's probably fine. Also, yes looks like it handles the case of it being XML too.

danielrbradley:
Squashed, rebased and switch PULUMI_USE_AUTOREST to PULUMI_ENABLE_AZCORE_BACKEND in the test

PR #3663, opened 2024-10-23
Custom resource for Azure PIM role management policies

thomas11:
## About

This PR adds support for [Role Management Policies](https://learn.microsoft.com/en-us/rest/api/authorization/privileged-role-policy-rest-sample), part of [Privileged Identity Management (PIM)](https://learn.microsoft.com/en-us/entra/id-governance/privileged-identity-management/pim-configure) in the  Microsoft.Authentication namespace. It's one part of #2455.

Note that this is about the ARM part of PIM; there's also a Microsoft Graph API part which is not covered by this provider.

This resource wasn't automatically included because it supports only GET and PATCH. The policies are singletons that cannot be created or deleted, only modified via PATCH.

## Implementation

Role Management Policies essentially consist of a name which is actually a GUID, and a list of ~20 rules.

Using our existing singleton support `defaults.GetDefaultResourceState` was tricky because
1. there are many policies with many rules, for a total of ~300k lines of JSON for a subscription scope, and there are more scopes, plus
2. I believe the defaults can vary per scope and possibly also per customer.

So instead, I've implemented a custom resource that captures the original state of a policy when it's first "created", i.e., added to Pulumi state. When a rule or the whole policy is removed from Pulumi, we look up the original state and re-apply it.

## Testing

The e2e/integration test for this resource is special because using PIM requires a paid [Entra ID P2 license](https://www.microsoft.com/en-us/security/business/microsoft-entra-pricing). We have one that you can see [here](https://admin.microsoft.com/Adminportal/Home?referrer=entra#/licensedetailpage/84a661c4-e949-4bd2-a560-ed7766fcaf2b).

### DIFF
--- a/provider/pkg/resources/customresources/custom_blob_container_legal_hold.go
+++ b/provider/pkg/resources/customresources/custom_blob_container_legal_hold.go
@@ -224,7 +224,7 @@ func blobContainerLegalHold(azureClient azure.AzureClient) *CustomResource {
 
 			return input, nil
 		},
-		Delete: func(ctx context.Context, id string, inputs, _ resource.PropertyMap) error {
+		Delete: func(ctx context.Context, id string, inputs, state resource.PropertyMap) error {
 			path := strings.TrimSuffix(id, "/legalHold") + "/clearLegalHold"
 
 			tags, err := readTags(inputs.Mappable())
--- a/provider/pkg/resources/customresources/custom_keyvault.go
+++ b/provider/pkg/resources/customresources/custom_keyvault.go
@@ -18,7 +18,7 @@ import (
 func keyVaultSecret(keyVaultDNSSuffix string, tokenCred azcore.TokenCredential) *CustomResource {
 	return &CustomResource{
 		path: "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}/secrets/{secretName}",
-		Delete: func(ctx context.Context, id string, inputs, _ resource.PropertyMap) error {
+		Delete: func(ctx context.Context, id string, inputs, state resource.PropertyMap) error {
 			vaultName := inputs["vaultName"]
 			if !vaultName.HasValue() || !vaultName.IsString() {
 				return errors.New("vaultName not found in resource state")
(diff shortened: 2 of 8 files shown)


provider/pkg/convert/sdkInputsToRequestBody.go
```diff
@@ -41,16 +42,16 @@ func (k *SdkShapeConverter) SdkInputsToRequestBody(props map[string]resources.Az
...
-	var err error
+
 	if len(unusedValues) > 0 {
 		unusedKeys := make([]string, 0, len(unusedValues))
 		for k := range unusedValues {
 			unusedKeys = append(unusedKeys, k)
 		}
 		sort.Strings(unusedKeys)
-		err = fmt.Errorf("unrecognized properties: %v", unusedKeys)
+		logging.V(9).Infof("Unrecognized properties in SdkInputsToRequestBody: %v", unusedKeys)
```
danielrbradley, line 52: This was downgraded to a warning in [this call site](https://github.com/pulumi/pulumi-azure-native/blob/f1cae56b80520aa3a905dd4736764db1fe35e162/provider/pkg/provider/provider.go#L1522-L1526) too. Do we just drop it? Does this actually catch any real issues?

provider/pkg/resources/customresources/custom_pim.go
```diff
@@ -0,0 +1,214 @@
...
+	"slices"
+
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider/crud"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/logging"
+)
+
+// OriginalStateKey is an internal key in the state of a PIM Role Management Policy that contains the original state of the policy.
+const OriginalStateKey = "__orig_state"
```
danielrbradley, line 16: This is very interesting living in here .. perhaps this should at least be in `customresources.go`?

Is there any way we can avoid this custom resource manually writing this property into the outputs and instead make that happen in the `provider.go` .. at which point this could just be a const within the provider module?

Nit: I wouldn't personally abbreviate "original" to "orig" for something users might be looking at.

danielrbradley (commented):
Initial comments inline. The overall approach looks good. Perhaps we can untangle the magic "original state" field from the individual custom resource implementations?

provider/pkg/provider/provider.go
```diff
@@ -991,6 +982,37 @@ func (k *azureNativeProvider) Create(ctx context.Context, req *rpc.CreateRequest
...
+
+	if customRes.CanCreate != nil {
+		err = customRes.CanCreate(ctx, id)
+		exists = err != nil
+	} else if customRes.Read != nil {
+		_, exists, err = customRes.Read(ctx, id, inputs)
+	} else {
+		err = crudClient.CanCreate(ctx, id)
+		exists = err != nil
+	}
```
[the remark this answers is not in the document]
thomas11: Regular resources don't use Read (directly) in this path, they use `crudClient.CanCreate`, which I replicated here.

danielrbradley (approved):
Let's open a follow up issue to expand this to capture initial states for all resources to be able to remove the need to manually curate a list of default resource states where there's no delete operation in the spec.

PR #3664, opened 2024-10-23
Migrate custom Key Vault resource to azidentity

### LINKED ISSUE #2432: Invalid audience for Key Vault secret operations when using MSI

thomas11:
This PR ended up having three distinct but related parts to it.

1. Migrate the custom KV resource to the new `azidentity` backend. We need to preserve the previous one because it uses a special Autorest authorizer for Key Vault.
2. Use a Key Vault secret in the azure-in-azure integration test. KV secrets need a different authentication audience/scope in the access token and we want to cover this case.
3. A fix that affects master as well: the azure-in-azure test didn't use the correct environment variable for the new backend toggle.

[Green run of the azcore workflow using the new backend](https://github.com/pulumi/pulumi-azure-native/actions/runs/11670925535)

Fixes #2432

### DIFF
--- a/examples/examples_go_sdk_test.go
+++ b/examples/examples_go_sdk_test.go
@@ -92,6 +92,10 @@ func TestServicebusRecreateSdk(t *testing.T) {
 }
 
 func TestAzureInAzureWithSystemManagedIdentity(t *testing.T) {
+	// This test fails on the autorest backend, see #2432
+	if os.Getenv("PULUMI_ENABLE_AZCORE_BACKEND") != "true" {
+		t.Skip("Skipping test because azcore backend is not enabled")
+	}
 	test := getGoBaseOptionsSdk(t, testDir(t, "go-azure-in-azure"))
 	integration.ProgramTest(t, &test)
 }


provider/pkg/resources/customresources/custom_keyvault.go
```diff
@@ -7,14 +7,15 @@ import (
...
 	"github.com/pkg/errors"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
 
 // keyVaultSecret creates a custom resource for Azure KeyVault Secret.
-func keyVaultSecret(keyVaultDNSSuffix string, kvClient *keyvault.BaseClient) *CustomResource {
+func keyVaultSecret(keyVaultDNSSuffix string, tokenCred azcore.TokenCredential) *CustomResource {
 	return &CustomResource{
 		path: "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}/secrets/{secretName}",
 		Delete: func(ctx context.Context, id string, properties resource.PropertyMap) error {
```
danielrbradley, line 21: Aside: we should probably explain why this custom resource only has a delete method. 

Assuming it's because the default implementation works fine for the other methods, but we want to customise the delete behaviour?

provider/pkg/resources/customresources/customresources.go
```diff
@@ -198,6 +192,21 @@ func BuildCustomResources(env *azureEnv.Environment,
 		customWebAppSlot,
 	}
 
+	// For Key Vault, we need to use separate token sources for azidentity and for the legacy auth. The
+	// `azCoreTokenCredential` adapter that we use elsewhere to translate legacy token sources to azidentity doesn't
+	// work here because KV needs a different token source for the KV endpoint.
+	useLegacyAuth := os.Getenv("PULUMI_USE_LEGACY_AUTH") != "false"
```
danielrbradley: Do we have a helper method for this which we can check instead of going directly to the OS env?

danielrbradley (approved):
Core implementation looks solid 👍

PR #3698, opened 2024-11-11
Port custom blob and static website resources to current SDK, azcore, and azidentity

### LINKED ISSUE #3685: Convert the custom storage resource to the new azure-sdk-for-go package and azidentity auth
Currently, custom_storage.go uses `github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-04-01/storage` which requires an Autorest `Authorizer` for authentication. As such, it blocks removing the legacy Autorest backend, and doesn't profit from improvements made on the new backend.

thomas11:
Resolves #3685.

The custom resources **StorageAccountStaticWebsite** and **Blob** are implemented based on the deprecated SDK `github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-04-01/storage`. This SDK authenticates via Autorest authorizers, making it doubly deprecated.

This PR introduces a new, parallel implementation of these resources, based on the current Azure SDK and the new azidentity-based authentication. Depending on how the `PULUMI_ENABLE_AZCORE_BACKEND` toggle is set, one of the two implementations will be used.

### DIFF
--- a/provider/pkg/resources/customresources/custom_storage_azidentity.go
+++ b/provider/pkg/resources/customresources/custom_storage_azidentity.go
@@ -25,46 +25,14 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-// newStorageAccountStaticWebsite creates a custom resource to mark Azure Storage Account as a static website.
-func storageAccountStaticWebsite_azidentity(env cloud.Configuration, creds azcore.TokenCredential) *CustomResource {
-	r := staticWebsite_azidentity{
-		env:   env,
-		creds: creds,
-	}
-	return &CustomResource{
-		path:   staticWebsitePath,
-		tok:    "azure-native:storage:StorageAccountStaticWebsite",
-		Create: r.createOrUpdate,
-		Update: r.update,
-		Read:   r.read,
-		Delete: r.delete,
-		LegacySchema: &schema.ResourceSpec{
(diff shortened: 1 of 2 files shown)


provider/pkg/resources/customresources/custom_storage_azidentity.go
```diff
@@ -0,0 +1,700 @@
...
+		env:   env,
+	}
+	return &CustomResource{
+		path:   blobPath,
+		tok:    "azure-native:storage:Blob",
+		Create: r.create,
+		Update: r.update,
+		Delete: r.delete,
+		Read:   r.read,
+		Types: map[string]schema.ComplexTypeSpec{
```
danielrbradley, line 257: Given the types and schema need to be identical, is it worth pulling these out in the original implementation to a function that can be called from here too?

Also, should we migrate these over from `SchemaLegacy`+`Types` to the new `Schema` function too? I think this is a very easy-to-verify refactor as it's easy to see in the SDK changes.

danielrbradley (approved):
Core implementation looks good.

How long do we think the dual implementations will stick around? Might be worth de-deuplicating the schemas to avoid one accidentally diverging.

provider/pkg/resources/customresources/custom_storage_azidentity.go
thomas11: I'll look into pulling out schema and types.

If we migrate to the new `Schema`, I think it should be in a different PR.

Issue #3687, opened 2024-11-06
Unable to create azure-native.machinelearningservices.ServerlessEndpoint resource


patrickstuddard:
### What happened?

I was trying to create a ServerlessEndpoint with Python code like this:

```
serverless_endpoint=azure.machinelearningservices.ServerlessEndpoint(
    "serverlessEndpoint",
    resource_group_name=resource_group.name,
    workspace_name=ml_project.name,
    serverless_endpoint_properties=azure.machinelearningservices.ServerlessEndpointArgs(
        offer=azure.machinelearningservices.ServerlessOfferArgs(
            offer_name="mistral-ai-large-2407-offer",
            publisher="000-000"
        ),
        auth_mode="AAD"
    ),
    sku = {
        "name": "Consumption"
    }
)
```

And I got this error back:

`error: Code="UserError" Message="Serverless offer 'mistral-ai-large-2407-offer' not found in service namespace '000-000'." Details=[] AdditionalInfo=[{"info":{"value":"managementfrontend"},"type":"ComponentName"},{"info":{"value":{"operation":"c2d892c742bbb4d3af4b4b803733023f","request":"901eee4ec8280b5c"}},"type":"Correlation"},{"info":{"value":"westus"},"type":"Environment"},{"info":{"value":"westus"},"type":"Location"},{"info":{"value":"2024-10-29T19:53:35.2992082+00:00"},"type":"Time"},{"info":{"value":{"code":"NotFound","innerError":{"code":"ServerlessOfferNotFound","innerError":null}}},"type":"InnerError"}]`

The valid values for `offer_name` and `publisher` don’t seem to be documented anywhere, but I tried all sorts of variations, and also with different models, and always got this error.

When I create the endpoint manually through the Azure portal and then import it into Pulumi, it gives me this warning:

`warning: One or more imported inputs failed to validate. This is almost certainly a bug in the `azure-native` provider. The import will still proceed, but you will need to edit the generated code after copying it into your program.
warning: azure-native:machinelearningservices:ServerlessEndpoint resource 'string' has a problem: missing required property 'serverlessEndpointProperties'`

And `offer_name` and `publisher` show up as empty strings.

[...]

PR #3704, opened 2024-11-13
Add docs about MachineLearningServices ServerlessEndpoint requiring a MarketplaceSubscription

thomas11:
Follow-up to #3687. While the user eventually figured out the correct way to provision the resource, it's obscure and undocumented, so add a note to the docs.


thomas11:
Markdown code blocks seem to be stripped in SDK docs, not great.

danielrbradley (approved):

Issue #2661, opened 2023-08-18
Use the Azure RP API when determining default API versions


thomas11:
Repair item for pulumi/home#3020. During the v2 release, in two sets of resources, DNS and IoT Hubs, new default Azure API versions were wrongly chosen, causing all Pulumi operations against these resources to fail.

For both RPs that we needed to fix, the RP API returns the correct API version we ended up going with:
```
$ az provider show --namespace Microsoft.Network | jq '.resourceTypes[] | select(.resourceType == "dnszones/recordsets") | .apiVersions[0]'
"2018-05-01"

$ az provider show --namespace Microsoft.Devices | jq '.resourceTypes[] | select(.resourceType == "IotHubs") | .apiVersions[0]'
"2022-04-30-preview"
```

The only part we have not automated yet is mapping our resources to the API’s resourceType: RecordSet = “dnszones/recordsets”, IotHubResource = “IotHubs”.

Once we have this mapping, we can implement an automated process that retrieves the latest supported version for each resource.

It turns out we already collect this data and store it in `azure-provider-versions/provider_list.json` but we don't use it currently.

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3552

danielrbradley:
After some experimenting trying to start this implementation, here's an initial plan to address some of the existing complexity to make this process easier to reason about:

1. Move squeeze into az repo
2. Output removed-invokes from squeeze
   1. Consider moving deletion exclusions into the squeeze too
3. Remove openapi.AzureProviders from calculateVersionMetadata
4. Unit test calculateVersionMetadata
5. Integrate az-provider-list into version selection so we only use active versions

PR #3707, opened 2024-11-15
Leverage az provider list

danielrbradley:
Fixes https://github.com/pulumi/pulumi-azure-native/issues/2661

1. Add inactive default versions report - which versions referenced from our default version are not currently marked as "live" by `az provider list`?
2. When picking a new tracking version, select from the live versions only, if any are live. Fall back to all versions if `az provider list` doesn't include any.
3. When checking for additional resources not in the default version, don't add new non-live versions.
4. Move most unit testing into gen_calculateVersionMetadata_test from defaultVersion_test as it's as close to e2e as possible and requires less mocking.

### DIFF
--- a/provider/pkg/versioning/defaultVersion_test.go
+++ b/provider/pkg/versioning/defaultVersion_test.go
@@ -5,292 +5,9 @@ import (
 	"time"
 
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/openapi"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/providerlist"
 	"github.com/stretchr/testify/assert"
 )
-
-func TestDefaultVersion(t *testing.T) {
-	t.Run("simple latest additive", func(t *testing.T) {
-		actual := BuildSpec(map[openapi.ProviderName]VersionResources{
-			"Provider": {
-				"2020-01-01": []openapi.ResourceName{
-					"Resource A",
-				},
-				"2021-02-02": []openapi.ResourceName{
-					"Resource A",
-					"Resource B",
-				},
-			},
-		}, nil, Spec{})
-		v2021 := "2021-02-02"
-		expected := Spec{
-			"Provider": {
-				Tracking: &v2021,
-			},
-		}
-		assert.Equal(t, expected, actual)
-	})
-	t.Run("two-previews", func(t *testing.T) {
-		actual := BuildSpec(map[openapi.ProviderName]VersionResources{
-			"Provider": {
-				"2020-01-01-preview": []openapi.ResourceName{
-					"Resource A",
-				},
-				"2021-02-02-preview": []openapi.ResourceName{
-					"Resource A",
-					"Resource B",
-				},
-			},
-		}, nil, Spec{})
-		v2021 := "2021-02-02-preview"
-		expected := Spec{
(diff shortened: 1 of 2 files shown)


thomas11 (commented):
Pretty tricky to review but overall LGTM. Thanks for the nice test coverage!

> Fall back to all versions if as provider list doesn't include any.

Is this in case `az provider list` is wrong? If it was 100% correct, we could disregard that resource altogether, no?

My main concern is that I don't think we should merge the schema and SDKs with these changes in v2. Even if the Azure API versions are not actually live, it's still a breaking code change. And there's a small chance `provider list` is wrong.

danielrbradley:
> My main concern is that I don't think we should merge the schema and SDKs with these changes in v2. Even if the Azure API versions are not actually live, it's still a breaking code change. And there's a small chance `provider list` is wrong.

Agreed. I was also concerned about removing all these. Perhaps we should feature switch the additions removals off until V3.

danielrbradley:
@thomas11 Ok, latest set of changes from the original design are up.

Instead of feature switching the removal of the non-live additions, I've just changed the logic slightly to not remove additions which are already part of the spec - even if they're not live. This now results in zero schema changes.

This maintains the current set of default resources but will naturally kick in for v3 when we generate a new default version from scratch.

Makefile
```diff
@@ -112,8 +112,8 @@ update_submodules:
 		echo "Updating submodule $$submodule" ; \
 		(cd $$submodule && git checkout main && git pull origin main); \
 	done
-	rm ./azure-provider-versions/provider_list.json
-	az provider list | jq 'map({ namespace: .namespace, resourceTypes: .resourceTypes | map({ resourceType: .resourceType, apiVersions: .apiVersions }) | sort_by(.resourceType) }) | sort_by(.namespace)' > ./azure-provider-versions/provider_list.json
+	rm versions/az-provider-list.json
+	az provider list --query 'sort_by([*].{ namespace: namespace, resourceTypes: sort_by(resourceTypes[*].{ resourceType: resourceType, defaultApiVersion: defaultApiVersion, apiVersions: apiVersions, locations: locations }, &resourceType) }, &namespace)' > versions/az-provider-list.json
```
thomas11, line 116: This line could use a comment

provider/pkg/providerlist/providerlist.go
```diff
@@ -24,8 +24,24 @@ type ApiVersion = string
 // ProviderPathVersions is a map of lowered provider names to Api Versions e.g. `analysisservices -> "locations/checknameavailability" -> [2019-03-01-preview, 2018-05-05]`
 type ProviderPathVersions = map[LoweredProviderName]map[ResourcePath]ApiVersions
 
+type ResourceType struct {
```
thomas11, line 27: Short doc comments for public types? I'm not sure what a "ResourceType" is without reading more code.

thomas11 (approved):
LGTM with some docs wishlist items

provider/pkg/providerlist/providerlist.go
danielrbradley: Good thoughts. I've added via a rebase 👍

Makefile
danielrbradley: Done 👍

PR #3714, opened 2024-11-18
Refactor versioning process

danielrbradley:
1. Copy squeeze code from schema-tools.
2. Combine `gen raw-schema` + `schema-tools squeeze` into `gen squeeze`.
3. Output removed invokes from squeeze & remove from `BuildSchema`.
4. Tidy and simplify `BuildSchema` internal methods.
5. Add unit tests as documentation/spec.
6. Fix bug to get tests to pass.

Precursor to implementing https://github.com/pulumi/pulumi-azure-native/issues/2661

### DIFF
--- a/provider/pkg/versioning/gen_calculateVersionMetadata_test.go
+++ b/provider/pkg/versioning/gen_calculateVersionMetadata_test.go
@@ -1,0 +1,175 @@
+package versioning
+
+import (
+	"testing"
+
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/openapi"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCalculateVersionMetadata(t *testing.T) {
+	t.Run("empty", func(t *testing.T) {
+		sources := VersionSources{}
+
+		result, err := calculateVersionMetadata(sources)
+
+		assert.NoError(t, err)
+		assert.NotNil(t, result)
+	})
+
+	t.Run("New module tracks latest version", func(t *testing.T) {
+		sources := VersionSources{
+			AllResourcesByVersion: map[string]map[string][]string{
+				"Module": {
+					"2019-01-01": {"Resource1", "Resource2"},
+					"2020-01-01": {"Resource1", "Resource2"},
+				},
+			},
+		}
+
+		result, err := calculateVersionMetadata(sources)
+
+		assert.NoError(t, err)
+		assert.Equal(t, Spec(map[openapi.ProviderName]ProviderSpec{
+			"Module": {
+				Tracking: strptr("2020-01-01"),
+			},
+		}), result.Spec)
+	})
+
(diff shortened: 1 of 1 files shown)


provider/pkg/versioning/defaultVersion.go
```diff
@@ -402,15 +403,19 @@ func timeBetweenVersions(from, to openapi.ApiVersion) (diff time.Duration, err e
 	return toTime.Sub(fromTime), err
 }
 
-func maxKey[K comparable, V any](m map[K]V) *K {
+func maxKey[K cmp.Ordered, V any](m map[K]V) *K {
 	keys := keys(m)
 	if len(keys) == 0 {
 		return nil
 	}
+	sort.SliceStable(keys, func(i, j int) bool {
```
thomas11, line 411: [nit] could use simpler [slices.Sort](https://pkg.go.dev/slices#Sort) without comparison func?

provider/pkg/versioning/gen_calculateVersionMetadata_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCalculateVersionMetadata(t *testing.T) {
+	t.Run("empty", func(t *testing.T) {
+		sources := VersionSources{}
+
+		result, err := calculateVersionMetadata(sources)
+
+		assert.NoError(t, err)
```
thomas11, line 16: [nit] all `assert.NoError` should probably be `require.NoError` since the following assertions don't make sense when there's an error.

thomas11 (approved):
LGTM with nits. Great start to a proper test suite! The readability of nested subtests is a matter of taste, I guess.

provider/pkg/versioning/defaultVersion.go
danielrbradley: Good shout

provider/pkg/versioning/gen_calculateVersionMetadata_test.go
danielrbradley: This function actually still often returns a partial result on error to let us output some version files even if something is invalid to help with debugging. So it's also kinda useful to see the result here too as it might give a hint as to what's happenned in the transformation.

danielrbradley:
> The readability of nested subtests is a matter of taste, I guess

Would you think a single level of nesting with longer test names is better instead?

provider/pkg/versioning/defaultVersion.go
danielrbradley: I always miss this one because I go to the sort package and look at the functions there 🙃

danielrbradley: Oh, one better - we can use `slices.Max(...)`

PR #3718, opened 2024-11-20
Correctly handle blob MD5 sums in custom blob resource, need to be base64-encoded

thomas11:
This was initially missed because the previously used Azure SDK in custom_storage.go, and the newer one used in custom_storage_azidentity.go, behave differently. The older SDK leaves the user-provided MD5 sum alone, so the user needs to provide it base64-encoded, as the integration test [examples/static-website/index.ts](https://github.com/pulumi/pulumi-azure-native/blob/master/examples/static-website/index.ts#L45) does. The newer SDK, however, expects the MD5 sum as a raw byte array and [does the base64-decoding internally](https://github.com/Azure/azure-sdk-for-go/blob/15c65ced02ca0b4875e8463146d7bd4a68a0fbb1/sdk/storage/azblob/internal/generated/zz_blockblob_client.go#L881). Providing an already encoded sum leads to double-encoding and an invalid sum being sent to Azure.

### DIFF
--- a/sdk/dotnet/Storage/Blob.cs
+++ b/sdk/dotnet/Storage/Blob.cs
@@ -127,7 +127,7 @@ namespace Pulumi.AzureNative.Storage
         public Input<string> ContainerName { get; set; } = null!;
 
         /// <summary>
-        /// The MD5 sum of the blob contents. Cannot be defined if blob type is Append.
+        /// The MD5 sum of the blob contents, base64-encoded. Cannot be defined if blob type is Append.
         /// </summary>
         [Input("contentMd5")]
         public Input<string>? ContentMd5 { get; set; }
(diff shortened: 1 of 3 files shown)


thomas11:
> Change looks good. Would be good to have an integration test for the blobs running ahead of merge at some point.
> 
> Options I can see here:
> 
> 1. Assuming you've run the test locally, that should suffice for now.
> 2. We could also do a manual CI run for the branch to include all tests.
> 3. We could also unmark one of the integration tests with `skipIfShort` so it's always run.
> 4. Or we should introduce a merge queue so it only merges after the full integration test suite has run.

I ran the full test suite via `gh workflow run acceptance-test.yml --ref tkappler/blob-md5`, [result here](https://github.com/pulumi/pulumi-azure-native/actions/runs/11931454936).

PR #3725, opened 2024-11-21
Don't upload codecov data from autorest and release jobs

mikhailshilkov:
We currently have a problem in historic codecov data where coverage jumps up and down between master commits. Examples
- https://app.codecov.io/gh/pulumi/pulumi-azure-native/commit/0e7c8eefcfa8efc2927d5a8acc147925bd72fb02
- https://app.codecov.io/gh/pulumi/pulumi-azure-native/commit/93fac42eb6f44197a625edce2aae360088ffd0f0

It turns out that the larger coverage number comes from two or three CI jobs (master + scheduled autorest + release) and lower number is just from master. Since we can't guarantee that we will run a release and an autorest nightly on every commit, this PR disables coverage data on those jobs.

We lose a little bit of coverage on that autorest legacy code, but I think stability is more important. Also, a PR vs. master coverage will be directly comparable.

### DIFF
--- a/.github/workflows/autorest-scheduled.yml
+++ b/.github/workflows/autorest-scheduled.yml
@@ -22,3 +22,4 @@ jobs:
       short_test: false
       retention_days: 7
       use_azcore: false
+      upload_codecov: false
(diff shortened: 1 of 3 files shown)


danielrbradley (approved):
Makes sense, thanks 👍

PR #3748, opened 2024-12-04
Preserve previous default versions in v3

### LINKED ISSUE #3424: When moving to a new major version, preserve all default API versions of the previous one
When Azure API version 2022-01-01 is the default for a resource in v2, and in v3 the new default is 2024-01-01, then the previous default 2022-01-01 should still be available in v3 as an explicit import. That enables to upgrade without changes, other than import paths.

We had already planned that for v2 but missed.

danielrbradley:
Fixes #3424

Before considering the resources marked for removal, if the resource version was in the previous default version, ensure we include it.

Only fix this for v3 onwards so we don't introduce schema changes now.

Bonus:
- Fix handling of "-beta" versions when converting back from SDK version to API version.

Stacked on top of #3753

### DIFF
--- a/provider/pkg/versioning/gen.go
+++ b/provider/pkg/versioning/gen.go
@@ -41,22 +41,12 @@ func (v VersionMetadata) ShouldInclude(provider string, version *openapi.ApiVers
 		return true
 	}
 	// Keep any resources in the default version lock
-	if resources, ok := v.Lock[provider]; ok {
-		if defaultResourceVersion, ok := resources[typeName]; ok {
-			if defaultResourceVersion == *version {
-				return true
-			}
-		}
+	if v.Lock.IsAtVersion(provider, typeName, *version) {
+		return true
 	}
 	// Keep any resources in the previous version lock for easier migration
-	if v.MajorVersion >= 3 {
-		if resources, ok := v.PreviousLock[provider]; ok {
-			if previousDefaultResourceVersion, ok := resources[typeName]; ok {
-				if previousDefaultResourceVersion == *version {
-					return true
-				}
-			}
-		}
+	if v.MajorVersion >= 3 && v.PreviousLock.IsAtVersion(provider, typeName, *version) {
+		return true
 	}
 	// Exclude versions from removed versions
 	if versions, ok := v.RemovedVersions[provider]; ok {
--- a/provider/pkg/openapi/discover.go
+++ b/provider/pkg/openapi/discover.go
@@ -170,7 +170,18 @@ func NewVersionResources() VersionResources {
 type ProviderVersionList = map[ProviderName][]ApiVersion
 
 // DefaultVersionLock is an amalgamation of multiple API versions
-type DefaultVersionLock = map[ProviderName]map[DefinitionName]ApiVersion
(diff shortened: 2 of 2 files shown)


provider/pkg/gen/schema.go
```diff
@@ -775,8 +775,16 @@ func (g *packageGenerator) genResourceVariant(apiSpec *openapi.ResourceSpec, res
 	path := resource.PathItem
 
 	resourceTok := fmt.Sprintf(`%s:%s:%s`, g.pkg.Name, module, resource.typeName)
-	if !g.versioning.ShouldInclude(g.provider, g.apiVersion, resource.typeName, resourceTok) {
-		return nil
+	// Never skip resources in the default version
+	if g.apiVersion != "" {
+		// The "apiVersion" is actually the SDK version, not the API version.
```
thomas11, line 780: This seems pretty fragile, relying on each caller to pass the correct kind of API version (with no type support). Can we make `ShouldInclude` check and convert internally?

thomas11 (commented):
I wouldn't expect SDK changes in this PR. I guess the change affects the v1-v2 comparison, but I don't think we should change v2 retroactively.

provider/pkg/gen/schema.go
danielrbradley: This would be a bit annoying as the conversion can error, so we'd need to add the error return to ShouldInclude too.

We could make this type-safe by switching our type aliases to proper type definitions, but that would require a large-scale refactor of all usages. The minimal change I think we should make here is to fix the naming - change `g.apiVersion` to `g.sdkVersion` so it's more obvious what's happening.

provider/pkg/versioning/gen.go
```diff
@@ -42,6 +42,14 @@ func (v VersionMetadata) ShouldInclude(provider string, version string, typeName
...
 		}
 	}
+	// Keep any resources in the previous version lock for easier migration
+	if resources, ok := v.PreviousLock[provider]; ok {
+		if previousDefaultResourceVersion, ok := resources[typeName]; ok {
+			if previousDefaultResourceVersion == version {
+				return true
+			}
+		}
+	}
```
danielrbradley, line 52: So, it turns out the check above is incorrect too - because it's converting the version from an API version to an SDK version, but the lock file uses API versions 😫

provider/pkg/versioning/gen.go
```diff
@@ -42,6 +42,16 @@ func (v VersionMetadata) ShouldInclude(provider string, version openapi.ApiVersi
 			}
 		}
 	}
+	// Keep any resources in the previous version lock for easier migration
+	if v.MajorVersion >= 3 {
+		if resources, ok := v.PreviousLock[provider]; ok {
```
thomas11, line 47: Nit: given `DefaultVersionLock` is already a type, we could define `Has(provider, version)` on it to encapsulate a bit more and clean up client code like this one.

danielrbradley: Ok, think we can do that by converting the type alias to a definition 👍

danielrbradley: Done 👍

PR #3749, opened 2024-12-04
Remove aliases for lower-cased types introduced for v2

### LINKED ISSUE #2411: Remove aliases for lower-cased types introduced for v2
In #2366, we introduced a number of type aliases to allow users to upgrade certain resources to v2.0 without replacement. We can remove these aliases when we ship v3.0.

EronWright:
## Description

Removes the generation of aliases for the v1 lower-cased types, when the major version is 3 or up.

The generated code is unchanged for version 2, which is why this PR doesn't contain a bunch of generated code. It is expected the generated code will change when the `PROVIDER_VERSION` (i.e. the `majorVersion` within the generator) reaches 3.x.

## Testing
A new unit test was added: gen_aliases_test.go

Closes #2411

### DIFF
--- a/provider/pkg/gen/__snapshots__/gen_aliases_test_v2.snap
+++ b/provider/pkg/gen/__snapshots__/gen_aliases_test_v2.snap
@@ -1,0 +1,1615 @@
+
+[TestAliasesGen/v2 - 1]
+{
+ "config": {
+  "variables": {
+   "auxiliaryTenantIds": {
+    "description": "Any additional Tenant IDs which should be used for authentication.",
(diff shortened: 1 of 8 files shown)


danielrbradley (approved):
This all looks good 👍 Appreciate the testing too .. better late than never 😅 

Guessing the plan post-launch will be to delete all this code (including the tests?)? Let's file an issue if so to track this clean-up linking here so we know which code should be removed.

EronWright:
https://github.com/pulumi/pulumi-azure-native/issues/3754

PR #3753, opened 2024-12-05
Strongly type ApiVersion and SdkVersion

danielrbradley:
- Rename `packageGenerator.apiVersion` to `packageGenerator.sdkVersion`.
- Rename `packageGenerator.allApiVersions` to `packageGenerator.allSdkVersions`.
- Add packageGenerator.apiVersion as pointer to an ApiVersion, if we can convert it.
- Add `collections.OrderableSet` as replacement for `codegen.StringSet` which allows for non-string types.
- Document where a string can be either API or SDK versions and is therefore remaining a string for now.
- Fix type mismatch in `ShouldInclude` with the RemovedVersions file which included removed versions in the description.


provider/pkg/versioning/gen.go
```diff
@@ -33,19 +33,19 @@ type VersionMetadata struct {
...
+			if defaultResourceVersion == version {
 				return true
 			}
 		}
 	}
 	// Exclude versions from removed versions
 	if versions, ok := v.RemovedVersions[provider]; ok {
 		for _, removedVersion := range versions {
-			if openapi.ApiToSdkVersion(removedVersion) == version {
+			if removedVersion == version {
```
danielrbradley, line 48: This was previously a bug ... we were passing in an API version, but the [removedVersions file](https://github.com/pulumi/pulumi-azure-native/blob/master/versions/v2-removed.json) already contains API versions, so when we converted it to an SDK version it would never match.

This is the cause of the schema descriptions changes removing some available versions because they are not actually available. 

The actual versions are removed by the `RemoveDeprecations` in `deprecations.go` which did not convert to an SDK version first and so functions correctly.

provider/pkg/versioning/findNewerVersions.go
```diff
@@ -1,32 +1,30 @@
...
-		sort.Strings(newerVersions)
-		olderProviderVersions[providerName] = newerVersions
+		olderProviderVersions[providerName] = newerVersions.SortedValues()
 	}
 	return olderProviderVersions
 }
 
 func findMaxDefaultVersion(resourceVersions map[openapi.DefinitionName]openapi.ApiVersion) openapi.ApiVersion {
-	minVersion := ""
+	// TODO: Consider using a pointer instead of empty string for when there is no version
```
thomas11, line 26: TODOs without issue or owner never get fixed

thomas11 (commented):
Great readability improvement! One comment about potentially passing the wrong kind of version needs to be resolved, the others are "should have".

provider/pkg/gen/schema.go
```diff
@@ -285,19 +286,25 @@ func PulumiSchema(rootDir string, providerMap openapi.AzureProviders, versioning
...
 		for version := range versionMap {
 			versions = append(versions, version)
 		}
-		sort.Strings(versions)
+		slices.Sort(versions)
 
-		for _, version := range versions {
+		for _, sdkVersion := range versions {
+			// Attempt to convert back to an API version for use elsewhere
+			var apiVersion *openapi.ApiVersion
```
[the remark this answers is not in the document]
danielrbradley: Good question. This should only be nil where we're in the default version .. so I'll make this stricter to fail if there an error for the non-empty sdkVersion.

provider/pkg/versioning/findNewerVersions.go
danielrbradley: I'll remove the TODO part. It's not really an issue as the end result is functionally correct, but it's useful to have a comment explaining why we're using empty string as it's a little confusing when reading - because empty string isn't a valid version.

provider/pkg/gen/schema.go
```diff
@@ -1050,8 +1058,8 @@ func (g *packageGenerator) genFunctions(typeName, path string, specParams []spec
 	}
 
 	// Generate the function to get this resource.
-	functionTok := g.generateTok(typeName, g.apiVersion)
-	if !g.shouldInclude(typeName, functionTok, g.apiVersion) {
+	functionTok := g.generateTok(typeName, g.sdkVersion)
+	if g.apiVersion != nil && !g.shouldInclude(typeName, functionTok, *g.apiVersion) {
```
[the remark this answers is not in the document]
danielrbradley: Done 👍

thomas11 (approved):

Issue #3421, opened 2024-07-08
Add upgrade tests to test suite


thomas11:
This provider currently does not have any upgrade tests (using providertest's [VerifyUpgrade](https://github.com/pulumi/providertest#verifyupgrade)). To release v3 with confidence, we should add them. TBD: which ones exactly.

Coverage to complete:
- [x] `resources.ResourceGroup`
- [x] `keyvault.Vault`
- [x] `keyvault.Secret`
- [x] `keyvault.Key`
- [x] `network.VirtualNetwork`
- [x] `network.Subnet`
- [x] `network.NetworkInterface`
- [x] `network.PublicIPAddress`
- [ ] `network.PrivateEndpoint`
- [ ] `network.PrivateRecordSet`
- [ ] `network.RecordSet`
- [ ] `network.NetworkSecurityGroup`
- [x] `compute.VirtualMachine`
- [x] `storage.BlobContainer`
- [x] `storage.StorageAccount`
- [x] `storage.Blob`
- [ ] `authorization.RoleAssignment`
- [ ] `authorization.PolicyDefinitionAtManagementGroup`
- [x] `sql.FirewallRule`
- [ ] `appconfiguration.KeyValue`
- [x] `sql.Database`
- [x] `sql.VirtualNetworkRule`
- [x] `sql.Server`
- [x] `web.WebApp`
- [x] `web.AppServicePlan`
- [x] `web.WebAppSlot`
- [x] `web.WebAppSlotConfigurationNames`
- [ ] `insights.DiagnosticSetting`
- [ ] `insights.MetricAlert`
- [ ] `insights.Component`
- [x] `servicebus.Subscription`
- [x] `servicebus.Topic`
- [x] `servicebus.Queue`
- [x] `servicebus.Rule`
- [ ] `managedidentity.UserAssignedIdentity`
- [ ] `containerservice.AgentPool`
- [ ] `app.ContainerApp`
- [x] `documentdb.SqlResourceSqlContainer`
- [x] `documentdb.DatabaseAccount`
- [x] `documentdb.SqlResourceSqlDatabase`
- [ ] `operationalinsights.Workspace`
- [ ] `apimanagement.ApiOperationPolicy`
- [ ] `apimanagement.Api`

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3552

danielrbradley:
Ideal resource coverage:
- `resources.ResourceGroup`
[...]

PR #3760, opened 2024-12-09
Switch to PulumiTest

danielrbradley:
Run existing provider e2e tests via PulumiTest instead of ProgramTest to improve readability and clarity of logs.

Part of https://github.com/pulumi/pulumi-azure-native/issues/3421

### DIFF
--- a/provider/pkg/provider/provider_e2e_test.go
+++ b/provider/pkg/provider/provider_e2e_test.go
@@ -67,7 +67,7 @@ func newPulumiTest(t *testing.T, testProgramDir string) *pulumitest.PulumiTest {
 	dir := filepath.Join("test-programs", testProgramDir)
 	azureLocation := getLocation()
 	rpFactory := providers.ResourceProviderFactory(providerServer)
-	attachOpt := opttest.AttachProvider("azure", rpFactory)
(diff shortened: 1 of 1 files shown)


provider/pkg/provider/provider_e2e_test.go
```diff
@@ -7,123 +7,86 @@
...
-		},
-	}
-	for _, editDir := range editDirs {
-		test.EditDirs = append(test.EditDirs, integration.EditDir{
-			Dir: filepath.Join(cwd, "test-programs", editDir),
-		})
-	}
-	return &test, nil
+	rpFactory := providers.ResourceProviderFactory(providerServer)
+	attachOpt := opttest.AttachProvider("azure", rpFactory)
```
thomas11, line 70: Not sure if the name here is just for logs but the provider's official name is "azure-native" (passed to `provider.Serve`).

thomas11 (approved):

provider/pkg/provider/provider_e2e_test.go
danielrbradley: Yes, good catch. Not sure exacly what it would do, but quite possible it would just use an auto-downloaded version of the provider.

PR #3762, opened 2024-12-09
Cleanup: duplicate import and deprecated func

thomas11:
Just a simple removal of warnings.


iwahbe (approved):
If both `schema` and `pschema` were available, I would default to the canonical name: `schema`.

Either is better then both though.

danielrbradley (approved):

Issue #3642, opened 2024-10-11
Sign provider binaries for Windows


mikhailshilkov:
@TechWatching reports:

The azure native provider (the pulumi-resource-azure-native.exe) is blocked by some security software on Windows  because it is not signed. I think it's a big issue because many companies have tools that will prevent them using Pulumi because of that.

Original error example is in French:

```
Action risquée bloquée Basse ˅
Wakers a bloqué cette action.
Application ou processus bloqué : explorer.exe
Bloqué par : Réduction de la surface d'attaque
Règle : Bloquer l'exécution de fichiers exécutables, sauf s'ils respectent un critère de prévalence, d'âge ou de liste approuvée
Éléments affectés : C:\Users\MyUser\.pulumi\plugins\resource-azure-native-v2.66.0\pulumi-resource-azure-native.exe
```

Translation:

```
Risky action blocked Low ˅
Wakers blocked this action.
Blocked application or process: explorer.exe
Blocked by: Attack surface reduction
Rule: Block executable files from running unless they meet a prevalence, age, or trusted list criterion
Affected items: C:\Users\MyUser\.pulumi\plugins\resource-azure-native-v2.66.0\pulumi-resource-azure-native.exe
```

How does this compare to other providers?

thomas11:
Parent: pulumi/home#3682

TechWatching:
Any progress on this? Depending on the security software in place in companies, that can be a huge red flag for some.

mikub:
It is surprising to see this issue closed, as as of now all the binaries are still unsigned. The only thing that is signed is the MSI installer, which is still kind of worthless.

PR #3764, opened 2024-12-09
Sign Windows binaries

blampe:
Port of https://github.com/pulumi/ci-mgmt/pull/1202.

The build step is re-used by the subsequent dist step, so overall clock time is unchanged around ~8m50s.

Fixes https://github.com/pulumi/pulumi-azure-native/issues/3642.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ CODEGEN         := pulumi-gen-azure-native
 WORKING_DIR     := $(shell pwd)
 
 PROVIDER_PKG    := $(shell find provider/pkg -type f)
-SPECS           := $(shell find azure-rest-api-specs/specification/*/resource-manager -type f -name "*.json" ! -path "**/examples/**")
+SPECS           := .git/modules/azure-rest-api-specs/HEAD
 
 # Fail fast if the specs submodule doesn't exist
 ifeq (,$(SPECS))
@@ -31,10 +31,12 @@ endif
 # Input during CI using `make [TARGET] PROVIDER_VERSION=""` or by setting a PROVIDER_VERSION environment variable
 # Local builds will just used this fixed default version unless specified
 PROVIDER_VERSION ?= 2.0.0-alpha.0+dev
-# Ensure the leading "v" is removed - use this normalised version everywhere rather than the raw input to ensure consistency.
-# These variables are lazy (no `:`) so they're not calculated until after the dependency is installed
-VERSION_GENERIC = $(shell bin/pulumictl convert-version -l generic -v "$(PROVIDER_VERSION)")
-VERSION_FLAGS   = -ldflags "-X github.com/pulumi/pulumi-azure-native/v2/provider/pkg/version.Version=${VERSION_GENERIC}"
+# Check version doesn't start with a "v" - this is a common mistake
+ifeq ($(shell echo $(PROVIDER_VERSION) | cut -c1),v)
+$(error PROVIDER_VERSION should not start with a "v")
+endif
+
+VERSION_FLAGS   = -ldflags "-X github.com/pulumi/pulumi-azure-native/v2/provider/pkg/version.Version=${PROVIDER_VERSION}"
 MAJOR_VERSION   = $(shell echo $(PROVIDER_VERSION) | cut -d. -f1)
 PREVIOUS_MAJOR_VERSION = $(shell echo $(MAJOR_VERSION)-1 | bc)
 NEXT_MAJOR_VERSION = $(shell echo $(MAJOR_VERSION)+1 | bc)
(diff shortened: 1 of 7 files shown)


.github/workflows/build-test.yml
```diff
@@ -366,6 +366,27 @@ jobs:
           make --touch codegen schema
           make provider_prebuild
 
+      - name: Build and sign windows provider
+        run: |
```
thomas11, line 370: Shouldn't we use `set -euxo pipefail` or similar here to detect issues and fail early?

.github/workflows/build-test.yml
```diff
@@ -366,6 +366,27 @@ jobs:
...
+          make bin/windows-amd64/pulumi-resource-azure-native.exe;
+
+          az login --service-principal \
+            -u ${{ secrets.AZURE_SIGNING_CLIENT_ID }} \
```
thomas11, line 374: [nit] long `--` args are more readable in scripts

.github/workflows/build-test.yml
```diff
@@ -366,6 +366,28 @@ jobs:
           make --touch codegen schema
           make provider_prebuild
 
+      - name: Build and sign windows provider
+        shell: bash # Runs with -eo pipefail
+        run: |
+          make bin/windows-amd64/pulumi-resource-azure-native.exe;
```
danielrbradley, line 372: So, this is building just the windows binary first, then relying on the `make dist` from skipping the rebuild as it should be up-to-date? 

Slightly more tempted to move this into the makefile given the implicit assumptions on the target associations here. Might be able to spike on this myself now you've done the legwork of the actual signing code.

danielrbradley: Quick hack of what the extra code in the platform-specific provider binary target might look like:

```makefile
ifeq ($(TARGET),windows-amd64)
ifeq (findstring ||,|${AZURE_SIGNING_CLIENT_ID}|${AZURE_SIGNING_CLIENT_SECRET}|${AZURE_SIGNING_TENANT_ID}|)
	echo "Skipping signing as Azure credentials are not set (AZURE_SIGNING_CLIENT_ID, AZURE_SIGNING_CLIENT_SECRET, AZURE_SIGNING_TENANT_ID)"
else
	az login --service-principal \
            --username ${{ secrets.AZURE_SIGNING_CLIENT_ID }} \
            --password ${{ secrets.AZURE_SIGNING_CLIENT_SECRET }} \
            --tenant ${{ secrets.AZURE_SIGNING_TENANT_ID }} \
            --output none && \
		wget https://github.com/ebourg/jsign/releases/download/6.0/jsign-6.0.jar && \
		java -jar jsign-6.0.jar \
             --storetype AZUREKEYVAULT \
             --keystore "PulumiCodeSigning" \
             --url ${{ secrets.AZURE_SIGNING_KEY_VAULT_URI }} \
             --storepass "$(az account get-access-token --resource "https://vault.azure.net" | jq -r .accessToken)" \
             bin/windows-amd64/pulumi-resource-azure-native.exe && \
		az logout
endif
endif
```

danielrbradley (approved):
Tested the makefile alternative for a little bit but can't get it working yet, so happy to ship this as-is.

.github/workflows/build-test.yml
danielrbradley: Ok, new bash version seems like it's half way there ... can't test fully due to no cred though:

```makefile
bin/%/$(PROVIDER) bin/%/$(PROVIDER).exe: bin/pulumictl .make/provider_mod_download .make/prebuild provider/cmd/$(PROVIDER)/*.go .make/provider_prebuild $(PROVIDER_PKG)
	@# check the TARGET is set
	test $(TARGET)
	cd provider && \
		export GOOS=$$(echo "$(TARGET)" | cut -d "-" -f 1) && \
		export GOARCH=$$(echo "$(TARGET)" | cut -d "-" -f 2) && \
		CGO_ENABLED=0 go build -o ${WORKING_DIR}/$@ $(VERSION_FLAGS) $(PROJECT)/v2/provider/cmd/$(PROVIDER)
	if [[ "${TARGET}" = "windows-amd64" && "|${AZURE_SIGNING_CLIENT_ID}|${AZURE_SIGNING_CLIENT_SECRET}|${AZURE_SIGNING_TENANT_ID}|" == *"||"* ]]; then \
			echo "Skipping signing as Azure credentials are not set (AZURE_SIGNING_CLIENT_ID, AZURE_SIGNING_CLIENT_SECRET, AZURE_SIGNING_TENANT_ID)"; \
	else \
		az login --service-principal \
				--username ${AZURE_SIGNING_CLIENT_ID} \
				--password ${AZURE_SIGNING_CLIENT_SECRET} \
				--tenant ${AZURE_SIGNING_TENANT_ID} \
				--output none; \
		wget https://github.com/ebourg/jsign/releases/download/6.0/jsign-6.0.jar --output-document=jsign-6.0.jar; \
		java -jar jsign-6.0.jar \
			--storetype AZUREKEYVAULT \
			--keystore "PulumiCodeSigning" \
			--url ${AZURE_SIGNING_KEY_VAULT_URI} \
			--storepass "$$(az account get-access-token --resource "https://vault.azure.net" | jq -r .accessToken)" \
			bin/windows-amd64/pulumi-resource-azure-native.exe; \
		az logout; \
	fi
```

danielrbradley: Pushed as alternative PR here: https://github.com/pulumi/pulumi-azure-native/pull/3774

PR #3771, opened 2024-12-10
[v3] Make PULUMI_FORCE_NEW_FROM_SUBTYPES the default

### LINKED ISSUE #3006: Make PULUMI_FORCE_NEW_FROM_SUBTYPES the default
#2970 extended the setting of `forceNew`, which causes resources to be replaced on certain property changes, to sub-properties. Since this is potentially disruptive, the new behavior is gated behind an environment variable `PULUMI_FORCE_NEW_FROM_SUBTYPES`.

In v3, this should be default as it's more correct behavior, and we can call it out in the change notes.

EronWright:
Enables the functionality of `PULUMI_FORCE_NEW_FROM_SUBTYPES` by default in v3. In v2, behavior is unchanged. In any version, setting `PULUMI_FORCE_NEW_FROM_SUBTYPES` overrides the default behavior.

Closes #3006

### DIFF
--- a/sdk/python/README.md
+++ b/sdk/python/README.md
@@ -60,12 +60,6 @@ quality and higher fidelity with the Azure platform.
 
 To learn how to configure credentials refer to the [Azure configuration options](https://www.pulumi.com/registry/packages/azure-native/installation-configuration/#configuration-options).
 
-## Other Configuration
-
-In addition to the configuration options in the official documentation linked above, the following environment variables can be used to tweak lower-level behavior of the provider:
-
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):

provider/pkg/version/version.go
```diff
@@ -2,5 +2,11 @@
 
 package version
 
+import "github.com/blang/semver"
+
 // Version is initialized by the Go linker to contain the semver of this build.
 var Version string
+
+func GetVersion() semver.Version {
+	return semver.MustParse(Version)
```
danielrbradley, line 11: If the Version string is empty (e.g. for unit tests), this should probably default to the same as the makefile (`2.0.0-alpha.0+dev`)

Issue #3754, opened 2024-12-05
Remove V2 code


EronWright:
This ticket tracks the (eventual) removal of "v2" code from the codebase.  We may act on this once we're confident that no more 2.x builds would be needed (e.g. for a security patch) from the `master` branch.  Consider making a "2.x" branch beforehand as a best practice.

Specifics:

- [ ] Remove "v2" type aliasing: https://github.com/pulumi/pulumi-azure-native/pull/3749#pullrequestreview-2482271145

PR #3772, opened 2024-12-10
[v3] Remove __inputs in state

### LINKED ISSUE #2686: Remove `__inputs` in state

EronWright:
This PR implements the next stage of removing the `__inputs` from state.  In this stage, the provider has "v2" and "v3" behavior. In "v2" mode, the behavior is unchanged.  In "v3" mode, `Configure` asserts that the engine is passing the old inputs, and doesn't write `__inputs` to state. The fallback code that reads the `__inputs` from state is effectively inoperative in the "v3" mode.

Added a line item to this ticket for the final phase which is to remove the "v2" code:
https://github.com/pulumi/pulumi-azure-native/issues/3754

Closes #2686

### DIFF
--- a/provider/pkg/version/version.go
+++ b/provider/pkg/version/version.go
@@ -8,5 +8,10 @@ import "github.com/blang/semver"
 var Version string
 
 func GetVersion() semver.Version {
-	return semver.MustParse(Version)
+	v := Version
+	if v == "" {
+		// fallback to a default version e.g. for unit tests (see: PROVIDER_VERSION in Makefile)
+		v = "2.0.0-alpha.0+dev"
+	}
+	return semver.MustParse(v)
 }


provider/pkg/version/version.go
```diff
@@ -2,5 +2,11 @@
...
 package version
 
+import "github.com/blang/semver"
+
 // Version is initialized by the Go linker to contain the semver of this build.
 var Version string
+
+func GetVersion() semver.Version {
+	return semver.MustParse(Version)
+}
```
danielrbradley, line 12: Same as other PR .. this needs a default for unit tests.

danielrbradley (approved):
Approach looks sound to me 👍

PR #3774, opened 2024-12-11
Sign Windows binaries

danielrbradley:
Sign windows binary via make and simplify CI. This makes it possible to produce signed binaries locally, if required.
- Print when not signing with instructions.
- Rename the unsigned binary if attempting to sign to avoid the target being up-to-date if a step fails so it's easier to retry.
- Download the jar to the bin directory so it's not committed.
- Set the required variables via ENV in CI for the duration of the dist target.

Alternative to https://github.com/pulumi/pulumi-azure-native/pull/3764

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -366,30 +366,13 @@ jobs:
           make --touch codegen schema
           make provider_prebuild
 
-      - name: Build and sign windows provider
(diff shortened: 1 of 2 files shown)


thomas11 (approved):
LGTM

blampe (approved):
Sure

PR #3775, opened 2024-12-11
Remove generic version calculation

danielrbradley:
Remove unrequired complexity of parsing the version through pulumictl and the fiddly dependency because it's in a variable but uses a binary downloaded through a target.

Instead, just validate that the version isn't in the wrong format with a leading "v".

## Speed boost

Speed up makefile initialisation by not scanning for updates of every individual spec file but instead just check the timestamp of the git internal file tracking what upstream HEAD is pointing at. This doesn't account for manual edits to the specs - this is not supported behaviour if using the default makefile build caching.

## Remove SDK version injection

Use the same approach as bridged providers and just rely on `gen-sdk` to insert the version into the SDK which it's already been doing since enabling `respectSchemaVersion` in the schema for each language.


thomas11 (approved):
awesome speed boost!

PR #3776, opened 2024-12-11
Remove pulumictl and java-gen

danielrbradley:
Stacked on #3775 

- Use java plugin bundled with the Pulumi CLI via `pulumi package gen-sdk` command.
- Insert the real version into the generated `build.gradle` to match other SDKs.

Comparing all non-committed files locally, the only difference is the "generated by" header comment changing from "java-gen" to "pulumi".

It appears that the only way to configure the java gen to include the gradle setup is by adding two fields to the java lanaguge config in the schema.

Depended on https://github.com/pulumi/pulumi-java/pull/1501 being released as part of the Pulumi CLI which is now complete in 3.143.0

### DIFF
--- a/provider/pkg/gen/__snapshots__/gen_aliases_test_v2.snap
+++ b/provider/pkg/gen/__snapshots__/gen_aliases_test_v2.snap
@@ -325,6 +325,8 @@
    "rootPackageName": "pulumiazurenativesdk"
   },
   "java": {
+   "buildFiles": "gradle",
+   "gradleNexusPublishPluginVersion": "2.0.0",
    "packages": {
     "aadiam/v20200301": "aadiam.v20200301",
     "azure-native": "azurenative"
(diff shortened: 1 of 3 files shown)


t0yv0:
Not super sure. Wonder which version of the javagen code the new invocation is calling, though sounds like that code is present on latest.

danielrbradley:
This should be shippable once the pulumi-java then pulumi (with the upgraded dependency) are released.

danielrbradley:
This is still failing due to https://github.com/pulumi/pulumi-java/issues/1508

thomas11:
> This is still failing due to:
> 
> * [dedupeTypes not run from Pulumi CLI generation pulumi-java#1508](https://github.com/pulumi/pulumi-java/issues/1508)

@danielrbradley the issue was fixed and shipped p-java v1.0.0.

danielrbradley:
Yup, hoping to get time to finish this up soon

PR #3779, opened 2024-12-11
Implement autonaming configuration protocol

mikhailshilkov:
This PR implement the Azure Native part of https://github.com/pulumi/pulumi/issues/1518. See https://github.com/pulumi/pulumi/discussions/17592 for the full design.

In short, https://github.com/pulumi/pulumi/pull/17810 introduced the protobuf changes required for the provider-side implementation. With those, a provider can:

- Declare that it supports autonaming configurations with a response flag in Configure
- Accept two extra properties in CheckRequest: a proposed name and a mode to apply it

This PR applies those parameters to auto-name calculation if they are specified:

- In Disabled mode, we don't calculate auto-names and let the resource validation fail if an explicit name is not specified
- In Enforce mode, we use whatever ProposedName is supplied by the engine
- In Propose mode, almost the same as Enforce with exception of calculating a UUID for UUID fields and ignoring a proposed name

Added unit tests to cover most autonaming cases (old or new). Also, added an end-to-end test in YAML.

Resolves https://github.com/pulumi/pulumi-azure-native/issues/3777

### DIFF
--- a/provider/pkg/provider/provider_e2e_test.go
+++ b/provider/pkg/provider/provider_e2e_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/util"
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/version"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/pulumi/providertest/providers"
 	"github.com/pulumi/providertest/pulumitest"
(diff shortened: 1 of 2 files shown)


provider/pkg/provider/provider.go
```diff
@@ -550,13 +551,39 @@ func (k *azureNativeProvider) getDefaultLocation(ctx context.Context, olds, news
...
+			// with a missing required property error.
+			return nil, false, false
+		case rpc.CheckRequest_AutonamingOptions_ENFORCE:
+			return result(autonaming.ProposedName), true, true
+		case rpc.CheckRequest_AutonamingOptions_PROPOSE:
+			switch strategy {
+			case resources.AutoNameRandom:
+				fallthrough
+			case resources.AutoNameCopy:
+				return result(autonaming.ProposedName), true, true
```
danielrbradley, line 582: Not seen `fallthrough` in Go code before .. assume this is the same as:

```suggestion
			case resources.AutoNameRandom, resources.AutoNameCopy
				return result(autonaming.ProposedName), true, true
```

examples/examples_yaml_test.go
```diff
@@ -0,0 +1,31 @@
...
+				assert.True(t, ok)
+				assert.Contains(t, rgname, "autonaming-rg-") // project + name + random suffix
+				saname, ok := stackInfo.Outputs["saname"].(string)
+				assert.True(t, ok)
+				assert.Contains(t, saname, "autonamingsa") // project + name + random suffix, no dashes
+			},
+		})
+
+	integration.ProgramTest(t, &test)
+}
```
danielrbradley, line 31: We've already got yaml based tests set up in the [provider/pkg/provider_e2e_test.go](https://github.com/pulumi/pulumi-azure-native/blob/5ec5f1388dbba8fd5fa38bd1ed48de3cd715f16c/provider/pkg/provider/provider_e2e_test.go#L26-L32) which leverages the better pulumitest library to allow debugging from the integration test into the provider. I'd prefer to put these tests there rather than adding more example tests which aren't examples.

danielrbradley (commented):
Core implementation looks okay, would like to see the tests alongside where our existing yaml test are.

examples/examples_yaml_test.go
mikhailshilkov: Ah, thank you for the hint, done!

danielrbradley (approved):

PR #3783, opened 2024-12-12
Replace verbose azcore errors with more concise ones

### LINKED ISSUE #3778: Storage Account creation error prints the entire HTTP response
### What happened?

The following program:

```ts
const rg = new resources.ResourceGroup("rg");
const sa = new storage.StorageAccount("MySa", {
    resourceGroupName: rg.name,
    sku: {
        name: "Standard_LRS",
    },
    kind: "StorageV2",
});
```

fails to create a storage account, as expected. However, it prints out a very verbose error:

```
error: PUT https://management.azure.com/subscriptions/<redacted>/resourceGroups/rg-ce2m/providers/Microsoft.Storage/storageAccounts/MySa6198fc68
    --------------------------------------------------------------------------------
    RESPONSE 400: 400 Bad Request
    ERROR CODE: AccountNameInvalid
    --------------------------------------------------------------------------------
    {
      "error": {
        "code": "AccountNameInvalid",
        "message": "MySa6198fc68 is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only."
      }
    }
    --------------------------------------------------------------------------------
```

This is likely a consequence of azcore migration. The earlier version would print

```
error: Code="AccountNameInvalid" Message="MySa6198fc68 is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only."
```

### Example

See above

### Output of `pulumi about`

```
@pulumi/azure-native      2.76.0
@pulumi/pulumi            3.140.0
```

### Additional context

_No response_

### Contributing

thomas11:
Fixes #3778

Before:
```
$ pulumi up --skip-preview
...
Diagnostics:
  pulumi:pulumi:Stack (azn-http-errors-dev):
    error: update failed

  azure-native:storage:StorageAccount (sa):
    error: PUT https://management.azure.com/subscriptions/123/resourceGroups/resourceGroup84362436/providers/Microsoft.Storage/storageAccounts/INVALID
    --------------------------------------------------------------------------------
    RESPONSE 400: 400 Bad Request
    ERROR CODE: AccountNameInvalid
    --------------------------------------------------------------------------------
    {
      "error": {
        "code": "AccountNameInvalid",
        "message": "INVALID is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only."
      }
    }
    --------------------------------------------------------------------------------

...
```

After:
```
$ PATH=$HOME/pulumi/pulumi-azure-native/bin:$PATH pulumi up --skip-preview
...
  azure-native:storage:StorageAccount (sa):
    error: Code="AccountNameInvalid" Message="INVALID is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only."

...
```

### DIFF
--- a/provider/pkg/azure/client_azcore.go
+++ b/provider/pkg/azure/client_azcore.go
@@ -4,6 +4,7 @@ package azure
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -21,6 +22,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/log"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/util"
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/version"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/logging"
 )
@@ -167,7 +169,7 @@ func (c *azCoreClient) Get(ctx context.Context, id, apiVersion string, queryPara
 		return nil, err
 	}
 	if !runtime.HasStatusCode(resp, http.StatusOK) {
-		return nil, runtime.NewResponseError(resp)
+		return nil, newResponseError(resp)
 	}
 
 	var responseBody map[string]interface{}
@@ -218,7 +220,7 @@ func (c *azCoreClient) Delete(ctx context.Context, id, apiVersion, asyncStyle st
 	}
 
 	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusAccepted, http.StatusNoContent, http.StatusNotFound) {
-		return runtime.NewResponseError(resp)
+		return newResponseError(resp)
 	}
 	return err
 }
@@ -259,7 +261,7 @@ func (c *azCoreClient) putOrPatch(ctx context.Context, method string, id string,
 		return nil, false, err
 	}
 	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated, http.StatusAccepted) {
-		return nil, false, runtime.NewResponseError(resp)
+		return nil, false, newResponseError(resp)
 	}
 
 	var outputs map[string]any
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):

provider/pkg/azure/client_azcore.go
```diff
@@ -517,3 +519,37 @@ func (r *requestAssertingTransporter) Do(req *http.Request) (*http.Response, err
 	r.assertions(r.t, req)
 	return nil, nil
 }
+
+func newResponseError(resp *http.Response) error {
```
mikhailshilkov, line 523: Do we want to call it in shouldRetryConflict too?

provider/pkg/azure/client_azcore.go
```diff
@@ -517,3 +519,37 @@ func (r *requestAssertingTransporter) Do(req *http.Request) (*http.Response, err
...
+	}
+
+	body, err2 := runtime.Payload(resp)
+	if err2 != nil {
+		return err
+	}
+
+	var payload map[string]any
+	if err2 := json.Unmarshal(body, &payload); err2 != nil {
+		return err
```
mikhailshilkov, line 537: If body is not JSON (e.g. a string), do we really want to return a JSON deserialization error to the user? Shouldn't we show the string from the body?

provider/pkg/azure/client_azcore.go
```diff
@@ -517,3 +519,37 @@ func (r *requestAssertingTransporter) Do(req *http.Request) (*http.Response, err
...
+	errMsg := string(body)
+	if e, ok := util.GetInnerMap(payload, "error"); ok {
+		if msg, ok := e["message"]; ok {
+			errMsg = msg.(string)
+		}
+	}
+
+	errCode := azcoreErr.ErrorCode
+	if errCode == "" {
+		errCode = fmt.Sprintf("%d", resp.StatusCode)
```
mikhailshilkov, line 549: Should we just return a message in this case? Or is Code= Message= format important for some reason?

mikhailshilkov: Maybe it's useful to return a status code

provider/pkg/azure/client_azcore.go
mikhailshilkov: Ah, I misread it, it's `err` not `err2`

mikhailshilkov (approved):

provider/pkg/azure/client_azcore.go
thomas11: The error in `shouldRetryConflict` is never shown to the user. We only access its `ErrorCode`.

EronWright:
A possible follow-up is to handle the final error message of long-running operations, as returned by the poller as an `azcore.ResponseError`.

PR #3801, opened 2024-12-17
[v3] Skip flattening if properties would clash

### LINKED ISSUE #3195: Skip flattening if properties would clash
Draft PR: https://github.com/pulumi/pulumi-azure-native/pull/3016

Skip the flattening of nested properties indicated by [x-ms-client-flatten](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-client-flatten) if it would lead to overwriting a property, creating incorrect schema and SDKs. This case happens when inner and outer property have the same name, for a report on all occurrences see #3013.

This change is breaking and could therefore only be applied to v3 of the provider. The purpose of the PR is mainly to illustrate the SDK changes that would result.

Note: remember to remove exception added in #3007.

thomas11:
Skip the flattening of nested properties indicated by [x-ms-client-flatten](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-client-flatten) if it would lead to overwriting a property, creating incorrect schema and SDKs. This case happens when inner and outer property have the same name. For a report on all occurrences see #3013.

This change is breaking and could therefore only be applied to v3 of the provider. 

Resolves #3195

### DIFF
--- a/provider/pkg/gen/properties.go
+++ b/provider/pkg/gen/properties.go
@@ -42,6 +42,17 @@ type propertyBag struct {
 	requiredContainers RequiredContainers
 }
 
+// propertyIntersection returns the property names that are present in both bags.
+func (p *propertyBag) propertyIntersection(other *propertyBag) []string {
+	result := []string{}
+	for propName := range p.properties {
+		if _, ok := other.properties[propName]; ok {
+			result = append(result, propName)
+		}
+	}
+	return result
+}
+
 type RequiredContainers [][]string
 
 // genPropertiesVariant is a set of flags that control the behavior of property generation
@@ -121,7 +132,7 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
 			flatten = false
 		}
 
-		// Prevent a property collision due to flatterning. From v3, we detect and avoid this case automatically.
+		// Prevent a property collision due to flattening. From v3, we detect and avoid this case automatically.
 		if version.GetVersion().Major < 3 {
 			if m.resourceName == "DefenderForStorage" && (name == "malwareScanning" || name == "sensitiveDataDiscovery") {
 				flatten = false
@@ -135,7 +146,7 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
 
 				// Check that none of the inner properties already exists on the outer type. This
 				// causes a conflict when flattening, and will be handled in v3.
-				for _, propName := range propertyUnion(result, bag) {
+				for _, propName := range result.propertyIntersection(bag) {
 					m.flattenedPropertyConflicts[fmt.Sprintf("%s.%s", name, propName)] = struct{}{}
 				}
 
@@ -170,7 +181,7 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
 				}
 
 				// Check that none of the inner properties already exists on the outer type.
-				conflictingProperties := propertyUnion(result, bag)
+				conflictingProperties := result.propertyIntersection(bag)
 				if len(conflictingProperties) == 0 {
 					// Adjust every property to mark them as flattened.
 					newProperties := map[string]resources.AzureAPIProperty{}
@@ -300,17 +311,6 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
 	}
 
 	return result, nil
-}
-
-// propertyUnion returns the property names that are present in both bags.
-func propertyUnion(outerBag, innerBag *propertyBag) []string {
-	result := []string{}
-	for propName := range innerBag.properties {
-		if _, ok := outerBag.properties[propName]; ok {
-			result = append(result, propName)
-		}
-	}
-	return result
 }
 
 func (m *moduleGenerator) genProperty(name string, schema *spec.Schema, context *openapi.ReferenceContext, resolvedProperty *openapi.Schema, variants genPropertiesVariant) (*pschema.PropertySpec, *resources.AzureAPIProperty, error) {
--- a/provider/pkg/gen/properties_test.go
+++ b/provider/pkg/gen/properties_test.go
@@ -197,26 +197,26 @@ func TestNonObjectInvokeResponses(t *testing.T) {
 
 func TestPropertyUnion(t *testing.T) {
 	t.Run("no conflict", func(t *testing.T) {
-		outer := map[string]resources.AzureAPIProperty{
+		outer := propertyBag{properties: map[string]resources.AzureAPIProperty{
 			"foo": {},
-		}
-		inner := map[string]resources.AzureAPIProperty{
+		}}
+		inner := propertyBag{properties: map[string]resources.AzureAPIProperty{
 			"bar":  {},
 			"foo2": {},
-		}
-		assert.Empty(t, propertyUnion(&propertyBag{properties: outer}, &propertyBag{properties: inner}))
+		}}
+		assert.Empty(t, outer.propertyIntersection(&inner))
 	})
 
 	t.Run("conflict", func(t *testing.T) {
-		outer := map[string]resources.AzureAPIProperty{
+		outer := propertyBag{properties: map[string]resources.AzureAPIProperty{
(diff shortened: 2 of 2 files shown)


provider/pkg/gen/properties.go
```diff
@@ -260,6 +260,17 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
...
 
+func (m *moduleGenerator) hasConflictingPropertiesForFlattening(outerBag, innerBag *propertyBag, propertyName string, providerVersion semver.Version) bool {
+	hasConflict := false
+	for propName := range innerBag.properties {
+		if _, ok := outerBag.properties[propName]; ok {
+			m.flattenedPropertyConflicts[fmt.Sprintf("%s.%s", propertyName, propName)] = struct{}{}
+			hasConflict = true
+		}
+	}
+	return hasConflict && providerVersion.Major >= 3
```
mikhailshilkov, line 271: I would move the version check outside this function, it's somewhat confusing that the function result depends on the version. If possible, I would have a single `if version.GetVersion().Major < 3` where you already have it, and branch the logic entirely from that point on. Then add a TODO and open an issue to remove one branch once 3.0 lands. This feels cleaner than surgery to take it out from multiple places.

mikhailshilkov (commented):
Is there a way to see the effects in 3.0?

thomas11:
> Is there a way to see the effects in 3.0?

There _should_ be, by combining this change and the v3 branch and running schemagen, but at the moment that fails for unrelated reasons.

provider/pkg/gen/properties.go
thomas11: I agree, it's a bit weird. A complicating factor is that we want the report - the write to `m.flattenedPropertyConflicts` - in both cases. I'll look into refactoring this.

mikhailshilkov: I'd personally prefer code duplication (short lived until 3.0) over more complex logic, if that's possible.

provider/pkg/gen/properties.go
```diff
@@ -100,20 +102,22 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
...
+		flatten, ok := property.Extensions.GetBool(extensionClientFlatten)
 		isDict := resolvedProperty.AdditionalProperties != nil
 
 		// TODO: Remove when https://github.com/Azure/azure-rest-api-specs/pull/14550 is rolled back
 		workaroundDelegatedNetworkBreakingChange := property.Ref.String() == "#/definitions/OrchestratorResourceProperties" ||
 			property.Ref.String() == "#/definitions/DelegatedSubnetProperties" ||
 			property.Ref.String() == "#/definitions/DelegatedControllerProperties"
 
-		// TODO: can be removed when https://github.com/pulumi/pulumi-azure-native/pull/3016 is merged
-		if m.resourceName == "DefenderForStorage" && (name == "malwareScanning" || name == "sensitiveDataDiscovery") {
```
t0yv0, line 116: 🤯

provider/pkg/gen/properties.go
```diff
@@ -128,36 +132,32 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
...
 			}
 
-			// Check that none of the inner properties already exists on the outer type. This
-			// causes a conflict when flattening, and will probably need to be handled in v3.
-			for propName := range bag.properties {
-				if _, has := result.properties[propName]; has {
-					m.flattenedPropertyConflicts[fmt.Sprintf("%s.%s", name, propName)] = struct{}{}
+			// Check that none of the inner properties already exists on the outer type.
+			hasConflict := m.hasConflictingPropertiesForFlattening(result, bag, name, version.GetVersion())
+			if !hasConflict {
```
t0yv0, line 137: Is this the same as

```
if hasConflict {
   continue
}
```


But maybe this form would have a smaller textual diff?  Or is some code changing besides just going under `if`?

provider/pkg/gen/properties_test.go
```diff
@@ -194,3 +195,56 @@ func TestNonObjectInvokeResponses(t *testing.T) {
 		assert.NotContains(t, props.properties, resources.SingleValueProperty)
 	})
 }
+
+func TestHasConflictingPropertiesForFlattening(t *testing.T) {
```
t0yv0, line 199: This is great for the simple helper function but is there a way to see how this change in behavior plays out at a higher level? Or it's not worth it?

t0yv0 (approved):

t0yv0:
codecov actually indicates a few lines that might be interesting to check out, but I live up to you to decide what's worth it.

provider/pkg/gen/properties_test.go
thomas11: The problem is that the method we're in here is enormous and does many things and has many dependencies and side effects. It would be great to disentangle it but it's out of scope for now.

provider/pkg/gen/properties.go
thomas11: Welcome to our world :) This is just a sliver of the [customizations](https://github.com/pulumi/pulumi-azure-native/blob/master/playbooks/customizing-resources.md) we're doing.

provider/pkg/gen/properties_test.go
t0yv0: Yea my thought exactly. I'm a small method person myself if I can afford it.

provider/pkg/gen/properties.go
```diff
@@ -100,64 +101,105 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
...
 		// It's ok if a new API version is fixed, this is a no-op then.
 		if m.resourceName == "CIAMTenant" && name == "tier" && strings.HasPrefix(m.module, "azureactivedirectory") {
 			flatten = false
 		}
 
-		if (ok && flatten && !isDict) || workaroundDelegatedNetworkBreakingChange {
-			bag, err := m.genProperties(resolvedProperty, variants.noResponse())
-			if err != nil {
-				return nil, err
+		// Prevent a property collision due to flatterning. From v3, we detect and avoid this case automatically.
```
EronWright, line 124:
```suggestion
		// Prevent a property collision due to flattening. From v3, we detect and avoid this case automatically.
```

EronWright (commented):
Seems like some v2 code is moving around but isn't really changing much?  I wonder if the change could be done without re-organizing the v2 code. Like, don't change the indentation, and use a goto instead.

thomas11:
> Seems like some v2 code is moving around but isn't really changing much? I wonder if the change could be done without re-organizing the v2 code. Like, don't change the indentation, and use a goto instead.

I don't see how the v2 code was re-organized. I mostly added a new case for v3. The additional branching adds a level of indentation but other than that, the v2 case is mostly preserved. See also the [discussion here](https://github.com/pulumi/pulumi-azure-native/pull/3801#discussion_r1890088899).

The method as a whole is badly structured and way too long, but a proper refactoring was beyond the scope of this PR.

danielrbradley (commented):
I'm also struggling to parse the diff. I also wonder if a first refactor commit to pull out a function could make the actual change more clear and remove the need for the extra level of nesting.

provider/pkg/gen/properties.go
```diff
@@ -260,6 +302,17 @@ func (m *moduleGenerator) genProperties(resolvedSchema *openapi.Schema, variants
 	return result, nil
 }
 
+// propertyUnion returns the property names that are present in both bags.
+func propertyUnion(outerBag, innerBag *propertyBag) []string {
```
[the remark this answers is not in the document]
thomas11: ah, good catch :)

thomas11:
> I'm also struggling to parse the diff. I also wonder if a first refactor commit to pull out a function could make the actual change more clear and remove the need for the extra level of nesting.

I don't think this is a good investment of time. The code is quite convoluted and half of it (v2) will be removed soon anyways. It's more important here that there is a single code block to remove post v3 for simple cleanup, as Mikhail also pointed out above.

thomas11:
Superseded by #3839

PR #3802, opened 2024-12-17
azcore fix: initialize request pipeline with configured cloud

### LINKED ISSUE #3795: New AzCore API Breaks Environment

thomas11:
Fixes #3795, hopefully.

The way Azure's azcore and azidentity SDKs are designed makes it hard to test this, but reading [the source here](https://github.com/Azure/azure-sdk-for-go/blob/sdk/azcore/v1.16.0/sdk/azcore/arm/runtime/pipeline.go#L60) it looks like this change should fix the issue.

### DIFF
--- a/provider/pkg/provider/provider_test.go
+++ b/provider/pkg/provider/provider_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider/crud"
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources"
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources/customresources"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/util"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


provider/pkg/azure/client_azcore.go
```diff
@@ -49,7 +49,11 @@ func NewAzCoreClient(tokenCredential azcore.TokenCredential, userAgent string, a
 	})
 
 	if opts == nil {
-		opts = &arm.ClientOptions{}
+		opts = &arm.ClientOptions{
+			ClientOptions: policy.ClientOptions{
+				Cloud: azureCloud,
+			},
+		}
```
mjeffryes, line 56: is there no test we could add? At a minimum, it seems like we could extract out a "InitClientOptions" helper and add a unittest?

We should maybe add a comment too, pointing out why it's necessary to forward the `cloud.Configuration` via opts?

mjeffryes (approved):
Approving b/c I don't want to block you on timezone delta and it makes sense that this could fix the issue, but would be nice to capture this requirement in some sort of test or at least a comment.

provider/pkg/azure/client_azcore.go
thomas11: I've extracted `InitClientOptions` and added a unit test for it, but that tests only the mechanics of the change, not that it fixes the user's problem. Still, additional coverage is always good.

PR #3815, opened 2024-12-23
Remove unused schemagen code, cleanup

thomas11:
- **Small schema.go cleanups, remove unused endpoints collection**
- **Remove unused warnings reporting**


iwahbe (commented):
Are we safe to remove warnings because they are never read, or because the code we see doesn't actually generate warnings?

thomas11:
> Are we safe to remove warnings because they are never read, or because the code we see doesn't actually generate warnings?

Should've mentioned that, sorry. Currently, the code doesn't generate any warnings.

iwahbe (approved):

provider/pkg/gen/schema.go
```diff
@@ -808,7 +800,6 @@ func (g *packageGenerator) genResourceVariant(apiSpec *openapi.ResourceSpec, res
 		inlineTypes:                map[*openapi.ReferenceContext]codegen.StringSet{},
 		nestedResourceBodyRefs:     nestedResourceBodyRefs,
 		flattenedPropertyConflicts: map[string]struct{}{},
-		allEndpoints:               g.allEndpoints,
```
danielrbradley, line 811: I think the definition for `allEndpoints` on line 1329 also needs removing

danielrbradley (approved):
Looks good, one trailing cleanup in there.

Issue #2753, opened 2023-09-13
Split DBforPostgreSQL into PostgreSQL and CosmosDBForPostgresSQL


mheers:
There's two distinct services from Azure:
1. [PostgreSQL](https://learn.microsoft.com/en-us/azure/templates/microsoft.dbforpostgresql/flexibleservers)
2. [Azure Cosmos DB for PostgreSQL](https://learn.microsoft.com/en-us/azure/templates/microsoft.dbforpostgresql/servergroupsv2)

Both of these services are using the Microsoft.DBforPostgreSQL namespace which results in us combining the types and resources. This means the resources `PrivateEndpointConnection` and `FirewallRule` conflict and one is ignored and is not available in the default namespace.

Move each of these services into their own module name to avoid the naming conflicts:
1. PostgreSQL
2. CosmosDBForPostgresSQL

---

### What happened?

I'm using Go and the azure-native-v2 provider and try to setup resources from dbforpostgresql. I recognized that:

- The [PrivateEndpointConnection](https://www.pulumi.com/registry/packages/azure-native/api-docs/dbforpostgresql/privateendpointconnection/#inputs) input just accepts a `ClusterName` (CosmosDB). There is no way to define a `ServerName` (FlexibleServer)

- The [FirewallRule](https://www.pulumi.com/registry/packages/azure-native/api-docs/dbforpostgresql/firewallrule/#inputs) input just accepts a `ServerName` (FlexibleServer). There is no way to define a `ClusterName` (CosmosDB)

It would be nice to have `ServerName` AND `ClusterName` in both structs.

### Expected Behavior

- Compatibility with https://learn.microsoft.com/en-us/rest/api/postgresqlhsc/firewall-rules/create-or-update?tabs=HTTP

### Steps to reproduce

```golang
pec, err := dbforpostgresql.NewFirewallRule(ctx, "my-cosmo", &dbforpostgresql.FirewallRuleArgs{
	ClusterName: pulumi.String("my-cluster"),
})
```

### Output of `pulumi about`

CLI          
Version      3.78.0
Go Version   go1.20.6
Go Compiler  gc

Plugins
NAME          VERSION
azure-native  2.6.0
azure-native  2.6.0
azure-native  2.6.0
azure-native  2.6.0
azure-native  2.6.0
go            unknown

Host     
OS       ubuntu
Version  20.04
Arch     x86_64

[...]

PR #3817, opened 2024-12-24
[v3] Detect resource path conflicts and disambiguate

### LINKED ISSUE #2495: Disambiguate resources with different paths in any API version
#2484 disambiguated path conflicts between the default versions of v1 and v2. There are further conflicts between other versions, however, that we should fix. See @danielrbradley's [comment](https://github.com/pulumi/pulumi-azure-native/pull/2484#pullrequestreview-1462353826) for details.

thomas11:
_This PR is based on #3815 and contains its commits._

#2495
#2753

Previously, schema and metadata generation of this provider assumed that a type token like `azure-native:netapp:Pool` uniquely identifies a resource. In the Azure spec, that's not always the case across API versions.

For instance, as shown by #2753, resources such as `PrivateEndpointConnection ` can point to the /servers/ API in one API version but to the newer /flexibleServers/ API, which is a completely different product underneath, in another version.

This PR does two things:
1. Detect such conflicts. So far, it only prints a warning when it finds one, but when all such conflicts are investigated and fixed it should error.
2. Disambiguate several known instances of such conflicts. In the above example of `PrivateEndpointConnection`, we rename the resource to `SingleServerPrivateEndpointConnection` when it points to the /servers/ API.

Any change to the generated schema is gated behind a "version >= 3" check to avoid breaking changes in v2.

### DIFF
--- a/versions/v3-removed.json
+++ b/versions/v3-removed.json
@@ -25,7 +25,8 @@
     "2017-07-14"
   ],
   "ApiManagement": [
-    "2016-07-07"
+    "2016-07-07",
+    "2016-10-10"
   ],
   "App": [],
   "AppComplianceAutomation": [],
@@ -382,7 +383,9 @@
   "Kusto": [
     "2017-09-07-privatepreview"
   ],
-  "LabServices": [],
+  "LabServices": [
+    "2018-10-15"
+  ],
   "LoadTestService": [],
   "Logic": [],
   "Logz": [],
@@ -430,7 +433,8 @@
   "MarketplaceOrdering": [],
   "Media": [],
   "Migrate": [
-    "2017-11-11-preview"
+    "2017-11-11-preview",
+    "2018-02-02"
   ],
   "MixedReality": [
     "2019-02-28-preview",
@@ -479,7 +483,8 @@
     "2018-02-01",
     "2018-03-01",
     "2018-03-01-preview",
-    "2018-04-01"
+    "2018-04-01",
+    "2021-05-01-preview"
   ],
   "NetworkCloud": [],
   "NetworkFunction": [],
@@ -584,6 +589,9 @@
   "SignalRService": [
     "2018-03-01-preview",
     "2018-10-01"
+  ],
+  "Sim": [
+    "2022-03-01-preview"
   ],
   "SoftwarePlan": [],
   "Solutions": [


provider/pkg/gen/schema.go
```diff
@@ -640,6 +645,8 @@ type packageGenerator struct {
 	forceNewTypes              []ForceNewType
 	flattenedPropertyConflicts map[string]map[string]struct{}
 	majorVersion               int
+	// A tok -> path map to detect when the same resource tok is mapped to different API paths.
+	pathConflicts map[string]map[string]struct{}
```
danielrbradley, line 649: Do we need to do something with these pathConflicts to print them as a report?

provider/pkg/gen/schema.go
```diff
@@ -777,16 +784,125 @@ func (g *packageGenerator) makeTypeAlias(alias string, apiVersion openapi.SdkVer
 	return pschema.AliasSpec{Type: &fqAlias}
 }
 
+// dedupResourceNameByPath returns a modified resource name (`typeName`) if the resource is mapped to multiple API
+// paths. For instance, the deprecated "single server" resources in `dbformysql` and `dbforpostgresql` are renamed
+// to `SingleServerResource`.
+// TODO,tkappler check each one if we can just get rid of an old API version instead of doing this.
+func dedupResourceNameByPath(provider, typeName, canonPath string) string {
```
danielrbradley, line 791: This method looks like reasonable overrides. Are you hoping to resolve the TODO before merging?

Also, do we need to add aliases at the point we rename the existing resource, or do we always rename only the new version of the resource name?

danielrbradley (commented):
A couple of comments inline. Looks like a good direction but maybe not quite finished yet.

provider/pkg/gen/schema.go
thomas11: We need to resolve the TODO before we release v3, not necessarily before we merge into the v2 master.

Good point on aliases - will check.

provider/pkg/gen/schema.go
```diff
@@ -777,16 +796,116 @@ func (g *packageGenerator) makeTypeAlias(alias string, apiVersion openapi.SdkVer
...
+}
+
 func (g *packageGenerator) genResourceVariant(apiSpec *openapi.ResourceSpec, resource *resourceVariant, nestedResourceBodyRefs []string, typeNameAliases ...string) error {
 	module := g.moduleName()
 	swagger := resource.Swagger
 	path := resource.PathItem
+	canonPath := paths.NormalizePath(resource.Path)
+
+	typeName := resource.typeName
+	if g.majorVersion > 3 {
```
thomas11, line 898: `>=` !!

danielrbradley:
Summary from conversation: add typed strings for `ModuleName, `ResourceName` and `ResourcePath`; and make the report a `map[ModuleName]map[ResourceName]map[ResourcePath]any`.

danielrbradley (approved):
Approving tentatively once one remaining comment resolved.

Issue #3448, opened 2024-07-19
Choose new default API versions


thomas11:
Depends on #2661.

Can be done iteratively.

danielrbradley:
First selection process is done. The next step is continuing testing and resolving any failing tests on the v3 branch.

Issue #690, opened 2021-03-23
Choose modules to structure according to Azure SDK not ARM Templates


mikhailshilkov:
At the moment, Azure-Native puts resources into modules that are named after ARM URL resource providers. Some of those names are outdated or confusing, and some resources clash because they end up in the same large module (like `Network`, e.g. #583).

In many aspects, Azure-Native follows Azure SDKs more closely than ARM Templates: naming resources after operations, flattening properties, renaming properties to SDK names. It would be logical to name modules similar to Azure SDKs too.

For example (old -> new):

- Network -> split out TrafficManager, PrivateDNS, FrontDoor, DNS
- DocumentDB -> Cosmos DB
- DBforPostgreSQL -> PostgreSQL
- DBforMySQL -> MySQL
- Cache -> split to Redis and RedisEnterprise
- Insights -> split to Monitor and ApplicationInsights
- Devices -> IoTHub
- StorSimple -> split to StorSimple1200Series and StorSimple8000Series

etc.

Even, [ARM template docs](https://docs.microsoft.com/en-us/azure/templates/microsoft.documentdb/cassandraclusters?tabs=json) use some of those names in the menu structure: Cosmos DB, DNS, FrontDoor etc.

Every SDK has its own set of mappings, done mostly with `readme.<language>.md` files. So, mappings for different SDKs differ.

One possibility is to use top-level folder names in the azure-rest-api-spec repository folder structure, which gives unambiguous mapping. They are all lowercase so we'd have to map to PascalCase manually.

Obviously, any module naming change is largely breaking.

I'm curious to learn other opinions about this potential change.

mikhailshilkov:
Here is a potential list of changes:

```
Insights->ApplicationInsights
Portal->CloudShell
DocumentDB->CosmosDB
Devices->DeviceProvisioningServices
DevTestLab->DevTestLabs
HybridData->HybridDataManager
Kubernetes->HybridKubernetes
VirtualMachineImages->ImageBuilder
Devices->IotHub
Management->ManagementGroups
DBforMariaDB->MariaDB
Media->MediaServices
OffAzure->Migrate
Migrate->MigrateProjects
Insights->Monitor
ManagedIdentity->Msi
DBforMySQL->MySql
DBforPostgreSQL->PostgreSql
DBforPostgreSQL->PostgreSqlHsc
PowerBI->PowerBIEmbedded
RecoveryServices->RecoveryServicesBackup
[...]

PR #3838, opened 2025-01-08
Customise module naming

danielrbradley:
Choosing default versions (https://github.com/pulumi/pulumi-azure-native/issues/3448) is hard where there's multiple versions mixed into a single module. This is sometimes because we structure our modules around Azure's namespaces rather than the SDK folders. This PR allows us to selectively restructure problematic modules to match Azure's SDK structure (https://github.com/pulumi/pulumi-azure-native/issues/690) and avoid mixing multiple versions into a single module. An example of the result of turning on the feature flag can be seen in 14db6d8eefea9e9b55a7e00e051cfda17fff0e4e where we split out Dns, DnsResolver, FrontDoor, PrivateDns and TrafficManager from the Network module.

Side note: when splitting the modules out, there's all previous versions generated for the new module. These will be reduced when we run the squeeze process.

Manual report of all specification folders and the namespaces within them: [spec-dirs.csv](https://github.com/user-attachments/files/18345878/spec-dirs.csv)

For the first Network module split, we've hard coded the new module names to use because the casing is not able to be derived from the folder name alone. In a follow-up PR we can investigate parsing the Readme.md files from the specifications to automatically infer the correct names and avoid hard-coding.

Generate implementation approach:
1. Make naming overrides a functions instead of a map so we can make decisions based on the version switch and other information.
2. When overriding a module name, return the old name too and add aliases so it's easy for users to migrate.

### DIFF
--- a/provider/pkg/gen/schema_test.go
+++ b/provider/pkg/gen/schema_test.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/go-openapi/spec"
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/openapi"
-	"github.com/pulumi/pulumi/pkg/v3/codegen"
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


provider/pkg/openapi/discover.go
```diff
@@ -569,11 +570,12 @@ func addResourcesAndInvokes(version VersionResources, fileLocation, path, provid
...
-			PathItem:     &pathItem,
-			Swagger:      swagger,
-			DefaultBody:  defaultBody,
-			PathItemList: pathItemList,
+			Path:                 path,
+			PathItem:             &pathItem,
+			Swagger:              swagger,
+			DefaultBody:          defaultBody,
+			PathItemList:         pathItemList,
+			PreviousProviderName: oldProvider,
```
thomas11, line 578: Should we add it to invokes as well?

provider/pkg/gen/schema.go
```diff
@@ -1106,27 +1106,32 @@ func (g *packageGenerator) genFunctions(typeName, path string, specParams []spec
...
-	return g.providerApiToModule(g.sdkVersion)
+	return providerApiToModule(g.provider, g.sdkVersion)
 }
 
-func (g *packageGenerator) versionedModuleName() string {
-	versionedModule := strings.ToLower(g.provider) + goModuleVersion
-	if g.sdkVersion == "" {
+// versionedGoModuleName produces the *Go* module name from the provider name and the version e.g. `compute/v20200701`.
+// or just the provider name if the version is empty (default version) e.g. `compute`.
+func versionedGoModuleName(provider openapi.ProviderName, sdkVersion openapi.SdkVersion) string {
```
thomas11, line 1134: I don't understand _"Prefer pure functions for simple string manipulation so the inputs are transparent to the caller."_. To me, the point of methods is that the caller doesn't need to know about all the inputs that it's not responsible for. `versionedGoModuleName(gen.provider, gen.sdkVersion)` is more complex and error-prone than `gen.versionedModuleName()`.

provider/pkg/gen/schema.go
```diff
@@ -1106,27 +1106,32 @@ func (g *packageGenerator) genFunctions(typeName, path string, specParams []spec
...
+		return strings.ToLower(provider)
 	}
-	return fmt.Sprintf("%s/%s", strings.ToLower(g.provider), apiVersion)
+	return fmt.Sprintf("%s/%s", strings.ToLower(provider), sdkVersion)
 }
 
-func (g *packageGenerator) generateTok(typeName string, apiVersion openapi.SdkVersion) string {
-	return fmt.Sprintf(`%s:%s:%s`, g.pkg.Name, g.providerApiToModule(apiVersion), typeName)
+// pulumiProviderName is the name of the provider as used in all tokens.
+const pulumiProviderName = "azure-native"
```
thomas11, line 1151: Searching for `"azure-native"`, we use this string in a few places. Can we put the const in a more central place and re-use it?

provider/pkg/resources/resources.go
```diff
@@ -258,36 +259,64 @@ var wellKnownProviderNames = map[string]string{
...
+}
+
+// For the cases below, we use folder (SDK) name for module names instead of the ARM name.
+var folderModulePattern = regexp.MustCompile(`.*/specification/([a-z]+)/resource-manager/.*`)
+var folderModuleNames = map[string]string{
+	"videoanalyzer": "VideoAnalyzer",
+	"webpubsub":     "WebPubSub",
+}
+
+// getNameOverride returns a name override for a given spec folder name, file provider, and API provider.
```
thomas11, line 291: Not sure why you call it "file provider" - why "file"?

provider/pkg/resources/resources.go
```diff
@@ -312,6 +341,15 @@ func resourceProvider(path, defaultValue string) string {
 	return defaultValue
 }
 
+func getSpecFolderName(path string, defaultValue string) string {
```
thomas11, line 344: Is `defaultValue` ever not `""`?

thomas11 (commented):
Overall, LGTM. Left a few questions and comments.

However, there's no test coverage for any of the new code that I can see? That's a must have, I think. Also Godoc's, though less important.

provider/pkg/gen/schema.go
danielrbradley: Yeah, would like to do this too. Couldn't immediately see a good module which makes sense for a global const and can be referenced from everywhere else. Only `util` or `version` could functionally work, but neither makes logical sense to me.

danielrbradley: Actually, I'll move this to the top of the file along with the other consts as a slight clean-up.

provider/pkg/gen/schema.go
danielrbradley: Philosophically I'd argue that's it's not more complex as functions, but we're making the complexity more visible to the caller rather than hiding it and making the user look at the implementation to understand where it comes from. Making them functions makes them trivial to test without needing mocking (though I know I still need to actually write the tests).

The original motivation was needing a `generateTok` function where I can control all three arguments rather than them being hard coded based on the contextual object.

- I've kept `g.moduleName()` as that's the most common use which is fairly obvious what it's composed of.
- The signatures should be strictly typed to avoid being error prone so you can't just pass any string in: `func versionedGoModuleName(provider openapi.ProviderName, sdkVersion openapi.SdkVersion)`.

I think what might be helpful here is to refactor one step further to include the `filepath.Join(goModuleRepoPath, ...)` within the function and rename it to `goModule(openApiProvider, sdkVersion)` so the call site is just:

```diff
- golangImportAliases[filepath.Join(goModuleRepoPath, versionedGoModuleName(gen.provider, gen.sdkVersion))] = strings.ToLower(providerName)
+ golangImportAliases[goModule(gen.provider, gen.sdkVersion)] = strings.ToLower(providerName)
```

provider/pkg/resources/resources.go
danielrbradley: Ah yeah, I just copied the pattern from the `resourceProvider` above, but in this case it doesn't need to vary.

provider/pkg/openapi/discover.go
danielrbradley: Invokes don't need migrating so I don't think we need the old name - they can just use the new name only.

danielrbradley:
@thomas11 changes since last review:

1. Fixed broken test (we deleted the method it was testing)
2. Removed the `getSpecFolderName` default arg.
3. Moved the `pulumiProviderName` const to the top of the file.
4. Tested and simplified `goModuleName()` (previously `versionedGoModuleName`)
5. Refactored and tested `getNameOverride()`.

I've squashed some of these into previous commits as I'll not squash this PR when merging.

thomas11:
> @thomas11 changes since last review:
> 
> 1. Fixed broken test (we deleted the method it was testing)
> 2. Removed the `getSpecFolderName` default arg.
> 3. Moved the `pulumiProviderName` const to the top of the file.
> 4. Tested and simplified `goModuleName()` (previously `versionedGoModuleName`)
> 5. Refactored and tested `getNameOverride()`.

Nice, thanks! The `PreviousProviderName` aliases in `generateAliases()` don't have test coverage yet, would that be reasonably easy to add?

danielrbradley:
> Nice, thanks! The `PreviousProviderName` aliases in `generateAliases()` don't have test coverage yet, would that be reasonably easy to add?

Ah yes, will add now 👍

danielrbradley:
Added last tests, rebased and updated example commit for the impact when enabled: 14db6d8eefea9e9b55a7e00e051cfda17fff0e4e

thomas11 (approved):

PR #3842, opened 2025-01-09
Fix azure.IsNotFound for new azcore error type

thomas11:
Fixes an issue I discovered today when trying to reproduce another issue report. `pulumi refresh` is broken for this provider:
```
$ PATH="$HOME/pulumi/pan/bin:$PATH" pulumi refresh --skip-preview -s dev

     pulumi:pulumi:Stack                      recoveryservices-protecteditem-dev  **failed**                1 error; 4 warnings
 ~   ├─ azure-native:resources:ResourceGroup  resourceGroup                       **refreshing failed**     1 error

    error: Code="ResourceGroupNotFound" Message="Resource group 'resourceGroup3f871416' could not be found."
```

The reason is that #3783 regressed `azure.IsNotFound`, causing `Read` to report an error when a resource doesn't exist in Azure.

This PR fixes the problem and adds test coverage.

### DIFF
--- a/provider/pkg/azure/azure.go
+++ b/provider/pkg/azure/azure.go
@@ -49,6 +49,11 @@ func IsNotFound(err error) bool {
 	if responseError, ok := err.(*azcore.ResponseError); ok {
 		return responseError.StatusCode == http.StatusNotFound
 	}
+
+	if responseError, ok := err.(*PulumiAzcoreResponseError); ok {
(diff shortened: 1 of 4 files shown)


danielrbradley (approved):

EronWright:
I have to wonder if using `errors.As` would have helped here. That is, wrap the underlying error to improve the formatting, and have client code use As consistently.

PR #3859, opened 2025-01-16
Remove default state of deleted resources

danielrbradley:
Tentative fix for failing unit test.

To confirm with @thomas11 - are we expecting these to be completely removed or should they just be renamed or apply to a specific version?


danielrbradley:
These should not need to be removed. Should be addressed with https://github.com/pulumi/pulumi-azure-native/pull/3861 instead

PR #3865, opened 2025-01-16
Make new fileshares discoverable as protectable items

thomas11:
This is a fix for this error:
```
{"error":{"code":"BMSUserErrorProtectedItemNameIncorrectFormat","message":"Protected Item name is not in the correct format."}}
```

This happens when we fail to look up the internal or system name of a fileshare.

As [this TF PR](https://github.com/hashicorp/terraform-provider-azurerm/pull/14238) shows, it seems that one needs tp call the pretty much undocumented /inquire API to make new fileshares discoverable to Recovery Services.

One problem is that the [docs](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/recoveryservices/armrecoveryservicesbackup/v4@v4.1.0#ProtectionContainersClient.Inquire) say
> Inquire - This is an async operation and the results should be tracked using location header or Azure-async-url.

but the SDK provides no way that I could see to actually track the operation. The "x-ms-location" header is unaccessible and the `Inquire` method returns an empty struct. As the next best thing, I'm waiting for up to ten minutes for the fileshare to appear.

### DIFF
--- a/provider/pkg/resources/customresources/custom_recoveryservices.go
+++ b/provider/pkg/resources/customresources/custom_recoveryservices.go
@@ -18,7 +18,8 @@ import (
 )
 
 const (
-	protectedItemPath = "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.RecoveryServices/vaults/{vaultName}/backupFabrics/{fabricName}/protectionContainers/{containerName}/protectedItems/{protectedItemName}"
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
Core change from a single retry at 30 seconds to up to 5 minutes of attempts looks reasonable, plus the extra ping to inquireContainer.

Is there any cost associated with the inquireContainer? Should we only call that if the first `readSystemNameFromProtectableItem` returns empty?

thomas11:
> Is there any cost associated with the inquireContainer? Should we only call that if the first `readSystemNameFromProtectableItem` returns empty?

In my tests it's been very fast, since it only kicks off an asynchronous operation.

Issue #3848, opened 2025-01-10
Add aliases for v3 case changes


danielrbradley:
Cases are changing due to changes between versions in the specifications. Add aliases to make migration seamless.

danielrbradley:
Likely affected resources:

```
IoTOperations/Dataflow
IoTOperations/DataflowEndpoint
IoTOperations/DataflowProfile
Workloads/SapApplicationServerInstance
Workloads/SapDatabaseInstance
Workloads/SapVirtualInstance
```

We should scan all tokens in the schema to find other cases.

danielrbradley:
Manual review of default schema diff:

Process:
1. Extract new and old docs schema from git (only contains default version)
2. Extract resource tokens: 
   ```shell
   jq -r '.resources | keys[]' [...]/schema.json > bin/tokens-[old|new].txt`
   ```
3. Generate diff:
   ```shell
   diff tokens-old.txt tokens-new.txt > bin/tokens-diff
   ```

Result:
```diff
3d2
< azure-native:aadiam:AzureADMetric
211d209
< azure-native:automanage:Account
216d213
< azure-native:automanage:ConfigurationProfilePreference
406d402
< azure-native:azurestackhci:EdgeDevice
409a406
> azure-native:azurestackhci:HciEdgeDevice
574d570
< azure-native:containerinstance:CGProfile
604d599
< azure-native:containerservice:OpenShiftManagedCluster
844a840,852
> azure-native:dns:DnssecConfig
> azure-native:dns:RecordSet
> azure-native:dns:Zone
> azure-native:dnsresolver:DnsForwardingRuleset
> azure-native:dnsresolver:DnsResolver
> azure-native:dnsresolver:DnsResolverDomainList
> azure-native:dnsresolver:DnsResolverPolicy
> azure-native:dnsresolver:DnsResolverPolicyVirtualNetworkLink
> azure-native:dnsresolver:DnsSecurityRule
> azure-native:dnsresolver:ForwardingRule
> azure-native:dnsresolver:InboundEndpoint
> azure-native:dnsresolver:OutboundEndpoint
> azure-native:dnsresolver:PrivateResolverVirtualNetworkLink
851d858
< azure-native:documentdb:FirewallRule
886d892
< azure-native:edgeorder:AddressByName
888d893
< azure-native:edgeorder:OrderItemByName
945a951
> azure-native:frontdoor:Policy
1057,1059c1063,1065
< azure-native:iotoperations:DataFlow
[...]

PR #3872, opened 2025-01-17
V3 alias refactoring

danielrbradley:
Try to get the right information into the right places in order to implement robust auto-aliasing to address https://github.com/pulumi/pulumi-azure-native/issues/3848

### DIFF
--- a/provider/pkg/gen/schema.go
+++ b/provider/pkg/gen/schema.go
@@ -284,11 +284,11 @@ func PulumiSchema(rootDir string, modules openapi.AzureModules, versioning Versi
 	resourcesPathTracker := newResourcesPathConflictsTracker()
 
 	for _, moduleName := range util.SortedKeys(modules) {
-		versionMap := modules[moduleName]
+		moduleVersions := modules[moduleName]
 
 		resourcePaths := map[openapi.ResourceName]map[string][]openapi.ApiVersion{}
 
-		versions := util.SortedKeys(versionMap)
+		versions := util.SortedKeys(moduleVersions)
 		// The version in the parsed module is "" for the default version.
 		for _, moduleVersion := range versions {
 			var sdkVersion openapi.SdkVersion
@@ -330,7 +330,7 @@ func PulumiSchema(rootDir string, modules openapi.AzureModules, versioning Versi
 			golangImportAliases[goModuleName(gen.moduleName, gen.sdkVersion)] = moduleName.Lowered()
 
 			// Populate resources and get invokes.
-			items := versionMap[moduleVersion]
+			items := moduleVersions[moduleVersion]
 			var resources []string
 			for resource := range items.Resources {
 				resources = append(resources, resource)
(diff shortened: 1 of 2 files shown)


provider/pkg/gen/schema.go
```diff
@@ -275,47 +276,38 @@ func PulumiSchema(rootDir string, providerMap openapi.AzureProviders, versioning
...
 
-	for _, providerName := range providers {
-		versionMap := providerMap[providerName]
-		var versions []openapi.SdkVersion
-		for version := range versionMap {
-			versions = append(versions, version)
-		}
-		slices.Sort(versions)
+	for _, moduleName := range util.SortedKeys(modules) {
+		versionMap := modules[moduleName]
```
thomas11, line 287: While we're renaming things, `versionMap` is pretty bad, too. `ResourcesByApiVersion`?

provider/pkg/gen/schema.go
```diff
@@ -1293,31 +1292,33 @@ func (g *packageGenerator) genFunctions(typeName, path string, specParams []spec
 	g.metadata.Invokes[functionTok] = f
 }
 
-// moduleName produces the module name from the provider name and the version e.g. `compute/v20200701`.
-func (g *packageGenerator) moduleName() string {
-	return providerApiToModule(g.provider, g.sdkVersion)
+type TokenModule string
```
thomas11, line 1295: Comment with examples. And wouldn't `ModuleToken` be more fitting?

provider/pkg/resources/resources.go
```diff
@@ -314,12 +351,46 @@ func getNameOverride(majorVersion uint64, specFolderName, namespaceWithoutPrefix
...
 		if folderName, ok := namespaceOverrides[specFolderName]; ok {
-			return folderName, &namespaceWithoutPrefix, true
+			oldModule := ModuleName(namespaceWithoutPrefix)
+			return ModuleName(folderName), &oldModule, true
 		}
 	}
 	return "", nil, false
 }
 
+// Identify the first segment of the path that contains a namespace.
```
thomas11, line 361: Isn't taking the first namespace a change in behavior? Previously, the code used `findNamespaceWithoutPrefixFromPath` which takes the last.

provider/pkg/versioning/gen.go
```diff
@@ -116,41 +116,41 @@ func calculateVersionMetadata(versionSources VersionSources) (VersionMetadata, e
...
 	}, nil
 }
 
 func (v VersionMetadata) WriteTo(outputDir string) ([]string, error) {
-	filePrefix := fmt.Sprintf("v%d-", v.MajorVersion)
-	specPath := filePrefix + "spec.yaml"
-	lockPath := filePrefix + "lock.json"
+	filePrefix := fmt.Sprintf("v%d", v.MajorVersion)
+	specPath := filePrefix + "-spec.yaml"
+	lockPath := filePrefix + ".yaml"
```
thomas11, line 128: I suggest including what it is in the file name, like `v3-default-versions.yaml`.

thomas11 (commented):
LGTM except maybe the "namespace from path" logic change?

provider/pkg/gen/schema.go
danielrbradley: Yup, will add a comment 👍 

This is the module _from_ the token (`[provider]:[module]:[resource]`) rather than a token representing a module .. hence I went the other way round. ModuleToken certainly sounds more natural though 🤔

thomas11 (approved):

PR #3881, opened 2025-01-23
Add information for cross-major-version aliases

danielrbradley:
- Track where each definition (resource or invoke) has come from (source path).
- Track the resource URI in the default versions file so we can compare between major versions where a rename has happened and add an alias.
- Track the Azure resource provider namespace and use it when checking if a RP version is active rather than using the module name which might not align to the RP name (and might not be prefixed `Microsoft.`).

### DIFF
--- a/provider/pkg/versionLookup/query.go
+++ b/provider/pkg/versionLookup/query.go
@@ -12,7 +12,9 @@ var rawVersionLock []byte
 // versionLock is a map from Azure provider name to resource name to API version.
 // This is actually the openapi.DefaultVersionLock type but we don't want to import openapi here
 // to avoid cycles.
-var versionLock map[string]map[string]struct{ ApiVersion string }
+var versionLock map[string]map[string]struct {
+	ApiVersion string `yaml:"ApiVersion"`
+}
 
 func init() {
 	err := yaml.Unmarshal(rawVersionLock, &versionLock)
--- a/provider/pkg/versionLookup/query_test.go
+++ b/provider/pkg/versionLookup/query_test.go
@@ -1,0 +1,17 @@
+package versionLookup_test
+
+import (
+	"testing"
+
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/openapi"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/versionLookup"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLookup(t *testing.T) {
+	result, ok := versionLookup.GetDefaultApiVersionForResource("Web", "WebApp")
+	assert.True(t, ok, "expected to find default API version for Web/WebApp")
+	// Verify we can convert it back to a date to prove it's in a valid format
+	_, err := openapi.ApiVersionToDate(openapi.ApiVersion(result))
+	assert.NoError(t, err, "expected to parse API version as date")
+}


provider/pkg/openapi/discover.go
```diff
@@ -191,13 +191,31 @@ func NewVersionResources() VersionResources {
 
 type ModuleVersionList = map[ModuleName][]ApiVersion
 
+// Can only by "Resource" or "Invoke"
```
thomas11, line 194: [nit] "be"

provider/pkg/openapi/discover.go
```diff
@@ -191,13 +191,31 @@ func NewVersionResources() VersionResources {
...
+type DefinitionType string
```
thomas11, line 195: I'm not loving the name "definition" yet because the spec also has type definitions but these are not represented here. Don't have a better idea ATM though.

provider/pkg/providerlist/providerlist.go
```diff
@@ -68,17 +68,17 @@ func ReadProviderList(providerListPath string) (*ProviderList, error) {
...
 
 type providerListIndex struct {
 	providerResourceVersionLookup ProviderPathVersions
 	providerVersions              map[LoweredProviderName]ApiVersions
 }
 
-func (index *providerListIndex) HasProviderVersion(providerName string, version openapi.ApiVersion) bool {
-	providerName = strings.ToLower(providerName)
-	versions, ok := index.providerVersions[providerName]
+func (index *providerListIndex) HasProviderVersion(namespace string, version openapi.ApiVersion) bool {
```
thomas11, line 79: Didn't we start to standardize on the term "module" in your previous PR?

provider/pkg/versioning/defaultVersion.go
```diff
@@ -150,16 +149,30 @@ func DefaultVersionsFromConfig(spec ModuleVersionResources, defaultConfig Spec)
 		}
 
 		if moduleConfig.Tracking != nil {
-			for _, resourceName := range versionResources[*moduleConfig.Tracking] {
-				definitions[resourceName] = *moduleConfig.Tracking
+			trackingDefinitions := versionResources[*moduleConfig.Tracking]
+			for _, definitionName := range util.SortedKeys(trackingDefinitions) {
+				definition := trackingDefinitions[definitionName]
+				definitions[definitionName] = openapi.DefinitionVersion{
+					ApiVersion:   definition.ApiVersion,
```
thomas11, line 156: Why are we essentially making a copy of `definition` here? Package boundary?

provider/pkg/versioning/defaultVersion.go
```diff
@@ -397,19 +411,17 @@ func (b moduleSpecBuilder) containsRecentStable(orderedVersions []openapi.ApiVer
 
 // findMinimalVersionSet returns the minimum set of versions required to produce all resources
 func findMinimalVersionSet(versions VersionResources) []openapi.ApiVersion {
-	orderedVersions := make([]openapi.ApiVersion, 0, len(versions))
-	for version := range versions {
-		orderedVersions = append(orderedVersions, version)
-	}
-	slices.Sort(orderedVersions)
+	// TODO: Consider using openapi.SortApiVersions
```
thomas11, line 414: What's the plan for the two TODOs in this function?

thomas11 (commented):
Looks like a great improvement. Especially v2.yaml!

Only suggestion in addition to the comments: there are a lot of new map accesses without an `, ok` guard. Also one `slice[0]` without length check. Even if the existence of these elements is an invariant, I'd feel better about checking and asserting the invariant instead of failing in hard-to-debug ways later.

provider/pkg/versioning/defaultVersion.go
danielrbradley: Oh good catch .. wanted to do this once I'd fixed the default version regression.

provider/pkg/providerlist/providerlist.go
danielrbradley: Yes, for within the schema it's a module, but this really is an Azure Provider identified by a full namespace.

provider/pkg/versioning/defaultVersion.go
danielrbradley: Ah .. I added two fields (Type and Name) which I didn't want cluttering the output files. Looking again, we can just remove these fields as they didn't end up getting used.

provider/pkg/openapi/discover.go
danielrbradley: Endpoint? URI? Maybe it's just a Resource, but invokes are like read-only resources?

provider/pkg/versioning/defaultVersion.go
danielrbradley: Oddly, removing these fields causes a diff in the `v2.yaml` 🙃 More investigation needed!

danielrbradley: Ok, found another pre-existing bug .. we didn't validate that the resource versions in the `additions` part of the default version spec actually exist. Added extra validation of this - again caught out by not checking that map retrivals were successful.

provider/pkg/util/data_structures.go
```diff
@@ -31,6 +31,14 @@ func GetInnerMap(m map[string]any, keys ...string) (map[string]any, bool) {
 	return nil, false
 }
 
+func UnsortedKeys[K comparable, V any](m map[K]V) []K {
```
thomas11, line 34: `SortedKeys` could now be `UnsortedKeys` + sort

thomas11 (approved):
I'd still recommend some checks when accessing maps but otherwise LGTM

PR #3888, opened 2025-01-24
WIP: Auto-alias renames between major versions

danielrbradley:


danielrbradley:
[The first diff](https://github.com/pulumi/pulumi-azure-native/compare/2b4f45e4d986770c821e60f64b391cad2cff10ce..7cd9fdf34ee4e933e994d86c32a58d171793768f) shows a larger set of changes than expected. Perhaps we're not accounting for later adjustments to the namings. 

Also, it appears like we're loosing some of the docs around previous default resource version.

danielrbradley:
Replaced with https://github.com/pulumi/pulumi-azure-native/pull/3906

PR #3894, opened 2025-01-27
Remove type merging overrides we don't need anymore

### LINKED ISSUE #2749: Remove type merging overrides
As new resources are added from API versions that are newer than our current selection in the default version, we add only specific resources. However, this means that we need to merge the type definitons across the versions. Some types are not mergable and so we need to manually modify them to allow them to be merged. This is see here: 

https://github.com/pulumi/pulumi-azure-native/blob/b3c267f37bca06915eaaf057e3d7ab864dcf55c7/provider/pkg/gen/types.go#L101-L118

When doing the next major version upgrade, we should select simplified sets of resources for the default version and so we should no longer need any of these manual merge overrides.

This was created as a follow-on issue from:
- https://github.com/pulumi/pulumi-azure-native/pull/2747
- https://github.com/pulumi/pulumi-azure-native/pull/2924

thomas11:
Resolves #2749 but not completely yet.

So far, these are the overrides that can be removed without any further changes, and `make schema` succeeds.

### DIFF
--- a/sdk/dotnet/ScVmm/Inputs/GuestCredentialArgs.cs
+++ b/sdk/dotnet/ScVmm/Inputs/GuestCredentialArgs.cs
@@ -18,14 +18,14 @@ namespace Pulumi.AzureNative.ScVmm.Inputs
         /// <summary>
         /// Gets or sets the password to connect with the guest.
         /// </summary>
-        [Input("password")]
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
Looking good 👍

Issue #3422, opened 2024-07-08
V3: shrink package size


thomas11:
This is an open-ended issue to capture ideas and investigations on how we could shrink the size the provider and its SDK. The v3 major version bump might provide opportunities for breaking changes.

olafurnielsen:
Size in megabytes isn't the most concerning problem for this provider IMO – rather the number of symbols/objects the IDE's language server has to analyze & index, for properly typed experience.

As mentioned in #3124 the sheer number of objects in this provider, due to the breadth of the Azure API and autogenerated nature of this provider, causes the Python Language Server to run out of memory and blow up.

Now with the introduction of #3400 number of symbols in the provider seems to have increased by roughly 30k:
```shell
$ grep -Ri 'ArgsDict(TypedDict)' . | wc -l
29958
```

I really like the direction of TypedDicts and assume they will supersede the Args classes in the next major version but this seems ot have made matters even worse than before, causing more frequent crashes of the Python Language Server in VScode.

AaronFriel:
Is LSP performance improved or affected by how pulumi_azure_native is imported?

```py
# import named types:
from pulumi_azure_native.storage import ...
# import submodule
from pulumi_azure import storage
# import package
import pulumi_azure
```

olafurnielsen:
Don't know if this is getting out of scope but I guess these are in the end symptoms caused by the size/breadth of the SDK – @thomas11 feel free to comment if you want Python/IDE/typing specific issues/symptoms to be tracked in a sub/seperate issue.

> Is LSP performance improved or affected by how pulumi_azure_native is imported?

For sure, `import pulumi_azure_native` will trigger the LSP into scanning the whole library and should be avoided. I've made several attempts in optimizing the settings and import strategies but importing a sub-package (`... import storage`) vs. specific objects `from [...].storage import X, Y Z` doesn't seem to make much of a difference – I suspect the LSP will analyze `__init__.py` anyways and follow the trail to other modules in the same sub-package.

I took another shot at investigating the behaviour in a freshly created azure-native Python project and even when being extra careful with your imports something like hovering over a module and navigating to its definition will kick off a library-wide scan.

After some unstructured browsing through issues of similar nature tracked in Pyright & MyPy affecting other large Python libraries I wanted to throw some ideas into the mix that _might_ (?) improve IDE experience in generated Python SDKs:

[...]

PR #3899, opened 2025-01-28
A new report listing API versions that are old enough to be removed

thomas11:
Part of #3422. Determine API versions that are older than any _previous_ default version in a module. These are good candidates for removal.

### DIFF
--- a/reports/README.md
+++ b/reports/README.md
@@ -48,3 +48,7 @@ This, most likely very unelegant, `jq` query filters the endpoints to only those
 ```jq
 jq '[ keys[] as $rp | .[$rp] | keys[] as $path | .[$path][] ] | map(select(.HttpVerbs != null) | select(.HttpVerbs | contains(["POST"])))' allEndpoints.json
 ```
+
+## `oldApiVersions.json`
+
+This file contains a list of all API versions that are older than the oldest previous default version for each resource provider. These API versions are good candidates for removal. Versions that are already explicitly removed via `vN-removed.json` are not included.


provider/pkg/versioning/gen.go
```diff
@@ -85,6 +86,86 @@ func (v VersionMetadata) GetAllVersions(moduleName openapi.ModuleName, resource
...
+		for _, version := range v.RemovedVersions[moduleName] {
+			removedFromModule[version] = struct{}{}
+		}
+
+		// for resourceName := range resourceVersions {
+		// 	previousDefault := v.PreviousDefaultVersions[moduleName][resourceName].ApiVersion
+		// 	mustKeep[previousDefault] = struct{}{}
+		// 	currentDefault := v.DefaultVersions[moduleName][resourceName].ApiVersion
+		// 	mustKeep[currentDefault] = struct{}{}
+		// }
```
danielrbradley, line 105: Let's clean this up if we don't need it

provider/pkg/versioning/gen.go
```diff
@@ -85,6 +86,86 @@ func (v VersionMetadata) GetAllVersions(moduleName openapi.ModuleName, resource
...
+					oldVersionsSet[version] = struct{}{}
+				}
+			}
+			oldVersions := []openapi.ApiVersion{}
+			for version := range oldVersionsSet {
+				oldVersions = append(oldVersions, version)
+			}
+			slices.SortStableFunc(oldVersions, openapi.CompareApiVersions)
+			if len(oldVersions) > 0 {
+				// result[moduleName] = append(result[moduleName], oldVersions)
```
danielrbradley, line 121: Why is this commented?

danielrbradley (commented):
Overall approach looks sensible 👍 

The main algorithm could do with a couple of comments helping explain the rational of each major step.

Also, looks like there's a diff in the generated file in CI that we'll need to resolve.

provider/pkg/versioning/gen.go
thomas11: I'm not seeing this locally??? Investigating...

provider/pkg/versioning/gen.go
```diff
@@ -85,6 +86,53 @@ func (v VersionMetadata) GetAllVersions(moduleName openapi.ModuleName, resource
...
+			}
+		}
+
+		// If there are any old versions, sort them and add them to the result.
+		if len(oldVersionsSet) > 0 {
+			oldVersions := []openapi.ApiVersion{}
+			for version := range oldVersionsSet {
+				oldVersions = append(oldVersions, version)
+			}
+			slices.SortStableFunc(oldVersions, openapi.CompareApiVersions)
```
danielrbradley, line 129: Nit: could simplify this to:

```suggestion
			oldVersions := util.UnsortedKeys(oldVersionsSet)
			slices.SortStableFunc(oldVersions, openapi.CompareApiVersions)
```

danielrbradley (approved):
Looks good, minor nit inline, also could add a line to the `reports/README.md` to explain the purpose of the new report.

provider/pkg/versioning/gen.go
danielrbradley: Oh, or even `maps.Keys()`! TIL

PR #3902, opened 2025-01-28
Sort (and deduplicate) type aliases

danielrbradley:
- Use a set to ensure we don't add duplicate aliases.
- Render the aliases ordered.

This just affects explicit versions, moving the default version from the first alias to the last in the list.

This will allow us to add additional sources of aliases (from the previous major version) without causing duplicates.

### DIFF
--- a/provider/pkg/gen/schema.go
+++ b/provider/pkg/gen/schema.go
@@ -1075,8 +1075,7 @@ func (g *packageGenerator) generateAliases(resource *resourceVariant, typeNameAl
 
 	var aliasSpecs []pschema.AliasSpec
 	for _, v := range typeAliases.SortedValues() {
-		typeToken := v
-		aliasSpecs = append(aliasSpecs, pschema.AliasSpec{Type: &typeToken})
+		aliasSpecs = append(aliasSpecs, pschema.AliasSpec{Type: &v})
 	}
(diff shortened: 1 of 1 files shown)


provider/pkg/gen/schema.go
```diff
@@ -1037,47 +1038,47 @@ func isSingleton(resource *resourceVariant) bool {
...
-				addAlias(*resource.ModuleNaming.PreviousName, alias, version.ToSdkVersion())
+				typeAliases.Add(generateTok(*resource.ModuleNaming.PreviousName, alias, version.ToSdkVersion()))
 			}
 		}
 	}
 
-	return aliases
+	var aliasSpecs []pschema.AliasSpec
+	for _, v := range typeAliases.SortedValues() {
+		typeToken := v
```
thomas11, line 1078: do we still need this assignment with Go 1.22+?

thomas11 (approved):

provider/pkg/gen/schema.go
danielrbradley: ¯\_(ツ)_/¯ hmm, maybe not

danielrbradley: Fixed

PR #3906, opened 2025-01-29
Add aliases from compatible paths

danielrbradley:
1. Generate a report after generation with just the resource paths for each token.
2. Manually copy this into the version folder to be used as a source for the next version.
3. Create a lookup which finds "compatible resources" based on the path being the same or the module/resource names matching (case insensitively).
4. Add aliases for the compatible resources.


provider/pkg/gen/compatibleTokens.go
```diff
@@ -0,0 +1,84 @@
...
+func NewCompatibleTokensLookup(tokenPaths map[string]string) (*CompatibleTokensLookup, error) {
+	lookup := CompatibleTokensLookup{
+		byLoweredModuleResourceName: map[ModuleLowered]map[ResourceLowered]map[Token]struct{}{},
+		byNormalizedPath:            map[NormalizedPath]map[Token]struct{}{},
+	}
+	if tokenPaths == nil {
+		tokenPaths = map[string]string{}
+	}
+	for token, path := range util.MapOrdered(tokenPaths) {
+		tokenParts := strings.Split(token, ":")
```
thomas11, line 32: We have `resources.ParseToken`

provider/pkg/gen/compatibleTokens.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// NewCompatibleTokensLookup creates a new CompatibleResourceLookup from the given Azure modules and default versions.
+func NewCompatibleTokensLookup(tokenPaths map[string]string) (*CompatibleTokensLookup, error) {
+	lookup := CompatibleTokensLookup{
+		byLoweredModuleResourceName: map[ModuleLowered]map[ResourceLowered]map[Token]struct{}{},
+		byNormalizedPath:            map[NormalizedPath]map[Token]struct{}{},
+	}
+	if tokenPaths == nil {
+		tokenPaths = map[string]string{}
```
thomas11, line 29: We're not writing to this map, are we? So nil should be ok. If `util.MapOrdered` can't handle nil, it should be fixed, since nil maps are normal in Go.

provider/pkg/gen/compatibleTokens.go
```diff
@@ -0,0 +1,84 @@
...
+	lookup.byLoweredModuleResourceName[moduleLowered][resourceLowered][typedToken] = struct{}{}
+	if _, ok := lookup.byNormalizedPath[normalizedPath]; !ok {
+		lookup.byNormalizedPath[normalizedPath] = map[Token]struct{}{}
+	}
+	lookup.byNormalizedPath[normalizedPath][typedToken] = struct{}{}
+}
+
+func (lookup *CompatibleTokensLookup) FindCompatibleTokens(moduleName openapi.ModuleName, resourceName openapi.ResourceName, path string) []string {
+	if lookup == nil || lookup.byLoweredModuleResourceName == nil || lookup.byNormalizedPath == nil {
+		return nil
```
thomas11, line 65: Silently returning an empty result could lead to surprising behavior. Return `([]string, error)`?

provider/pkg/gen/compatibleTokens.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	matches := map[string]struct{}{}
```
thomas11, line 67: `collections.OrderableSet`? Since it's sorted at the end of the func.

thomas11: Could actually return an OrderableSet so the caller wouldn't need to create one from the slice.

provider/pkg/gen/compatibleTokens.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/openapi/paths"
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/util"
+)
+
+type Token string
+type ModuleLowered string
+type ResourceLowered string
+type NormalizedPath string
+
+type CompatibleTokensLookup struct {
```
thomas11, line 17: godoc

provider/pkg/gen/compatibleTokens.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	lookup.byLoweredModuleResourceName[moduleLowered][resourceLowered][typedToken] = struct{}{}
+	if _, ok := lookup.byNormalizedPath[normalizedPath]; !ok {
+		lookup.byNormalizedPath[normalizedPath] = map[Token]struct{}{}
+	}
+	lookup.byNormalizedPath[normalizedPath][typedToken] = struct{}{}
+}
+
+func (lookup *CompatibleTokensLookup) FindCompatibleTokens(moduleName openapi.ModuleName, resourceName openapi.ResourceName, path string) []string {
```
thomas11, line 63: godoc

provider/pkg/gen/compatibleTokens.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil, fmt.Errorf("malformed token '%s'", token)
+		}
+		tokenModule, resourceName := tokenParts[1], tokenParts[2]
+		moduleName, _, _ := strings.Cut(tokenModule, "/")
+		lookup.add(openapi.ModuleName(moduleName), resourceName, path, token)
+	}
+	return &lookup, nil
+}
+
+func (lookup *CompatibleTokensLookup) add(moduleName openapi.ModuleName, resourceName openapi.ResourceName, path string, token string) {
```
thomas11, line 43: I searched but can't see where this is called.

danielrbradley:
Observations from merging with the v3 branch:
- Quite a few are in the explicit versions which never previously got an alias back to the default version.
- There's a few resources where two resources seem to share the same path and so it's adding in a bunch of extra aliases:

  - cache:EnterpriseDatabase links back to cache:Database
  - apimanagement:GlobalSchema <--> Schema
  - costmanagement:CloudConnector <--> Connector
  - dbformysql:SingleServer links back to Server for versions v20171201 and v20180601privatepreview only (and similar for the related FirewallRule and Database)
  - Similar for dbforpostgresql to mysql above
  - databox:*Role seem to share paths (CloudEdgeManagementRole, IoTRole, KubernetesRole, MECRole)
  - Similar for datashare:*DataSet (Blob... KustoCluster...,  etc)

thomas11:
> Observations from merging with the v3 branch:
> 
> * Quite a few are in the explicit versions which never previously got an alias back to the default version.
> * There's a few resources where two resources seem to share the same path and so it's adding in a bunch of extra aliases:
>   
>   * cache:EnterpriseDatabase links back to cache:Database
>   * apimanagement:GlobalSchema <--> Schema
>   * costmanagement:CloudConnector <--> Connector
>   * dbformysql:SingleServer links back to Server for versions v20171201 and v20180601privatepreview only (and similar for the related FirewallRule and Database)
>   * Similar for dbforpostgresql to mysql above
>   * databox:*Role seem to share paths (CloudEdgeManagementRole, IoTRole, KubernetesRole, MECRole)
>   * Similar for datashare:*DataSet (Blob... KustoCluster...,  etc)

That's a bit concerning. Several of those were explicitly addressed in #3817, e.g.
```go
if strings.Contains(canonPath, "/redisenterprise/") {
 	result = prefix("RedisEnterprise")
 }
```

provider/pkg/gen/compatibleTokens.go
danielrbradley: 5 lines above - it's just internally from NewCompatibleTokensLookup

provider/pkg/gen/compatibleTokens.go
danielrbradley: I _think_ that's true, but actually, we can just do this instead:

```suggestion
		return &lookup, nil
```

provider/pkg/gen/compatibleTokens.go
```diff
@@ -0,0 +1,84 @@
...
+type ModuleLowered string
+type ResourceLowered string
+type NormalizedPath string
+
+type CompatibleTokensLookup struct {
+	byLoweredModuleResourceName map[ModuleLowered]map[ResourceLowered]map[Token]struct{}
+	byNormalizedPath            map[NormalizedPath]map[Token]struct{}
+}
+
+// NewCompatibleTokensLookup creates a new CompatibleResourceLookup from the given Azure modules and default versions.
```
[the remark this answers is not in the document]
danielrbradley: Oop, missed in the cherry-pick from my old branch.

provider/pkg/gen/compatibleTokens.go
danielrbradley: Yeah, I did have a separate method `IsPopulated` to assert there's something in the collection, but then we'd still just carry on for now even if there's nothing in here because it's not populated for v2.

Perhaps I add that check back in and fail hard if it's not populated and the major version is v3?

provider/pkg/gen/compatibleTokens.go
danielrbradley: Yup, nice the Set is much neater. No need to return the Set as we're just looping over to ensure they're added to another set.

danielrbradley:
Ok, I think this is nearly ready to go.

I'll remove the code added for the ModuleNaming.PreviousName as that's now unused and update the unit test suite to better describe the desired behaviour.

danielrbradley:
Ok, I think this is ready to go - unit tests are fixed.

In a follow-up PR, I'll remove the code added for the ModuleNaming.PreviousName as that's now unused and update the unit test suite to better describe the desired behaviour.

An initial test showed that removing the module naming approach only removes a handful of aliases in v3 where the explicit version wasn't actually present in v2 due to the squeeze process.

Additionally, after v2 we can completely remove the `typeName` aliases as they have zero impact on the v3 schema (but they are still used in v2 so it will have to go into the v3 branch).

PR #3907, opened 2025-01-29
Include API Version in input/state

### LINKED ISSUE #3766: Include API version as output
When looking at Azure Native resources in state it's difficult to work out which API version they were actually created with. This is additionally much harder when changing major version as the API version will likely change, but without referring to the documentation it's diffucult to know what API version the old state was using.

Therefore, we should add a new hidden output field to all resource outputs to capture this information. This would be useful for:
- Users looking at their state to indentify which API version was being used for a resource for debugging issues with Azure's APIs.
- When applying migrations within the provider, to know what the previous version of the resource was.
- Reporting within insights to see the version in the console for the resource.

We already have the `type` output which is included in all resources, [added via the common types](https://github.com/Azure/azure-rest-api-specs/blob/2bde125befabb21807a2021765901f20e3e74ec8/specification/common-types/resource-management/v5/types.json#L25-L29). However, no "apiVersion" field is included by the API, so we could add this ourselves, if it doesn't conflict with any API-returned fields.

Example resource state:

```diff
      {
        "urn": "urn:pulumi:test::upgrade-networked-vm::azure-native:resources:ResourceGroup::rg",
        "custom": true,
        "id": "/subscriptions/<redacted>/resourceGroups/rg8baa4592",
        "type": "azure-native:resources:ResourceGroup",
        "inputs": {
          "__createBeforeDelete": true,
          "location": "westus2",
          "resourceGroupName": "rg8baa4592"
        },
        "outputs": {
          "__inputs": {
            "4dabf18193072939515e22adb298388d": "1b47061264138c4ac30d75fd1eb44270",
            "plaintext": "\"REDACTED BY PROVIDERTEST\""
          },
+         "apiVersion": "2022-09-01"
          "id": "/subscriptions/<redacted>/resourceGroups/rg8baa4592",
          "location": "westus2",
          "name": "rg8baa4592",
          "properties": {
            "provisioningState": "Succeeded"
          },
          "type": "Microsoft.Resources/resourceGroups",
        },
```

EronWright:
## Overview

Closes #3766 

This PR is aimed at improving the experience around api versions, e.g. when upgrading a resource to a new api version. Note that major version upgrades of this provider tend to change the 'default' versions of many resources.
1. Show some diff output and and actually update the resource state when the api version of a resource is being changed.
2. Record the actual api version that was used in create/update/read, to inform subsequent operations.

In 2.x mode, the provider simply records the actual `azureApiVersion` as an output property, but doesn't expose it at the SDK level. The rationale is to have the information available in "old state" +before+ the major upgrade is performed; this allows the provider to avoid a spurious diff and to make fewer update calls during the major upgrade.

image

In 3.x mode, the provider introduces a new `azureApiVersion` input property, with a constant value. The input property is NOT added to the resource args; it is always defaulted in `Check`. In the future, a true input property might be added to support overrides.  The purpose of having a checked input is to produce a diff when the API version has changed (for whatever reason).  This leads to an update call against the new API version of the resource, to validate the inputs and refresh the state.  

The output property that was added to state in 2.x is fully exposed in 3.x at the SDK level.

### Migration
In a migration case, the old `azureApiVersion` isn't yet known, and this may lead to a spurious (though harmless) update because the value goes from empty to known.  We mitigate this by reading `azureApiVersion` from old state during Diff.  The recommended migration flow is:
1. User upgrades to the latest 2.x with `pulumi up` (to update the provider version) followed by `pulumi refresh` (to update the resource's state.  Now there's an `azureApiVersion` in state for each 2.x resource.
2. User upgrades to 3.0 with `pulumi up`, and the provider compares old state to new input to see whether the api version is actually changing, e.g. due to a change to the "default version" of a resource.

For example, here's the impact to the state file after completing the migration as described.  We see two affected resources: a resource group and a storage account, with the latter undergoing a change to the 'default' API version.

image

image

### Invokes
Another aspect to this PR is the `azureApiVersion` output property and supporting the `getFoo` functions.  The code generator must be careful to apply the output property to the resource type (as represented by `isTopLevel`), not to functions. There is, however, a special "get" function where it does make sense to return the `azureApiVersion`.

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -727,7 +727,7 @@ func (k *azureNativeProvider) validateProperty(ctx string, prop *resources.Azure
 
 		// Typed object: validate all properties by looking up its type definition.
 		typeName := strings.TrimPrefix(prop.Ref, "#/types/")
-		typ, ok, err := k.lookupTypeDefault(prop.Ref)
+		typ, ok, err := k.lookupType(prop.Ref)
 		if err != nil {
 			return append(failures, &rpc.CheckFailure{
 				Reason: fmt.Sprintf("error decoding type spec '%s': %v", typeName, err),
@@ -892,7 +892,7 @@ func (k *azureNativeProvider) Diff(_ context.Context, req *rpc.DiffRequest) (*rp
 		return nil, err
 	}
 
-	detailedDiff := diff(k.lookupTypeDefault, *res, oldInputs, newResInputs)
+	detailedDiff := diff(k.lookupType, *res, oldInputs, newResInputs)
 	if detailedDiff == nil {
 		return &rpc.DiffResponse{
 			Changes:             rpc.DiffResponse_DIFF_NONE,


EronWright:
Replaced by https://github.com/pulumi/pulumi-azure-native/pull/4015

PR #3912, opened 2025-01-30
Manually pin Container instance for v3

danielrbradley:
- Only ContainerGroup and ContainerGroupProfile are actually listed as live.
- The latest live version with both resources is 2024-05-01-preview so we'll track that for v3.
- Exclude the NGroup resource as it's not live and causes conflicts for now.

This lets us remove the manual type merging override.

### DIFF
--- a/provider/pkg/gen/schema.go
+++ b/provider/pkg/gen/schema.go
@@ -1062,36 +1062,17 @@ func (g *packageGenerator) generateAliases(resourceTok string, resource *resourc
 		typeAliases.Add(token)
 	}
 
-	// Add an alias for the same version of the resource, but in its old module.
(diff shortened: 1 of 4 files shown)


thomas11 (approved):
2024-05 is also [what the latest beta of the Go SDK uses](https://github.com/Azure/azure-sdk-for-go/blob/sdk/resourcemanager/containerinstance/armcontainerinstance/v2.5.0-beta.1/sdk/resourcemanager/containerinstance/armcontainerinstance/autorest.md) ([via](https://azure.github.io/azure-sdk/releases/latest/mgmt/go.html)). Latest stable is still on 2023-05, though.

PR #3924, opened 2025-02-04
Don't include the default version as an explicit version in v3

danielrbradley:
Due to low confidence on the migration path in v2, we added all resource being used in the default version as explicit versions too. 

With the improved aliasing approach from #3872, this is not required any more.

See the preview generation against the v3 branch: https://github.com/pulumi/pulumi-azure-native/compare/v3...dont-duplicate-default-as-explicit-v3-preview

This has a net reduction of 1,118,866 lines of SDK code and an uncompressed size reduction of just under 7% to deliver a 48% overall reduction for v3.

| SDK | v2 size | v3 size | new v3 size |
|-----|-------|------|---------|
| dotnet | 653M | 363M | 340M |
| nodejs | 300M | 166M | 156M |
| python | 717M | 401M | 376M |

### DIFF
--- a/provider/pkg/versioning/gen.go
+++ b/provider/pkg/versioning/gen.go
@@ -41,8 +41,8 @@ func (v VersionMetadata) ShouldInclude(moduleName openapi.ModuleName, version *o
 	if version == nil {
 		return true
 	}
-	// Keep any resources in the default version lock
-	if v.DefaultVersions.IsAtVersion(moduleName, typeName, *version) {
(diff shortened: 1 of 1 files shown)


thomas11 (approved):
nice find!!

EronWright (commented):
A key difference between versioned and unversioned resources is that the version is pinned across major upgrades.  Does this PR drop some of the versioned resources that would allow me to pin?

PR #3959, opened 2025-02-18
Support for PIM Role Eligibility Schedules

### LINKED ISSUE #2455: Add AzureRM Privileged Identity Management to Provider

thomas11:
Resolves #2455 - add support for PIM (Privileged Identity Management) Role Eligibility Schedules.

- [Azure docs](https://learn.microsoft.com/en-us/rest/api/authorization/privileged-role-eligibility-rest-sample)
- [Spec](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/authorization/resource-manager/Microsoft.Authorization/stable/2020-10-01)

This API has some peculiarities that required writing a custom resource:
1. It's divided into schedules and schedule requests (an endpoint `/roleEligibilityScheduleRequests`, not HTTP requests). Requests are created to create and also delete schedules; there's no DELETE endpoint. Requests are short-lived but can require approval.
2. Because of this duality where two types are used, schedule and schedule request, I extended custom resources with the ability to derive their schema from type A in the spec but define a different resource name B. We want the resource to be named `...Schedule` but need the `...ScheduleRequest` types.
3. While a schedule request is pending, a `/cancel` endpoint is used to delete the request.
4. Both schedules and schedule requests cannot be modified; new ones must be created instead. This isn't properly annotated in the spec, though. The SDKs are correct in that they don't offer update operations.

I tried to comment the code pretty well, please refer to the Godoc comments for more in-depth information.

### DIFF
--- a/provider/pkg/resources/customresources/custom_pim_eligibility_test.go
+++ b/provider/pkg/resources/customresources/custom_pim_eligibility_test.go
@@ -70,6 +70,8 @@ type fakePimEligibilityScheduleClient struct {
 		name       string
 		parameters armauthorization.RoleEligibilityScheduleRequest
 	}
+
+	mapScheduleToOutputsCalls int
 }
 
 func (c *fakePimEligibilityScheduleClient) findSchedule(ctx context.Context, scope, principalID, roleDefinitionID string) (*armauthorization.RoleEligibilitySchedule, error) {
@@ -110,7 +112,10 @@ func (c *fakePimEligibilityScheduleClient) cancelSchedule(ctx context.Context, s
 }
 
 func (c *fakePimEligibilityScheduleClient) mapScheduleToOutputs(schedule *armauthorization.RoleEligibilitySchedule) (map[string]any, error) {
-	return map[string]any{}, nil
+	c.mapScheduleToOutputsCalls++
+	return map[string]any{
+		"name": *schedule.Name,
+	}, nil
 }
 
 func TestCreatePimEligibilitySchedule(t *testing.T) {
@@ -397,5 +402,66 @@ func TestInputsToSdk(t *testing.T) {
 		require.NotNil(t, sdk.Properties.TicketInfo)
 		assert.Equal(t, "123456", *sdk.Properties.TicketInfo.TicketNumber)
 		assert.Equal(t, "pulumi", *sdk.Properties.TicketInfo.TicketSystem)
+	})
+}
+
+func TestReadPimEligibilitySchedule(t *testing.T) {
+	t.Parallel()
+
+	guid := uuid.New().String()
+	validRequiredInputs := resource.PropertyMap{
+		"name":             resource.NewStringProperty(guid),
+		"scope":            resource.NewStringProperty("scope"),
+		"principalId":      resource.NewStringProperty("principalID"),
+		"roleDefinitionId": resource.NewStringProperty("roleDefinitionID"),
+	}
+
+	existingSchedule := &armauthorization.RoleEligibilitySchedule{
+		Name: pulumi.StringRef(guid),
+		Properties: &armauthorization.RoleEligibilityScheduleProperties{
+			PrincipalID:      pulumi.StringRef("principalID"),
+			RoleDefinitionID: pulumi.StringRef("roleDefinitionID"),
+			Scope:            pulumi.StringRef("scope"),
+		},
+	}
+
+	t.Run("found", func(t *testing.T) {
+		t.Parallel()
+
+		client := &fakePimEligibilityScheduleClient{
+			scheduleToFind: existingSchedule,
+		}
+
+		result, found, err := read(context.Background(), "/id/is/not/used", validRequiredInputs, client)
+		require.NoError(t, err)
+		assert.True(t, found)
+		assert.Equal(t, guid, result["name"])
+	})
+
+	t.Run("not found", func(t *testing.T) {
+		t.Parallel()
+
+		client := &fakePimEligibilityScheduleClient{}
+
+		inputs := validRequiredInputs.Copy()
+		result, found, err := read(context.Background(), "/id/is/not/used", inputs, client)
+		require.NoError(t, err)
+		assert.False(t, found)
+		assert.Nil(t, result)
+	})
+
+	t.Run("error converting inputs", func(t *testing.T) {
+		t.Parallel()
+
+		client := &fakePimEligibilityScheduleClient{}
+
+		inputs := validRequiredInputs.Copy()
+		delete(inputs, "scope")
+
+		result, found, err := read(context.Background(), "/id/is/not/used", inputs, client)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "scope is required")
+		assert.False(t, found)
+		assert.Nil(t, result)
 	})
 }


provider/pkg/openapi/discover.go
```diff
@@ -563,6 +563,17 @@ func (modules AzureModules) addAPIPath(specsDir, fileLocation, path string, swag
 	return addResourcesAndInvokes(version, fileLocation, path, moduleNaming, swagger)
 }
 
+// getTypeName returns the type name for a given operation and path. The path is used to check for custom resource
+// names. In the standard case, it's unused and the name is based on the operation id.
+func getTypeName(op *spec.Operation, path string) (string, *resources.NameDisambiguation) {
+	var disambiguation *resources.NameDisambiguation
+	typeName := customresources.GetCustomResourceName(path)
```
danielrbradley, line 570: Nit: would probably be nicer to have an explicit `ok` second return here rather than looking for the default value.

provider/pkg/resources/customresources/custom_pim_eligibility.go
```diff
@@ -0,0 +1,454 @@
...
+//
+// The APIs used are Microsoft.Authorization/roleEligibilitySchedules and Microsoft.Authorization/
+// roleEligibilityScheduleRequests.
+func pimRoleEligibilitySchedule(
+	lookupResource resources.ResourceLookupFunc,
+	crudClientFactory crud.ResourceCrudClientFactory,
+	token azcore.TokenCredential,
+) (*CustomResource, error) {
+	// A bit of a hack to initialize some resource. This func's parameters are all nil when the function is called for
+	// the first time, for customresources.featureLookup, which is ok but would break our initialization here.
```
danielrbradley, line 67: I can't quite grok what you're trying to say here. Is the hack that we want them to be nil or that something breaks inside featureLookup ... why does it break?

provider/pkg/resources/customresources/custom_pim_eligibility.go
```diff
@@ -0,0 +1,454 @@
...
+}
+
+// genSchema modifies the schema extracted from the Azure spec.
+func genSchema(resource *ResourceDefinition) (*ResourceDefinition, error) {
+	if resource == nil {
+		return nil, fmt.Errorf("resource %q not found in the spec", pimRoleEligibilityScheduleRequestTok)
+	}
+
+	// remove requestType, we're only supporting one so far (like TF)
+	delete(resource.Resource.InputProperties, "requestType")
```
danielrbradley, line 125: Should we be asserting that this exists first?

provider/pkg/resources/customresources/custom_pim_eligibility.go
```diff
@@ -0,0 +1,454 @@
...
+	resource.makePropertyNotRequired("requestType")
+
+	// remove the name, it's a random GUID so we generate it ourselves
+	delete(resource.Resource.InputProperties, "roleEligibilityScheduleRequestName")
+
+	// This resource doesn't support direct updates, all changes need to go through a new schedule request.
```
danielrbradley, line 131:
```suggestion
	// All properties cause replacement because this resource doesn't support updates.
```

provider/pkg/resources/customresources/custom_pim_eligibility.go
```diff
@@ -0,0 +1,454 @@
...
+	})
+	for pager.More() {
+		page, err := pager.NextPage(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("listing role eligibility schedules: %w", err)
+		}
+		for _, schedule := range page.Value {
+			if strings.EqualFold(*schedule.Properties.PrincipalID, principalId) &&
+				strings.EqualFold(*schedule.Properties.RoleDefinitionID, roleDefinitionId) &&
+				*schedule.Properties.MemberType == armauthorization.MemberTypeDirect {
```
danielrbradley, line 184: What's the significance of `MemberTypeDirect`? Perhaps worth a comment

provider/pkg/resources/customresources/custom_pim_eligibility.go
```diff
@@ -0,0 +1,454 @@
...
+		return nil, fmt.Errorf("converting role eligibility schedule from SDK to JSON: %w", err)
+	}
+	var js map[string]any
+	err = json.Unmarshal(j, &js)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshaling role eligibility schedule from JSON: %w", err)
+	}
+
+	// We don't need the "Microsoft.Authorization/roleEligibilitySchedules" type that the SDK adds.
+	delete(js, "type")
```
danielrbradley, line 225: Curious - why don't we need this? Does codegen also ignore it?

provider/pkg/resources/customresources/custom_pim_eligibility.go
```diff
@@ -0,0 +1,454 @@
...
+	sdkInputs, err := inputsToSdk(inputs)
+	if err != nil {
+		return nil, false, fmt.Errorf("converting inputs to SDK shape: %w", err)
+	}
+
+	existingSchedule, err := client.findSchedule(ctx, *sdkInputs.Properties.Scope,
+		*sdkInputs.Properties.PrincipalID, *sdkInputs.Properties.RoleDefinitionID)
+	if err != nil {
+		return nil, false, fmt.Errorf("looking up role eligibility schedule: %w", err)
+	}
```
danielrbradley, line 241: Perhaps a note here on why we use find rather than get ... I think it's related to its primary key being an artbitary generated value, which isn't in the inputs map, but the natural key is only usable via the listing?

Annoyingly, if we could put the generated id into the input set, then we could just do a `.get()` which would be more performant, though expect it's trivial in practice given that hopefully a single principleId wouldn't have thousands of schedules.

Guess we could do that via the Check provider method ... but that's not accessible to custom resources right now.

provider/pkg/resources/customresources/custom_pim_eligibility.go
```diff
@@ -0,0 +1,454 @@
...
+		return nil, fmt.Errorf("creating role eligibility schedule request: %w", err)
+	}
+
+	// Poll for completion. Success is when we find a schedule with the matching scope|principal|role tuple.
+	var schedule *armauthorization.RoleEligibilitySchedule
+	err = util.RetryOperation(maxWait, pimRoleEligibilityScheduleTickerInterval,
+		"waiting for role eligibility schedule to be created",
+		func() (bool, error) {
+			var err error
+			schedule, err = client.findSchedule(ctx, *payload.Properties.Scope,
```
danielrbradley, line 284: Why do we not use `client.get` here if we know what the name is?

Do we also need to check any status fields on the returned schedule?

provider/pkg/resources/customresources/customresources.go
```diff
@@ -76,6 +78,11 @@ type CustomResource struct {
 	Update func(ctx context.Context, id string, news, olds resource.PropertyMap) (map[string]interface{}, error)
 	// Delete an existing resource. Constructs the resource ID based on input values.
 	Delete func(ctx context.Context, id string, previousInputs, state resource.PropertyMap) error
+
+	// Only used when we want to specify a `tok` that exists in the spec so we can look up the resource, but we want
+	// the custom resource's name to be different.
+	CustomResourceName string
```
danielrbradley, line 84: Perhaps also add a comment to the `tok` field to explain how it's used to mirror this?

danielrbradley (approved):
A few bits of clean-up but the core implementation looks solid.

provider/pkg/resources/customresources/custom_pim_eligibility.go
thomas11: This is not in the schema, it's added by the SDK.

PR #3968, opened 2025-02-19
Fix schema generation when run under path with dots

tie:
This change fixes schema generation when specifications path contains dots (e.g. `/build/pulumi-azure-native-2.88.0/azure-rest-api-specs/…`).

I assume this a regression introduced in https://github.com/pulumi/pulumi-azure-native/pull/3872#discussion_r1925570106, cc @danielrbradley.

### DIFF
--- a/provider/pkg/resources/resources.go
+++ b/provider/pkg/resources/resources.go
@@ -419,7 +419,7 @@ func findNamespaceWithoutPrefixFromPath(path, defaultValue string) string {
 	return defaultValue
 }
 
-var folderModulePattern = regexp.MustCompile(`.*/specification/([a-zA-Z-]+)/resource-manager/(.*)`)
+var folderModulePattern = regexp.MustCompile(`.*/specification/([a-zA-Z0-9-]+)/resource-manager/(.*)`)
 
 func getSpecFolderNameAndFilePath(path string) (string, string) {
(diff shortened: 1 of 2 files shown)


thomas11 (commented):
Hi @tie, thank you for your contribution! I can see how `if strings.Contains(part, ".")` would falsely match any dots in the path.

Would you mind adding some unit tests to resources_test.go? That way we can also make sure the new code doesn't regress anything. Some edge cases to watch out for are parameters with dots (`/{foo.bar}`) and nested `/specification/`, as you point out in your comment.

thomas11:
Hi @tie, just checking back in if you'd be up for adding some tests? We'd love to have your contribution.

tie:
@thomas11, hi, sorry for the late reply! Added two test cases for `GetModuleName`.

thomas11:
Running `make schema` shows an unexpected difference in `versions/v2.yaml`. A few resources seem to be missing. I haven't investigated yet.

tie:
Looks like regexp does not allow numbers, e.g.
```
/go/pulumi-azure-native/azure-rest-api-specs/specification/storsimple8000series/resource-manager/Microsoft.StorSimple/stable/2017-06-01/storsimple.json
```

thomas11 (approved):
LGTM, thank you!

PR #3980, opened 2025-02-21
Fix for async deletion of non-existent resource

### LINKED ISSUE #3978: Unable to destroy stack containing a deleted subnet
### What happened?

I tried to destroy a stack that had a resource group, subnet, and other resources.  Those resources had already been cleaned up in the cloud.  Pulumi seems unable to destroy the stack, it crashes with a generic error.

```
> pulumi destroy -f
View in Browser (Ctrl+O): https://app.pulumi.com/eron-pulumi-corp/issue-3964/dev/updates/12

     Type                            Name            Status                  Info
     pulumi:pulumi:Stack             issue-3964-dev  **failed**              1 error; 1 warning
 -   └─ azure-native:network:Subnet  subnet          **deleting failed**     1 error

Diagnostics:
  pulumi:pulumi:Stack (issue-3964-dev):
    error: update failed

  azure-native:network:Subnet (subnet):
    error: the operation failed or was cancelled
```

### Example

```ts
import { Config } from '@pulumi/pulumi';
import { ResourceGroup } from '@pulumi/azure-native/resources';
import { VirtualNetwork,Subnet } from '@pulumi/azure-native/network';

const location = 'eastus2';
const azConfig = new Config('azure-native');
const rg = new ResourceGroup('rg', {
  location,
});

const vnet = new VirtualNetwork('vnet', {
  resourceGroupName: rg.name,
  addressSpace: {
    addressPrefixes: ['10.0.0.0/16'],
  }
});

const subnet = new Subnet('subnet', {
  resourceGroupName: rg.name,
  virtualNetworkName: vnet.name,
  addressPrefix: '10.0.0.0/24',
  serviceEndpoints: [
    {
      locations: ['*'],
      service: 'Microsoft.KeyVault',
    }
  ]
});

```

### Output of `pulumi about`

```
❯ p version
v3.150.0
❯ p about
CLI          
Version      3.150.0
Go Version   go1.23.6
Go Compiler  gc

Plugins
resource  azure-native  2.88.0
resource  azure-native  2.88.0
language  nodejs        3.150.0-dev.0

Host     
OS       darwin
Version  14.5
Arch     arm64

Dependencies:
NAME                  VERSION
@pulumi/azure-native  2.88.0
@pulumi/pulumi        3.150.0
@types/node           18.19.76
typescript            5.7.3
```

### Additional context

https://github.com/pulumi/pulumi-azure-native/pull/3536

### Contributing

EronWright:
This PR fixes an error that occurs in the deletion of a (non-existent) Azure resource. The issue impacts resources that have an async-style DELETE call (as is the case for Subnet, see [schema](https://github.com/Azure/azure-rest-api-specs/blob/1b038fc40563bdafb343a71cd5f7d3dee6bc4b5b/specification/network/resource-manager/Microsoft.Network/stable/2019-02-01/virtualNetwork.json#L396-L399)).  The technical detail is that the 404 may come in the initial response as opposed to during polling, and that wasn't handled correctly..

The user-facing error is:
```
  azure-native:network:Subnet (subnet):
    error: the operation failed or was cancelled
```

The call to `runtime.NewPoller` should not be made with a failed response, otherwise we hit a precondition within it:
```go
	// this is a back-stop in case the swagger is incorrect (i.e. missing one or more status codes for success).
	// ideally the codegen should return an error if the initial response failed and not even create a poller.
	if !poller.StatusCodeValid(resp) {
		return nil, errors.New("the operation failed or was cancelled")
	}
```

To fix, I encapsulated the request-to-final-response logic into an anonymous function, to be able to uniformly check for the 404 result. And I ensure that the initial response was successful before making a poller.

This PR relaxes the logic for detecting a 404 during resource deletion, to not care about the specific error code. I find that at least two codes are possible: `ResourceNotFound` and `ResourceGroupNotFound`.  Note that the sync path did not check the error code anyway and a uniform solution is desirable.

TODO:

- [x] new unit tests
- [x] manual testing for various types of resources (w/ sync and async delete operations).

Closes #3978
See also: https://github.com/pulumi/pulumi-azure-native/pull/3783

### DIFF
--- a/provider/pkg/azure/client_azcore_test.go
+++ b/provider/pkg/azure/client_azcore_test.go
@@ -429,6 +429,17 @@ func TestErrorStatusCodes(t *testing.T) {
 		require.NoError(t, err)
 	})
 
+	t.Run("DELETE initial success on 404 when asyncStyle is given", func(t *testing.T) {
+		client := newClientWithPreparedResponses([]*http.Response{
+			{
+				StatusCode: 404,
+				Body:       io.NopCloser(strings.NewReader(`{"error": {"code": "ResourceNotFound"}}`)),
+			},
+		})
+		err := client.Delete(context.Background(), "/subscriptions/123/rg/rg", "2022-09-01", "the actual value doesn't matter!", nil)
+		require.NoError(t, err)
+	})
+
 	t.Run("DELETE polling success on 404 when asyncStyle is given", func(t *testing.T) {
 		client := newClientWithPreparedResponses([]*http.Response{
 			{
@@ -443,6 +454,19 @@ func TestErrorStatusCodes(t *testing.T) {
 		})
 		err := client.Delete(context.Background(), "/subscriptions/123/rg/rg", "2022-09-01", "the actual value doesn't matter!", nil)
 		require.NoError(t, err)
+	})
+
+	t.Run("DELETE initial failure when asyncStyle is given", func(t *testing.T) {
+		client := newClientWithPreparedResponses([]*http.Response{
+			{
+				StatusCode: 400,
+				Body:       io.NopCloser(strings.NewReader(`{"error": {"code": "BadRequest"}}`)),
+			},
+		})
+		err := client.Delete(context.Background(), "/subscriptions/123/rg/rg", "2022-09-01", "the actual value doesn't matter!", nil)
+		require.Error(t, err)
+		require.IsType(t, &PulumiAzcoreResponseError{}, err)
+		require.Equal(t, "BadRequest", err.(*PulumiAzcoreResponseError).ErrorCode)
 	})
 
 	t.Run("DELETE polling failure when asyncStyle is given", func(t *testing.T) {


provider/pkg/azure/client_azcore.go
```diff
@@ -202,36 +202,39 @@ func (c *azCoreClient) Delete(ctx context.Context, id, apiVersion, asyncStyle st
...
-				// Work around https://github.com/pulumi/pulumi-azure-nextgen/issues/120
-				return nil
+		if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusAccepted, http.StatusNoContent) {
+			return newResponseError(resp)
+		}
+
+		// Some APIs are explicitly marked `x-ms-long-running-operation` and we should only do the
+		// poll for the deletion result in that case.
+		if asyncStyle != "" {
+			pt, err := runtime.NewPoller[any](resp, c.pipeline, nil)
```
EronWright, line 217: @thomas11 , I notice that `normalizeLocationHeader` isn't being called here, as is in `putOrPatch`.
Given [this discussion](https://github.com/pulumi/pulumi-azure-native/pull/3536/files#r1728816555), it probably doesn't matter, but I bring it up in case you'd like me to act on that.

provider/pkg/azure/client_azcore.go
```diff
@@ -297,6 +299,9 @@ func (c *azCoreClient) putOrPatch(ctx context.Context, method string, id string,
 			Frequency: time.Duration(c.updatePollingIntervalSeconds * int64(time.Second)),
 		})
 		if err != nil {
+			if err, ok := err.(*azcore.ResponseError); ok {
+				return nil, created, newResponseError(err.RawResponse)
+			}
```
EronWright, line 304: This fix is actually a follow-up to https://github.com/pulumi/pulumi-azure-native/pull/3783.  I think it is justsifiably in here because the Delete call is being similarly fixed here.

provider/pkg/azure/client_azcore_test.go
```diff
@@ -418,7 +429,7 @@ func TestErrorStatusCodes(t *testing.T) {
 			{
 				StatusCode: 503, // temporary failure
 				Header:     http.Header{"Location": []string{"https://management.azure.com/operation"}},
-				Body:       io.NopCloser(strings.NewReader(`{"status": "Unavailable"}`)),
```
EronWright, line 421: When we fake an error response, it is good to use the error schema because you can make assertions about the final error code.  The changes in this section are related to that.

provider/pkg/azure/client_azcore.go
thomas11: nice catch!

provider/pkg/azure/client_azcore.go
```diff
@@ -202,34 +202,35 @@ func (c *azCoreClient) Delete(ctx context.Context, id, apiVersion, asyncStyle st
...
-	resp, err := c.pipeline.Do(req)
-	if err != nil {
-		return err
-	}
-
-	// Some APIs are explicitly marked `x-ms-long-running-operation` and we should only do the
-	// poll for the deletion result in that case.
-	if asyncStyle != "" {
-		pt, err := runtime.NewPoller[any](resp, c.pipeline, nil)
+	err = func() error {
```
thomas11, line 205:
```suggestion
	// Wrap the deletion logic so we can handle potential errors uniformly afterwards.
	err = func() error {
```

thomas11 (approved):

PR #3982, opened 2025-02-24
Re-compress the Python Wheel package with bzip2 to shave off 7 MB.

thomas11:
The provider is currently blocked from releasing the Python SDK because the size of the Wheel package reached 100 MB, see #3977.

While we're waiting on PyPi to increase our limit (if they do), I looked into better compression. Turns out that the package is currently compressed via zip's deflate method, while the better bzip2 would also be valid. 

I first tried to use bzip2 directly when building the package, but failed due to https://github.com/pypa/setuptools/issues/2491. This PR instead builds the Wheel as before, then unzips it and re-zips it with bzip2.

A [test run](https://github.com/pulumi/pulumi-azure-native/actions/runs/13507283628/job/37742131092) shows that it works. The package is 93.8 MB and is uploaded. It fails after upload with _The use of local versions in `Version(2.89.0a1740429149+ded80e4)` is not allowed_ but that's because [provider-version-action adds the local commit when it's a workflow run that's not on the main branch](https://github.com/pulumi/provider-version-action/tree/v1/?tab=readme-ov-file#scenarios).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -403,6 +403,19 @@ export FAKE_MODULE
 		./venv/bin/python -m pip install build && \
 		cd ./bin && \
 		../venv/bin/python -m build .
+
+# Re-encrypt with bzip2 to shrink the package a bit, see #3977
+# The top-level contents of the wheel are pulumi_azure_native and pulumi_azure_native-2.0.0a0+dev.dist-info.
+	cd sdk/python/bin/dist && \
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):
Looks like a good approach. Have downloaded the artifact and it looks okay to me too

PR #3985, opened 2025-02-25
Add exceptions for resources where the user needs to specify the API version

### LINKED ISSUE #3524: Api version `2024-04-02` for resources `Resource`
## Hello!

## Issue details

API version 2024-04-02 is needed to create a `Resource` with type `GitHub.Network/networkSettings`. Currently no API versions support this resource type.

Error message from Azure:
```
No registered resource provider found for location 'germanywestcentral' and API version '2022-09-01' for type 'networkSettings'. The supported api-versions are '2024-04-01-preview, 2023-03-15-alpha, 2023-03-15-beta, 2024-04-02, 2023-11-01-preview, 2024-04-01, 2023-11-01-rc'.
```
None of these API versions are implemented currently

thomas11:
Resolves #3524 

This PR adds an exception for two resources where the `api-version` query parameter works differently. Here, the endpoints receiving the API version merely pass it on to another service, and it's meant to express the API version the second service should use. Therefore, our usual logic of auto-adding the API version doesn't work. The user must instead specify the API version for the second service.

Example:
```go
resources.NewResource(ctx, "acc", &resources.ResourceArgs{
	ResourceProviderNamespace: pulumi.String("Microsoft.Storage"),
	ResourceType:              pulumi.String("storageAccounts"),
	ApiVersion:                pulumi.String("2022-09-01"),
})
```

The provider will make a request to the `resources` service (RP) here. The `resources` service will then in turn make a request to the Microsoft.Storage service. The API version is meant for the latter to receive, so we cannot use whatever API version we have as default for `resources.Resource`.

This PR, specifically fixing `resource.Resource`, unlocks the creation of `GitHub.Network/networkSettings` and other resources that are not in the Azure spec. It will also allow users to use `Resource` as an escape hatch when there are problems with the provider schema.

#2467 is closely related - it's just a matter of adding the `ApiVersion` property for _all_ resources, maybe with a different name.

Potential improvement: the description _"The API version to use for the operation."_ for `ApiVersion` is not useful, we could replace it.

### DIFF
--- a/sdk/dotnet/Authorization/GetManagementLockAtResourceLevel.cs
+++ b/sdk/dotnet/Authorization/GetManagementLockAtResourceLevel.cs
@@ -37,6 +37,12 @@ namespace Pulumi.AzureNative.Authorization
     public sealed class GetManagementLockAtResourceLevelArgs : global::Pulumi.InvokeArgs
     {
         /// <summary>
+        /// The API version to use for this operation.
+        /// </summary>
+        [Input("apiVersion", required: true)]
+        public string ApiVersion { get; set; } = null!;
+
+        /// <summary>
         /// The name of lock.
         /// </summary>
         [Input("lockName", required: true)]
@@ -80,6 +86,12 @@ namespace Pulumi.AzureNative.Authorization
 
     public sealed class GetManagementLockAtResourceLevelInvokeArgs : global::Pulumi.InvokeArgs
     {
+        /// <summary>
+        /// The API version to use for this operation.
+        /// </summary>
+        [Input("apiVersion", required: true)]
+        public Input<string> ApiVersion { get; set; } = null!;
+
         /// <summary>
         /// The name of lock.
         /// </summary>
--- a/sdk/dotnet/Authorization/ManagementLockAtResourceLevel.cs
+++ b/sdk/dotnet/Authorization/ManagementLockAtResourceLevel.cs
@@ -104,6 +104,12 @@ namespace Pulumi.AzureNative.Authorization
     public sealed class ManagementLockAtResourceLevelArgs : global::Pulumi.ResourceArgs
     {
         /// <summary>
+        /// The API version to use for this operation.
+        /// </summary>
+        [Input("apiVersion", required: true)]
+        public Input<string> ApiVersion { get; set; } = null!;
+
+        /// <summary>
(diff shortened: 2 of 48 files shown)


provider/pkg/gen/schema.go
```diff
@@ -1579,6 +1579,24 @@ func isMethodParameter(param *openapi.Parameter) bool {
...
 
+func (m *moduleGenerator) isApiVersionQueryParam(param *openapi.Parameter, bodySchema *openapi.Schema) bool {
+	if param.Name != "api-version" || param.In != "query" {
+		return false
+	}
+
+	// #3524: for this resource, the api-version is passed through to another Azure service, so the user needs to specify it.
+	if m.moduleName == "Resources" && m.resourceName == "Resource" {
+		return false
+	}
```
danielrbradley, line 1590: This naming is a little confusing it's saying the function is checking if it's a query param but then returns false for the _Resource_ resource even though it is going into the query. Perhaps this function is something more like "isApiVersionNotSettableByUser"?

provider/pkg/gen/schema_test.go
```diff
@@ -404,3 +404,71 @@ func TestResourcePathTracker(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			m := &moduleGenerator{
+				moduleName:   tt.moduleName,
+				resourceName: tt.resourceName,
+			}
+			got := m.isApiVersionQueryParam(tt.param, nil)
+			assert.Equal(t, tt.want, got)
+		})
+	}
```
danielrbradley, line 473: Nit: nested tests are better when reporting failures to users than table based tests because the line number of the error for a table based test will take you to the same line (:471) rather than the test case. Additionally, it makes it harder to read as the data for the test is not in the same place as code performing the steps.

For example, you could replace each case struct with the actual code. This is only 1 extra line of code per test, but then saves 17 lines overall with the table, struct and loop, making it more consice too.

```go
t.Run("standard api-version query param", func(t *testing.T) {
	m := &moduleGenerator{
		moduleName:  "SomeModule",
		resourceName: "SomeResource",
	}
	got := m.isApiVersionQueryParam(createParam("api-version", "query"), nil)
	assert.Equal(t, true, got, "should not be settable by user")
})
```

danielrbradley (commented):
I _think_ the logic looks okay based on the tests, but I think both the implementation function name and tests could be quite easily refactored to be more understandable.

danielrbradley (approved):
:shipit:

PR #3988, opened 2025-02-26
Revert unreleased spec updates to get Wheel size below 100m

thomas11:
Also reverts _"Re-compress the Python Wheel package with bzip2 to shave off 7 MB. (#3982)"_ which didn't work because Pypi doesn't accept bzip2 even though the Wheel standard allows it.


danielrbradley (commented):
Rather than reverting the whole spec upgrade, I think we previously just started adding exclusions so we get updates & fixes to the existing resource versions we're already using, but don't add any new resources to the SDK.

thomas11:
> Rather than reverting the whole spec upgrade, I think we previously just started adding exclusions so we get updates & fixes to the existing resource versions we're already using, but don't add any new resources to the SDK.

True. That seems like a bunch of extra work for a hopefully short-lived situation, though. We can still do a future release with such updates if we need it, which I think isn't very likely.

danielrbradley (approved):
Yup, this is fine to make the provider releasable as-is to get the associated fixes released. We can re-visit pinning next week if needed.

PR #3996, opened 2025-02-27
Read "official" default API versions from spec readmes

thomas11:
TODO
- 4 modules have "composite" or empty versions
- because we're reading the readme.md files from `main` or close to it, we're picking versions that are too new
- not taking into account [service groups](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/glossary.md#service-group)
  - most modules have a single readme.md here: `quota/resource-manager/readme.md`
  - but service groups have one per service:
    - `containerservice/resource-manager/Microsoft.ContainerService/aks/readme.md`
    - `containerservice/resource-manager/Microsoft.ContainerService/fleet/readme.md`
  - they can have different API versions
  - some modules like `workloads` even have the op-level readme and service group readmes; not sure what that means
- make this new way of getting default versions easily swappable with the previous one
- allow manual overrides


provider/pkg/openapi/defaultApiVersions.go
```diff
@@ -0,0 +1,91 @@
...
+			inYamlBlock = true
+		}
+		if inYamlBlock && line == "```" {
+			inYamlBlock = false
+		}
+
+		if inYamlBlock && strings.HasPrefix(line, "tag: package-") {
+			version = strings.TrimSpace(strings.TrimPrefix(line, "tag: package-"))
+			break
+		}
```
danielrbradley, line 50: Looking at some more readme files, I think we need to do at least one more step of processing to determine what the version selection is.

Where there's an initial code block such as:

````markdown
``` yaml
title: ApiManagementClient
description: ApiManagement Client
openapi-type: arm
tag: package-preview-2024-06
```
````

The tag will correspond with other, conditional code blocks such as:

````markdown
``` yaml $(tag) == 'package-preview-2024-06'
input-file:
  - Microsoft.ApiManagement/preview/2024-06-01-preview/apigateway.json
  - Microsoft.ApiManagement/preview/2024-06-01-preview/apimallpolicies.json
  - Microsoft.ApiManagement/preview/2024-06-01-preview/apimanagement.json
  - ...
```
````

The tag can be somewhat arbitrary, but the `input-file` list is the ultimate output we need.

Overall, I think the parsing approach for the data from a readme file is as follows (though we could possibly short-cut some of this, given we're not needing to support user inputs on the tag to use):
1. Iterate through each yaml code block.
2. Check its condition against the current collected state.
3. Merge the block with the current parsed yaml state.

## Worked example

With the two blocks defined above, after parsing the first block would result in a state of:
```yaml
title: ApiManagementClient
description: ApiManagement Client
openapi-type: arm
tag: package-preview-2024-06
```

When evaluating the second block, the value of `$(tag)` would be `package-preview-2024-06` and the condition evaluates to true. Therefore the state after parsing the second block is:

```yaml
title: ApiManagementClient
description: ApiManagement Client
openapi-type: arm
tag: package-preview-2024-06
input-file:
  - Microsoft.ApiManagement/preview/2024-06-01-preview/apigateway.json
  - Microsoft.ApiManagement/preview/2024-06-01-preview/apimallpolicies.json
  - Microsoft.ApiManagement/preview/2024-06-01-preview/apimanagement.json
  - ...
```

danielrbradley: We will probably need to turn the list of files back into a list of resources with version instead to produce a lockfile.

danielrbradley: https://github.com/Azure/autorest/blob/e691282a8919f9c850c2f77e6ba180ed1635e8de/packages/libs/configuration/src/configuration-manager/configuration-file.ts#L33-L57 🥳

EronWright:
I was thinking that we would avoid duplication of their core parsing logic by having our tools consume the "code model" that is produced by autorest. This is possible by running autorest with a certain flag to retain the code model, or by installing an autorest language plugin (that would target Pulumi schema).

Have you considered an approach like that?

PR #4003, opened 2025-03-03
Add overrideApiVersion to resources in v3+

thomas11:
- **Test rename, follow-up to 3985**
- **Generate only default versions**
- **Add overrideApiVersion parameter**
- **Integration test for overrideApiVersion**

### DIFF
--- a/examples/ts-override-api-version/index.ts
+++ b/examples/ts-override-api-version/index.ts
@@ -1,0 +1,19 @@
+// Copyright 2025, Pulumi Corporation.
+
+import * as resources from "@pulumi/azure-native/resources";
+import * as storage from "@pulumi/azure-native/storage";
+
+// Create an Azure Resource Group
+const resourceGroup = new resources.ResourceGroup("resourceGroup");
+
+// Create an Azure resource (Storage Account)
+const storageAccount = new storage.StorageAccount("sa", {
+    apiVersionOverride: "2024-01-01",
+    resourceGroupName: resourceGroup.name,
+    sku: {
+        name: storage.SkuName.Standard_LRS,
+    },
+    kind: storage.Kind.StorageV2,
+});
+
+export const storageAccountName = storageAccount.name;
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -412,6 +412,14 @@ func TestAccPIMRoleManagementPolicies(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestOverrideApiVersion(t *testing.T) {
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: filepath.Join(getCwd(t), "ts-override-api-version"),
+		})
+	integration.ProgramTest(t, &test)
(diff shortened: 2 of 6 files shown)


provider/pkg/gen/schema.go
```diff
@@ -1054,6 +1059,24 @@ func isSingleton(resource *resourceVariant) bool {
 	return resource.PathItem.Delete == nil || customresources.IsSingleton(resource.Path)
 }
 
+func (g *packageGenerator) addApiVersionOverride(resourceRequest *parameterBag, defaultApiVersion openapi.ApiVersion) {
+	if g.majorVersion < 3 {
+		return
+	}
+	if _, hasSettableApiVersion := resourceRequest.specs["apiVersion"]; !hasSettableApiVersion {
```
EronWright, line 1066: Could you add some commentary about which resources are targeted by this special case? I thought that `apiVersion` was seen in some of Azure's API Management stuff, and doesn't have the same meaning as here. Or does it? Either way please explain.

provider/pkg/provider/provider.go
```diff
@@ -1551,7 +1551,8 @@ func (k *azureNativeProvider) Delete(ctx context.Context, req *rpc.DeleteRequest
 					k.host.Log(ctx, diag.Warning, urn, fmt.Sprintf("error converting inputs to request body: %v", err))
 				}
 
-				queryParams := map[string]interface{}{"api-version": res.APIVersion}
+				// TODO,tkappler check if overrideApiVersion is set
+				queryParams := map[string]any{"api-version": res.APIVersion}
```
EronWright, line 1555: This is getting tricky. Consider how some delete operations are async, and the schema tells you which "await" strategy to use.  Imagine that the await strategy has changed over different versions of the resource, which is entirely possible.  Which strategy should you use? The answer seems clear.

This seems like another case where having schema is vital to correctness.  Should the user be allowed to use an override for which we don't have schema?  Seems wrong to me.

provider/pkg/gen/schema.go
```diff
@@ -50,6 +50,8 @@ const goModuleVersion = "/v2"
 // pulumiProviderName is the name of the provider as used in all tokens.
 const pulumiProviderName = "azure-native"
 
+const apiVersionOverride = "apiVersionOverride"
```
EronWright, line 53: In https://github.com/pulumi/pulumi-azure-native/pull/3907 we're adding a new property for tracking the api version that is in state, tentatively called `AzureApiVersion`.  There was a clash with `ApiVersion`.  So, I would advocate for this new property to be called `AzureApiVersionOverride`.  Or, maybe the two things should be unified as `AzureApiVersion`.

examples/ts-override-api-version/index.ts
EronWright, line 1: Could this example show more of the value proposition?  Setting the apiversion seems like means to an end, but what end? Is it to be able to set additional fields?

EronWright (changes requested):
We should reconcile this enhancement with https://github.com/pulumi/pulumi-azure-native/pull/3907, they are practically the same thing.

PR #4010, opened 2025-03-05
Add Parameterize to generate an SDK for a specific API version

thomas11:
```
> PULUMI_DEBUG_PROVIDERS="azure-native:60383" pulumi package add azure-native storage v20240101
Successfully generated a Nodejs SDK for the azure-native_storage_v20240101 package at /Users/tkappler/pulumi/repro/azn-param/sdks/azure-native_storage_v20240101

> head -3 index.ts
import * as storage_v20240101 from "@pulumi/azure-native_storage_v20240101";

> rg _storage package.json
10:        "@pulumi/azure-native_storage_v20240101": "file:sdks/azure-native_storage_v20240101",

> PULUMI_DEBUG_PROVIDERS="azure-native:60383" pulumi up
...
 +   └─ azure-native_storage_v20240101:storage/v20240101:StorageAccount  sa             created (27s) 
```

I picked an approach that's, I hope, pragmatic rather than elegant, in that I assume we can keep the full schema and metadata available for the provider and generate dedicated versioned schemas simply by picking the required parts from the full schema. This avoids having to publish and download or clone data at generation time. A more dynamic approach is equally possible, of course, and could be swapped in without changing the interface.

To understand the PR you should be familiar with [these docs](https://pulumi-developer-docs.readthedocs.io/latest/docs/architecture/providers/parameterized.html).

The input and behavior of Parameterize is the same in the initial "args" call of Parameterize and in the subsequent "serialized parameters" calls: a pair of (Azure module, API version). The existing full schema and metadata are narrowed down to members belonging to the given module at the given version.

_Implementation note on `resources.PartialAzureAPIMetadata` and `resources.APIMetadata`_: the provider used to store metadata in a `PartialAzureAPIMetadata`. During Parameterize, we couldn't replace this metadata with the new one which was of type `AzureAPIMetadata`. Using the new `APIMetadata` we can accomodate both because it uses `MapLike` which can be a full or a partial map. It turned out that this change also simplified some other code using these types.

Related to #2467

### DIFF
--- a/provider/pkg/provider/provider_parameterize.go
+++ b/provider/pkg/provider/provider_parameterize.go
@@ -10,8 +10,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/openapi"
 	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources"
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/logging"
@@ -48,6 +50,33 @@ func deserializeParameterizeArgs(in []byte) (*parameterizeArgs, error) {
 	return &args, nil
 }
 
+// parseApiVersion parses an Azure API version from the given string. Since the input comes from users (via `package
+// add`), it tries to be lenient and accept several formats. The returned result, if successful, is an "SDL version" of
+// the form v20200101.
+func parseApiVersion(version string) (string, error) {
+	v := strings.TrimSpace(version)
+	v = strings.TrimLeft(v, "vV")
+
+	isApiVersion, err := regexp.MatchString(`^20\d{2}-[01]\d-[0-3]\d(-preview|-privatepreview)?$`, v)
+	if err != nil {
+		return "", status.Errorf(codes.InvalidArgument, "unexpected error parsing version %s: %v", version, err)
+	}
+	if isApiVersion {
+		apiVersion := openapi.ApiVersion(v)
+		return string(apiVersion.ToSdkVersion()), nil
+	}
+
+	isDigits, err := regexp.MatchString(`^20\d{6}(preview|privatepreview)?$`, v)
+	if err != nil {
+		return "", status.Errorf(codes.InvalidArgument, "unexpected error parsing version %s: %v", version, err)
+	}
+	if isDigits {
+		return "v" + v, nil
+	}
+
+	return "", status.Errorf(codes.InvalidArgument, "invalid API version: %s", version)
+}
+
 // "When generating an SDK (e.g. using a pulumi package add command), we need to boot up a provider and parameterize it
 // using only information from the command-line invocation. In this case, the parameter is a string array representing
 // the command-line arguments (args)."
@@ -62,9 +91,13 @@ func getParameterizeArgs(req *rpc.ParameterizeRequest) (*parameterizeArgs, error
 		if len(args) != 2 {
 			return nil, status.Errorf(codes.InvalidArgument, "expected 2 arguments (module and API version), got %d", len(args))
 		}
+		targetApiVersion, err := parseApiVersion(args[1])
+		if err != nil {
+			return nil, err
+		}
 		return &parameterizeArgs{
 			Module:  args[0],
-			Version: args[1], // TODO be more accepting of version format and convert
+			Version: targetApiVersion,
 		}, nil
 
 	// provider has already been parameterized and the arguments were serialized into the SDK
--- a/provider/pkg/provider/provider_parameterize_test.go
+++ b/provider/pkg/provider/provider_parameterize_test.go
@@ -192,6 +192,53 @@ func TestFilterTokens(t *testing.T) {
 	require.Equal(t, "Application", names["azure-native:aad/v20221201:Application"])
 }
 
+func TestParseApiVersion(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Valid", func(t *testing.T) {
+		t.Parallel()
+
+		for _, v := range [][]string{
+			{"2020-01-01", "v20200101"},
+			{"2027-01-01", "v20270101"},
+			{"2020-12-31", "v20201231"},
+			{"v2020-12-31", "v20201231"},
+			{"V2020-12-31", "v20201231"},
+			{"v20230301", "v20230301"},
+			{"V20230301", "v20230301"},
+			{"20230301", "v20230301"},
+			{" 20230301 ", "v20230301"},
+			{"2023-03-01-preview", "v20230301preview"},
+			{"2023-03-01-privatepreview ", "v20230301privatepreview"},
+			{"v20230301preview", "v20230301preview"},
+			{"v20230301privatepreview ", "v20230301privatepreview"},
+		} {
+			parsed, err := parseApiVersion(v[0])
+			require.NoError(t, err)
+			assert.Equal(t, v[1], parsed)
+		}
+	})
+
+	t.Run("Invalid", func(t *testing.T) {
+		t.Parallel()
+
+		for _, v := range []string{
+			"1999-01-01",
+			"2027-011-01",
+			"2020-12-311",
+			"2020-12-41",
+			"2020-22-31",
+			"v2023030",
+			"V202303011",
+			"19990301",
+			"2023-03-01-alpha",
+		} {
+			_, err := parseApiVersion(v)
+			assert.Error(t, err, v)
+		}
+	})
+}
+
 // Ensure that we can run `pulumi package add` with a local provider binary and get a new SDK.
 func TestParameterizePackageAdd(t *testing.T) {
 	t.Parallel()


provider/pkg/provider/provider.go
```diff
@@ -133,11 +152,24 @@ func LoadMetadata(azureAPIResourcesBytes []byte) (*resources.AzureAPIMetadata, e
 func (k *azureNativeProvider) lookupTypeDefault(ref string) (*resources.AzureAPIType, bool, error) {
 	typeName := strings.TrimPrefix(ref, "#/types/")
 	t, ok, err := k.resourceMap.Types.Get(typeName)
-	return &t, ok, err
+	if err != nil {
+		return nil, false, err
+	}
+	if !ok {
+		t, ok = k.parameterizedMetadata.Types[typeName]
+	}
```
danielrbradley, line 160: I don't think this should be required as-is, because we can only copy types that already exist in the `k.resourceMap` into the `k. parameterizedMetadata` object. If we were to drop including all the explicit versions in the embedded metadata, then this would be required .. but perhaps the `Parameterize` method would just overwrite both the Metadata and Schema in the provider instance rather than adding fallbacks and switches throughout the code to determine if we're running in a "parameterized" mode.

thomas11: The values of the entries in `k.parameterizedMetadata` are the same than the ones in `k.resourceMap` but the keys are not.

danielrbradley: Oh, because we're stripping the module and version off the tokens?

Hmm, perhaps we should then replace the metadata at the point of parameterization then so we only have one state within the provider.

Also, how do we re-populate the `parameterizedMetadata` when starting the provider process from the generated SDK?

... oh .. this was tested in debug mode right ... so we didn't restart the provider .. so this would likely fail when used outside of debug attach. So we do need to implement the metadata rehydration within the `runtimeParameterize` method.

thomas11: > perhaps we should then replace the metadata at the point of parameterization then so we only have one state within the provider.

This is where my mental model is a bit hazy still. Is there a provider instance per parameterization, plus one non=parameterized one? Then we can just replace the metadata and the schema. But how would the lookup of the correct provider work, then? `GetSchema` gets a name and a version argument so it can look up the correct schema. See also the [p-k8s implementation](https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/provider/provider.go#L260). `crdSchemas` is populated during parameterization.

> we didn't restart the provider .. so this would likely fail when used outside of debug attach. So we do need to implement the metadata rehydration within the runtimeParameterize method.

Yes, exactly, that's the TODO (from before I had fully thought through it).

mjeffryes: Replacement is the current model; you can rely on the engine only calling the parameterized provider for resources from the parameterized SDK. (It's considered an entirely new and different provider once it's parameterized.)

The engine might eventually support "extension parameterization" which would have a different paradigm of each parameterization "extending" the schema of the original provider. The primary motivation for doing this would be so that the user doesn't have to configure the parameterized provider all over again.

thomas11: Doesn't that mean the p-k8s implementation I linked doesn't actually need the `crdSchemas` map because there's only ever one schema? Why does `GetSchema` have the `SubpackageName` and `SubpackageVersion` input parameters?

mjeffryes: k8s parameterization was written in anticipation of the extension parameterization support (which has yet to materialize). The problem of configuring multiple providers is more acute for CRDs b/c 1) k8s takes a lot of config and 2) it's very normal to use may CRDs with one k8s cluser.

rquitales: Are we not able to add in `provider.parameterizedSchemas` support in the future when it is necessary. Parameterization occurs during provider startup, so it should be seamless to start a secondary `parameterizedSchemas` map in the future when needed.

provider/pkg/provider/provider_parameterize.go
```diff
@@ -0,0 +1,315 @@
...
+	metadataBytes, err := serializeMetadata(metadata)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to serialize metadata: %v", err)
+	}
+	newSchema.Parameterization = &pschema.ParameterizationSpec{
+		BaseProvider: pschema.BaseProviderSpec{
+			Name:    p.name,
+			Version: newSchema.Version,
+		},
+		Parameter: metadataBytes,
```
rquitales, line 252: This is the most important part for parameterization in my opinion since this is what gets generated into the output SDKs and stored at rest. An update to the data stored here would require users to regenerate their SDKs to get these changes.

For my benefit, would you be able to describe more (I'm coming in with no context) on what we're serializing and storing here? Just want to make sure we are capturing all the required information we need, with leeway in case we plan on adding more features in the future.

provider/pkg/provider/provider.go
```diff
@@ -92,12 +105,18 @@ func makeProvider(host *provider.HostClient, name, version string, schemaBytes [
...
+		version:              version,
+		resourceMap:          resourceMap,
+		config:               map[string]string{},
+		schemaBytes:          schemaBytes,
+		parameterizedSchemas: &parameterizedPackageMap{},
+		parameterizedMetadata: &resources.AzureAPIMetadata{
+			Resources: map[string]resources.AzureAPIResource{},
+			Types:     map[string]resources.AzureAPIType{},
+			Invokes:   map[string]resources.AzureAPIInvoke{},
+		},
```
danielrbradley, line 119: Nit: Why immediately initialize these as empty rather than being a nil pointer until parameterization has taken place - I'd expect the nil would be an easier thing to check for than empty.

provider/pkg/provider/provider.go
danielrbradley, line 160: Rather than treating this as a fallback, perhaps we should be checking if `k.parameterizedMetadata` is not nil first?

provider/pkg/provider/provider_parameterize.go
```diff
@@ -0,0 +1,315 @@
...
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// parameterizedPackageMap maps parameterized package names to their schema, in form of serialized JSON. Will be empty
+// if Parameterize() hasn't been called.
+type parameterizedPackageMap struct {
+	sync.Mutex
+	schemas map[string][]byte
```
danielrbradley, line 30: Might be nicer to replace the lock based approach with a `sync.map` - with your methods just adding the typing.
```suggestion
	schemas sync.Map
```

danielrbradley (changes requested):
Looking at the implementation around the locked map of parameterized schema made me realise that there's possibly a misunderstanding of the lifetime of parameterized providers .. 

1. Each parameterization of a provider is started as its own process.
2. A provider can only be parameterized once.
3. A provider that's been parameterized won't be used as an unparameterized provider.

One caveat here is that if at some point in the future we implement "exetension parameterization", then the instance would be shared, but the tokens would also need reworking to live in the same provider module space as one another too.

Therefore, I think this code can be drastically simplified to remove all the maps and locks and change the approach on the lookup methods.

provider/pkg/provider/provider.go
thomas11: Only one of them can be populated, right? So order wouldn't matter?

thomas11:
> Looking at the implementation around the locked map of parameterized schema made me realise that there's possibly a misunderstanding of the lifetime of parameterized providers ..
> 
> 1. Each parameterization of a provider is started as its own process.
> 2. A provider can only be parameterized once.
> 3. A provider that's been parameterized won't be used as an unparameterized provider.
> 
> One caveat here is that if at some point in the future we implement "exetension parameterization", then the instance would be shared, but the tokens would also need reworking to live in the same provider module space as one another too.
> 
> Therefore, I think this code can be drastically simplified to remove all the maps and locks and change the approach on the lookup methods.

I based my implementation mostly on the one in p-kubernetes, which is written in a way that anticipates extension parameterization. For the sake of simplicity, I have now removed all that code.

provider/pkg/provider/provider.go
danielrbradley: Probably not critical .. it's just a little misleading that it's treated as a fallback rather that determining which mode we're running in, then choosing the correct variable to query.

provider/pkg/provider/autonaming_test.go
```diff
@@ -123,7 +123,7 @@ func TestGetDefaultName(t *testing.T) {
 
 		assert.True(t, ok)
 		assert.True(t, randomlyNamed)
-		assert.Equal(t, "test1ec836c1", result.StringValue())
+		assert.Equal(t, "test614742c2", result.StringValue())
```
thomas11, line 126: This changed with the p/p upgrade.

provider/pkg/provider/__debug_bin2036324868
EronWright, line 1: To be removed

provider/pkg/provider/provider.go
```diff
@@ -133,11 +144,24 @@ func LoadMetadata(azureAPIResourcesBytes []byte) (*resources.AzureAPIMetadata, e
...
 func (k *azureNativeProvider) LookupResource(resourceType string) (resources.AzureAPIResource, bool, error) {
-	return k.resourceMap.Resources.Get(resourceType)
+	res, found, err := k.resourceMap.Resources.Get(resourceType)
+	if err != nil {
+		return resources.AzureAPIResource{}, false, err
+	}
+	if !found {
+		res, found = k.parameterizedMetadata.Resources[resourceType]
+	}
+	return res, found, nil
```
EronWright, line 164: Hopefully these structures can be unified in future?  In a sense, all providers are "parameterized"?

EronWright (commented):
This is starting to be very cool.

provider/pkg/provider/provider_parameterize.go
thomas11: I changed the approach to only store the actual parameters now, the Azure API module and version.

provider/pkg/provider/provider.go
thomas11: simplified it down to a single map

provider/pkg/provider/provider.go
thomas11: unified in the latest commit!

provider/pkg/provider/provider_parameterize.go
```diff
@@ -0,0 +1,265 @@
...
+		Functions:   map[string]pschema.FunctionSpec{},
+	}
+
+	makeToken := func(name string) string {
+		return fmt.Sprintf("%s:%s/%s:%s", newPackageName, targetModule, targetApiVersion, name)
+	}
+
+	typeNames, err := filterTokens(schema.Types, targetModule, targetApiVersion)
+	if err != nil {
+		return nil, nil, status.Errorf(codes.InvalidArgument, "failed to parse token: %v", err)
```
EronWright, line 180: Consider mentioning "schema.Types" in the error message here, likewise for the other two sites below.

EronWright (approved):
Looks good to me.

provider/pkg/provider/autonaming_test.go
danielrbradley: Is this just for new values being generated? But existing values are unchanged, I assume?

provider/pkg/provider/provider_parameterize.go
```diff
@@ -0,0 +1,265 @@
...
+	case req.GetArgs() != nil:
+		args := req.GetArgs().Args
+
+		// "aad" "v20221201"
+		if len(args) != 2 {
+			return nil, status.Errorf(codes.InvalidArgument, "expected 2 arguments (module and API version), got %d", len(args))
+		}
+		return &parameterizeArgs{
+			Module:  args[0],
+			Version: args[1], // TODO be more accepting of version format and convert
```
danielrbradley, line 67: The version is definitely worth validating a little more up front to only the formats we accept. Though it will fail later when we try to find the spec, it would probably have a nicer error from here.

provider/pkg/provider/autonaming_test.go
thomas11: This is the name generator. Any names that are already out there are fixed.

provider/pkg/provider/provider_parameterize.go
```diff
@@ -0,0 +1,265 @@
...
+//
+// There are other ways of obtaining such a schema, for instance, generating it directly from the Azure spec. This way
+// was the most pragmatic approach since we already have the full schema in the provider.
+//
+// All names change because their "module" part needs to match the new schema, and the new package name can't be just
+// "azure-native" which already exists.
+//
+// To separate concerns, the `Parameterization` of the new schema is NOT populated yet, the caller is responsible for
+// doing that.
+func createSchema(p *azureNativeProvider, schema pschema.PackageSpec, targetModule, targetApiVersion string) (*pschema.PackageSpec, *resources.APIMetadata, error) {
```
danielrbradley, line 153: A general thought here .. we should measure how fast this step is and how much memory the schema consumes in memory. I'd guess it's not too much, but if we wanted to make this a little more efficient, we could just generate the metadata and delay doing the schema until we call GetSchema. Might not be worth the effort though if this is only taking a fraction of a second.

danielrbradley (approved):
Looks great!

provider/pkg/provider/provider_parameterize.go
thomas11: I'm planning to accept a few formats and then convert them to the v20xxxxxx we need for lookups.

provider/pkg/provider/provider_parameterize.go
thomas11: Good call.

I added `runtime.GC()` at the top and ` pprof.Lookup("heap")` at the end of Parameterize. From `go tool pprof`:

```
(pprof) focus=github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider.TestParameterizeCreatesSchemaAndMetadata

(pprof) top10
Active filters:
   focus=github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider.TestParamet…
Showing nodes accounting for 160.35MB, 40.29% of 398.03MB total
      flat  flat%   sum%        cum   cum%
  159.20MB 40.00% 40.00%   159.20MB 40.00%  os.readFileContents
    1.16MB  0.29% 40.29%     1.16MB  0.29%  runtime/pprof.StartCPUProfile
         0     0% 40.29%     1.16MB  0.29%  github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider.(*azureNativeProvider).Parameterize
         0     0% 40.29%   160.35MB 40.29%  github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider.TestParameterizeCreatesSchemaAndMetadata
```

Looks like Parameterize itself only uses 1.16MB, the rest is reading the full schema in the test. (Please double-check my reasoning.)

For CPU, I used `StartCPUProfile`/`StopCPUProfile`. From `go tool pprof`:

```
0     0% 28.97%      0.62s 28.97%  github.com/pulumi/pulumi-azure-native/v2/provider/pkg/provider.(*azureNativeProvider).Parameterize
```

Looks harmless to me although I'm not exactly a pprof expert.

provider/pkg/provider/provider_parameterize.go
thomas11: Pushed

provider/pkg/provider/provider_parameterize.go
danielrbradley: Coolio, yeah that looks fine .. can always dig into optimisations later .. e.g. if we find people wanting to use 20 different parameterizations or something and running out of memory - perhaps we can purge the old schema and metadata.

provider/pkg/provider/provider_parameterize.go
```diff
@@ -48,6 +50,33 @@ func deserializeParameterizeArgs(in []byte) (*parameterizeArgs, error) {
...
 }
 
+// parseApiVersion parses an Azure API version from the given string. Since the input comes from users (via `package
+// add`), it tries to be lenient and accept several formats. The returned result, if successful, is an "SDL version" of
+// the form v20200101.
+func parseApiVersion(version string) (string, error) {
+	v := strings.TrimSpace(version)
+	v = strings.TrimLeft(v, "vV")
+
+	isApiVersion, err := regexp.MatchString(`^20\d{2}-[01]\d-[0-3]\d$`, v)
```
danielrbradley, line 60: Sidethought: this would probably be a useful function to live in the openapi module

danielrbradley: Oh, does this handle all the preview/private-preview variations?

thomas11: It does now :)

Would be cool to move this to the `openapi` package but there are some other concerns there, like handling `*` wildcards and the empty string meaning the default version. Punting.

PR #4016, opened 2025-03-11
Generate a default API versions only schema and use it for sdkgen in version 3+

### LINKED ISSUE #3422: V3: shrink package size

thomas11:
For v3 of the provider, we want to generate SDKs only for the default API versions. However, to enable parameterization to generate SDKs for any API version at runtime, we need the full schema and metadata. This PR adds a build step of generating a default-versions-only schema that's then used to generate the SDKs. The full schema is embedded into the provider as before.

A branch with this PR but regenerated for `PROVIDER_VERSION=3.0.0` is [here](https://github.com/pulumi/pulumi-azure-native/tree/tkappler/v3-default-only-genv3).

Resolves #3422


EronWright (approved):
Looks good to me.

PR #4030, opened 2025-03-17
Only initialize custom resources if provider is not parameterized

thomas11:


provider/pkg/provider/provider.go
```diff
@@ -259,10 +259,13 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 		return nil, fmt.Errorf("creating Azure client: %w", err)
 	}
 
-	k.customResources, err = customresources.BuildCustomResources(&k.environment, k.azureClient, k.LookupResource, k.newCrudClient, k.subscriptionID,
-		resourceManagerBearerAuth, resourceManagerAuth, keyVaultBearerAuth, userAgent, credential)
-	if err != nil {
-		return nil, fmt.Errorf("initializing custom resources: %w", err)
+	// When the provider is parameterized, resources and types that custom resources are built on will probably not be available.
+	if !k.isParameterized() {
+		k.customResources, err = customresources.BuildCustomResources(&k.environment, k.azureClient, k.LookupResource, k.newCrudClient, k.subscriptionID,
```
danielrbradley, line 264: What if the custom resource is within the requested module?

I guess some custom resources only live in the default version, but should we pass an argument into `BuildCustomResources` and let it decide which custom resources to add?

thomas11: Interesting question. When a customized resource is in a module that the user parameterizes, we don't know if the API version the user requested is compatible with the API version the custom resource is written against. But we could match against (module, api-version). That, however, would require first updating our custom resources with explicit API version information.

danielrbradley: Looking through the actual cases:
- `storage:BlobContainerLegalHold` is a hard-coded to look up the default API verison.
- `keyvault:AccessPolicy` is written against the SDK.
- `keyvault:Secret` and `keyvault:Key` are written against the SDK, but only override Delete, so could be used for any version I think
- `authorization:PimRoleEligibilitySchedule` and `authorization:RoleEligibilityScheduleRequest` are written directly against the SDK and are fully custom.
- `authorization:RoleManagementPolicy` is just capturing the initial state and restoring - that should work for any API verison I think, but is currently only in the default version.
- `portal:Dashboard` is only in the default version, but is just altering the schema so should be able to do the same for other versions.
- `dbforpostgresql:Configuration` is just overriding the Delete.
- `recoveryservices:ProtectedItem` is just overriding the `id` and `query` params.
- `storage:StorageAccountStaticWebsite` is a full custom implementation.
- `storage:Blob` is a full custom implementation.
- `web:WebApp` and `web:WebAppSlot` have custom Read and Delete and lookup default API version.

So, I don't think any of these appear in explicit versions previously, and therefore won't appear in parameterized schemas either, so I think this is a non-issue.

Though I think some might be very easy to port over at some point in the future.

thomas11: thanks for checking them!

provider/pkg/provider/provider.go
```diff
@@ -273,6 +276,17 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 	}, nil
 }
 
+func (k *azureNativeProvider) isParameterized() bool {
+	var p struct {
+		Parameterization map[string]any `json:"parameterization"`
+	}
+	err := json.Unmarshal(k.schemaBytes, &p)
```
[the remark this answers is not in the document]
thomas11: True; updated.

danielrbradley (approved):

PR #4031, opened 2025-03-17
Update `$ref` references in metadata when parameterizing

thomas11:
This fixes an omission in #4010. Metadata contains `$ref`'s that need to be updated when parameterizing because all the tokens change, like in the schema.

This fixes this error I had brought up elsewhere:
> error: azure-native_storage_v20240101:storage/v20240101:StorageAccount resource 'sa' has a problem: schema type 'azure-native:storage/v20240101:Sku' not found

### DIFF
--- a/provider/pkg/provider/provider_parameterize.go
+++ b/provider/pkg/provider/provider_parameterize.go
@@ -148,11 +148,16 @@ func (p *azureNativeProvider) Parameterize(ctx context.Context, req *rpc.Paramet
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to marshal schema: %v", err)
 	}
+	s = updateRefs(s, newPackageName)
 
-	s = bytes.ReplaceAll(s, []byte(`"$ref":"#/types/azure-native`), []byte(`"$ref": "#/types/`+newPackageName))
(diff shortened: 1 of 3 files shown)


danielrbradley (commented):
Why are we using the approach of serializing then doing the matching on the string rather than iterating through the objects and replacing the values within the objects? Have you tried recursing through the objects to do the ref replacement?

I can't immediately see why this should break with the strings, but I'm also not super-confident that there wouldn't be some case where it doesn't work.

thomas11:
> Why are we using the approach of serializing then doing the matching on the string rather than iterating through the objects and replacing the values within the objects? Have you tried recursing through the objects to do the ref replacement?
> 
> I can't immediately see why this should break with the strings, but I'm also not super-confident that there wouldn't be some case where it doesn't work.

The string approach is not elegant but
1. much less code
2. much safer IMO - it's easy to forget some place when iterating properties, and references have a well-defined serialization
3. we need to do it anyway for the schema, unless we want to unmarshaler the raw schema, traverse it as well (requiring a different implementation), and marshaling it again.

To avoid these downsides, we could use a generic walker based on reflection that visits every property. But that's no safer than string replacement and probably slower as well.

danielrbradley (approved):
Happy with the rationalization 👍

PR #4032, opened 2025-03-17
Add an integration test for parameterization and moving from default resource to parameterized

thomas11:
This integration test proves that
1. `pulumi package add` works and produces a usable SDK,
2. users can migrate from a default resource (here, a storage account) to a parameterized one without replacement, and
3. the default provider and parameterized ones can be used in the same program without issues.

**This PR contains and requires #4030, otherwise the test will fail.**


examples/examples_nodejs_test.go
```diff
@@ -412,13 +418,103 @@ func TestAccPIMRoleManagementPolicies(t *testing.T) {
...
+	// Pending #4019
+	// pt.UpdateSource(t, cwd, "parameterize", "3-back-to-default")
+	// pt.Refresh(t, optrefresh.ExpectNoChanges())
+	// pt.Up(t, optup.ExpectNoChanges())
+
+	pt.Destroy(t)
+}
+
+// A helper to use with optup.DebugLogging() and friends to get verbose logging.
+func debugLogging() debug.LoggingOptions {
```
danielrbradley, line 457: Unused?

examples/examples_nodejs_test.go
```diff
@@ -412,13 +418,103 @@ func TestAccPIMRoleManagementPolicies(t *testing.T) {
...
+// engine to resolve, or a path to a provider binary.
+func runPulumiPackageAdd(
+	t *testing.T,
+	pt *pulumitest.PulumiTest,
+	provider string,
+	args ...string,
+) {
+	ctx := context.Background()
+	allArgs := append([]string{"package", "add", provider}, args...)
+	stdout, stderr, exitCode, err := pt.CurrentStack().Workspace().PulumiCommand().Run(
```
danielrbradley, line 491: That's pretty nice rather than just using the ambient `pulumi` binary 👍

danielrbradley (approved):
Overall looks good 👍 

Looks like it includes the same changes as https://github.com/pulumi/pulumi-azure-native/pull/4030 .. won't repeat my comments on those bits.

PR #4035, opened 2025-03-18
Support static website Read() from id in addition to state to enable import

### LINKED ISSUE #4028: Importing ` azure-native:storage:StorageAccountStaticWebsite` fails with `"accountName" not found in resource state`
### What happened?

When attempting to import a `StorageAccountStaticWebsite`, the preview fails with `"accountName" not found in resource state`

### Example

Create a Storage Account with a Static Website configuration.

Attempt to import it:

```
pulumi import azure-native:storage:StorageAccountStaticWebsite <storage account>/staticwebsite /subscriptions/<subId>/resourceGroups/<rgName>/providers/Microsoft.Storage/storageAccounts/<storage account>/staticWebsite
```

Error message:

```
     Type                                                 Name                          Plan       Info
     pulumi:pulumi:Stack                                  azure-py-import-dev                      1 error
 =   └─ azure-native:storage:StorageAccountStaticWebsite  importsapulkay/staticwebsite  import     1 error

Diagnostics:
  azure-native:storage:StorageAccountStaticWebsite (importsapulkay/staticwebsite):
    error: Preview failed: "accountName" not found in resource state

  pulumi:pulumi:Stack (azure-py-import-dev):
    error: preview failed
```

### Output of `pulumi about`

```
Version      3.156.0
Go Version   go1.24.1
Go Compiler  gc

Plugins
resource  azure-native  2.89.1
language  python        3.156.0

Host     
OS       darwin
Version  15.3.2
Arch     arm64

This project is written in python: executable='azure-py-import/.venv/bin/python' version='3.13.2'

TYPE                                  URN
pulumi:pulumi:Stack                   urn:pulumi:dev::azure-py-import::pulumi:pulumi:Stack::azure-py-import-dev
pulumi:providers:azure-native         urn:pulumi:dev::azure-py-import::pulumi:providers:azure-native::default_2_89_1
azure-native:resources:ResourceGroup  urn:pulumi:dev::azure-py-import::azure-native:resources:ResourceGroup::rs_rg_import
azure-native:storage:StorageAccount   urn:pulumi:dev::azure-py-import::azure-native:storage:StorageAccount::importsapulkay

Dependencies:
NAME                 VERSION
pulumi_azure_native  2.89.1
```

### Additional context

Possibly somewhat related to #3357?

### Contributing

thomas11:
When importing, there is no state yet and `Read` is called with only the id. That's enough but the previous code didn't expect that and failed.

Resolves #4028.

### DIFF
--- a/provider/pkg/resources/customresources/custom_storage_azidentity.go
+++ b/provider/pkg/resources/customresources/custom_storage_azidentity.go
@@ -130,13 +130,20 @@ func (r *staticWebsite_azidentity) newStorageAccountClient(properties resource.P
 	return newStorageAccountClient(properties, r.env, r.creds)
 }
 
+func (r *staticWebsite_azidentity) newStorageAccountClientForAccount(accName string) (*azblob.Client, error) {
+	return newStorageAccountClientForAccount(accName, r.env, r.creds)
+}
+
 func newStorageAccountClient(properties resource.PropertyMap, env cloud.Configuration, creds azcore.TokenCredential) (*azblob.Client, error) {
 	acc := properties[accountName]
 	if !acc.HasValue() || !acc.IsString() {
 		return nil, errors.Errorf("%q not found in resource state", accountName)
 	}
 	accName := acc.StringValue()
+	return newStorageAccountClientForAccount(accName, env, creds)
+}
 
+func newStorageAccountClientForAccount(accName string, env cloud.Configuration, creds azcore.TokenCredential) (*azblob.Client, error) {
 	saUrl, err := getStorageAccountURL(accName, env)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


provider/pkg/resources/customresources/custom_storage_azidentity.go
```diff
@@ -130,13 +130,20 @@ func (r *staticWebsite_azidentity) newStorageAccountClient(properties resource.P
 	return newStorageAccountClient(properties, r.env, r.creds)
 }
 
+func (r *staticWebsite_azidentity) newStorageAccountClientForAccount(accName string) (*azblob.Client, error) {
+	return newStorageAccountClientForAccount(accName, r.env, r.creds)
+}
```
danielrbradley, line 135: This new method threw me a little when reading the diff being the same name as the thing it calls within. Looks like it's only called in one place, could we just call inner function directly there?

danielrbradley (approved):
This looks okay to me .. I assume this was only a bug in the azidentity version?

provider/pkg/resources/customresources/custom_storage_azidentity.go
thomas11: This pattern mirrors the existing `newStorageAccountClient` that also exists twice. I think it makes sense - they do the same thing but the method on the resource needs fewer arguments.

thomas11:
> This looks okay to me .. I assume this was only a bug in the azidentity version?

It's a bug in both but I don't think we should spend time fixing the autorest code anymore. I'm hoping to remove it soon.

PR #4036, opened 2025-03-18
In v3, generate default versions only SDK for Go, too

thomas11:
This is a follow-up to #4016 where I had forgotten that Go SDK generation works differently than the other languages. Fortunately, the code already accepts the required arguments, we just need to pass them.

This PR has no effect on the SDK in v2.x, only in v3.x.

I wonder if we should remove the "default" schema location altogether and just make the argument mandatory.


danielrbradley:
@thomas11 could you expand the PR description to explain what changes are expected to be seen in in the generated Go SDK?

danielrbradley:
Comparing the generation results of running `make build_go` between this branch and master, there is no reported difference.

image

danielrbradley (approved):

EronWright (approved):

PR #4042, opened 2025-03-19
Disambiguate enums for Single Server MySQL/Postgres from Flexible Server counterparts

thomas11:
Fixes [this test failure on v3](https://github.com/pulumi/pulumi-azure-native/actions/runs/13947231847/job/39038489959):
> azure-native:dbforpostgresql:Server server creating (4s) error: Status=400 Code="ParameterOutOfRange" Message="The value of the 'Sku.Tier' should be in: [Burstable, GeneralPurpose, MemoryOptimized]. Verify that the specified parameter value is correct."

The failure is caused by #3817 being incomplete. It disambiguated resource names, but types are still shared and not disambiguated.

This PR is only a band-aid for the particular error. The other resources that were disambiguated in #3817 are missing, as well as types that are not enums.

### DIFF
--- a/examples/cs-postgres-configuration/Stack.cs
+++ b/examples/cs-postgres-configuration/Stack.cs
@@ -19,7 +19,7 @@ class MyStack : Stack
             Sku = new SkuArgs
             {
                 Name = "Standard_B1ms",
-                Tier = SkuTier.Basic
+                Tier = SkuTier.Burstable
             },
             Storage = new StorageArgs
             {


EronWright:
Reference:
- [postgresql.SkuTier](https://learn.microsoft.com/en-us/python/api/azure-mgmt-rdbms/azure.mgmt.rdbms.postgresql.models.skutier?view=azure-python)
- [postgresql_flexibleservers.SkuTier](https://learn.microsoft.com/en-us/python/api/azure-mgmt-rdbms/azure.mgmt.rdbms.postgresql_flexibleservers.models.skutier?view=azure-python)

provider/pkg/gen/types.go
```diff
@@ -262,6 +262,13 @@ func (m *moduleGenerator) genEnumType(schema *spec.Schema, context *openapi.Refe
 	}
 	enumName = m.typeNameOverride(ToUpperCamel(enumName))
 
+	if m.moduleName == "DBforMySQL" && m.resourceName == "SingleServer" {
+		enumName = "SingleServer" + enumName
+	}
+	if m.moduleName == "DBforPostgreSQL" && m.resourceName == "SingleServer" {
+		enumName = "SingleServer" + enumName
+	}
```
mjeffryes, line 270: should we add a test case for this?

EronWright:
Looking at the upstream Azure SDK, seems like the flexible servers are in a different package (postgresql_flexibleservers), so that the SkuTier type doesn't clash.  Are other types likely to clash if we continue down this road of renaming the types?  Can we use a different namespace?

danielrbradley (approved):
This looks okay to me. The actual change is scoped nice and tight to the single resources.

Tricky to fully review the SDK diffs, but again, look fine on the surface.

PR #4043, opened 2025-03-19
Fix generic resources by passing the correct API version to Read and Delete

thomas11:
This PR is a follow-up to #3985. It was incomplete which went undetected because of an accidental poor choice of API version in the integration test. The test used to use 2022-09-01 which happened to be a valid version for both the `resources` and the `storage` module.

### DIFF
--- a/provider/pkg/provider/crud/crud.go
+++ b/provider/pkg/provider/crud/crud.go
@@ -44,7 +44,8 @@ type ResourceCrudOperations interface {
 	CanCreate(ctx context.Context, id string) error
 	CreateOrUpdate(ctx context.Context, id string, bodyParams, queryParams map[string]any) (map[string]any, bool, error)
 	// `overrideApiVersion` is (rarely) used for generic resources like resources:Resource where the API version is
-	// specified by the user and passed on to another Azure service.
+	// specified by the user and passed on to another Azure service. Leaving it empty means use the API version from
+	// resource metadata, which is the standard case.
 	Read(ctx context.Context, id string, overrideApiVersion string) (map[string]any, error)
 	HandleErrorWithCheckpoint(ctx context.Context, err error, id string, inputs resource.PropertyMap, properties *structpb.Struct) error
 }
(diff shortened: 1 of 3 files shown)


provider/pkg/provider/crud/crud.go
```diff
@@ -43,7 +43,9 @@ type AzureRESTConverter interface {
 type ResourceCrudOperations interface {
 	CanCreate(ctx context.Context, id string) error
 	CreateOrUpdate(ctx context.Context, id string, bodyParams, queryParams map[string]any) (map[string]any, bool, error)
-	Read(ctx context.Context, id string) (map[string]any, error)
+	// `overrideApiVersion` is (rarely) used for generic resources like resources:Resource where the API version is
+	// specified by the user and passed on to another Azure service.
+	Read(ctx context.Context, id string, overrideApiVersion string) (map[string]any, error)
```
danielrbradley, line 48: It looks like a blank string means it's not overridden? Might be worth noting in the comment

provider/pkg/provider/crud/crud.go
```diff
@@ -83,7 +83,9 @@ type ResourceCrudClient interface {
 	ResourceCrudOperations
 	AzureRESTConverter
 	SubresourceMaintainer
-	ApiVersion() string
+	// ApiVersion returns the API version for the resource. The second return value is the value of
+	// `ApiVersionIsUserInput` on the underlying resources.AzureAPIResource.
+	ApiVersion() (string, bool)
```
danielrbradley, line 88: Why not have these two return values as separate methods which indicate what their value means directly rather than relying on the comment?

provider/pkg/resources/resources.go
```diff
@@ -99,7 +99,10 @@ type AzureAPIType struct {
 type AzureAPIResource struct {
 	// API version in "2020-10-01" format.
 	APIVersion string `json:"apiVersion"`
-	Path       string `json:"path"`
+	// True if the "apiVersion" parameter is set by the user, not the provider.
+	ApiVersionIsUserInput bool `json:"apiVersionIsUserInput,omitempty"`
```
danielrbradley, line 103: Is this only true for the generic resource or are there other resources affected by this too?

thomas11: There's one other that we know of, `authorization:ManagementLockAtResourceLevel`. The definite list is [here](https://github.com/pulumi/pulumi-azure-native/blob/999017ccc376b938675f0c590c180c3741c6de91/provider/pkg/gen/schema.go#L1597).

thomas11:
> I think this is fine, though it's a little odd because I can't actually see any changes which show how this manifests or its impact - as I assume it only affects the metadata which we don't commit?

It only affects metadata, yes. The visible effect is that the integration test passes on v3.

> Does this only affect the generic resource or are there others?

It affects the generic resource and `authorization:ManagementLockAtResourceLevel`.

danielrbradley (approved):

PR #4045, opened 2025-03-20
Add more type name overrides for conflicting enums in v3

thomas11:
This PR replaces #4042 with a different solution, using an override mechanism we already have.

There are no visible changes here because the overrides are gated to v3+. Here's [the v3 branch with this change cherry-picked and regenerated](https://github.com/pulumi/pulumi-azure-native/tree/v3) and a [master build with this change](https://github.com/pulumi/pulumi-azure-native/actions/runs/13970791886).


danielrbradley (approved):
Ah that SDK change looks much better - adding the new types alongside the old ones then switching some references

PR #4049, opened 2025-03-20
On v3, we need to get the user API version from previous inputs, not `__inputs`

thomas11:
Follow-up to #4043 which works on v2 but not on v3, because of #3772.

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -1270,6 +1270,15 @@ func (k *azureNativeProvider) Read(ctx context.Context, req *rpc.ReadRequest) (*
 	if err != nil {
 		return nil, err
 	}
+	stateWithPreviousInputs := oldState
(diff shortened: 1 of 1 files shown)


danielrbradley (commented):
This feels somewhat backwards, if I'm reading it correctly ...

If we're on v3 .. then we're going to do extra work to take the inputs from the request and replace the oldState with those inputs?

Should we not be modifying the methods to no longer assume there's `__inputs` in the oldState?

thomas11:
> This feels somewhat backwards, if I'm reading it correctly ...
> 
> If we're on v3 .. then we're going to do extra work to take the inputs from the request and replace the oldState with those inputs?
> 
> Should we not be modifying the methods to no longer assume there's `__inputs` in the oldState?

Well, the code in this PR is written to work on v2 as well, where it will read `__inputs`. On v3, we don't need it and can later remove it.

PR #4053, opened 2025-03-24
Update path conflict resolution for newer API versions

thomas11:
Turns out that some of them are not necessary anymore and removing them simplifies the resource names. We know they're not necessary because of the check in schema.go that fails the build when there are path conflicts. This PR also makes this check required for non-release builds now that all de-duplications are in place.


danielrbradley (commented):
> Turns out that some of them are not necessary anymore and removing them simplifies the resource names

What was the process of determining they're not needed? Is it just removing and seeing what happens?

thomas11:
> > Turns out that some of them are not necessary anymore and removing them simplifies the resource names
> 
> What was the process of determining they're not needed? Is it just removing and seeing what happens?

I noticed it originally when documenting the deduplications for the upgrade guide. Re-running schemagen without them confirmed it.

PR #4056, opened 2025-03-25
Add a message about other API versions to v3 registry API docs

### LINKED ISSUE #4054: Update resource descriptions to mention parameterization

thomas11:
Resolves #4054

This results in no changes for v2, and looks like in this screenshot in v3.

Screenshot 2025-03-25 at 10 48 43

Details to be discussed, like the exact wording and whether there should be a paragraph break before "_These versions are not included ..._".

The biggest issue I see is hard-coding the registry link in provider code, but I couldn't think of a better way.


provider/pkg/gen/schema_test.go
```diff
@@ -462,3 +464,135 @@ func TestIsApiVersionQueryParam(t *testing.T) {
...
+
+		desc := g.formatDescription("This is a wonderful resource.", "SomeResource", "2022-02-22", "2021-01-01", pulumi.StringRef("Additional docs."))
+		assert.Equal(t, `This is a wonderful resource.
+
+Azure REST API version: 2022-02-22. Prior API version in Azure Native 2.x: 2021-01-01.
+
+Other available API versions: 2023-03-03, 2024-04-04.
+
+`+generateApiVersionsMsg+`
+
```
danielrbradley, line 595: An idea for the wording ... and perhaps just give the CLI command rather than a generic docs link?

```suggestion
This resource uses the Azure REST API version: `2022-02-22`. In version 2 of the Azure Native provider, it used the Azure REST API version: `2021-01-01`.

This resource also supports the Azure REST API versions: `2023-03-03`, `2024-04-04`. These can be accessed by generating a local SDK pacakge using the CLI command: `pulumi package add azure-native SomeModule [ApiVersion]`.

```

mjeffryes: I like the wording here better. Ambivalent whether we give the command or link to the doc (or both).

thomas11: I assume "here" = Daniel's?

thomas11: Reworded by taking most of the suggestion but trying to keep it a bit more concise. Let me know what you think.

provider/pkg/gen/schema_test.go
```diff
@@ -462,3 +464,134 @@ func TestIsApiVersionQueryParam(t *testing.T) {
...
+				},
+			},
+		}
+
+		desc := g.formatDescription("This is a wonderful resource.", "SomeResource", "2022-02-22", "2021-01-01", pulumi.StringRef("Additional docs."))
+		assert.Equal(t, `This is a wonderful resource.
+
+Azure REST API version: 2022-02-22. In version 2.x of the Azure Native provider, it used API version 2021-01-01.
+
+Other available API versions: 2023-03-03, 2024-04-04. These can be accessed by generating a local SDK package using the CLI command `+"`"+`pulumi package add azure-native authorization [ApiVersion]`+"`"+`. See the [version guide](../../../version-guide/#accessing-any-api-version-via-local-packages) for details.
```
mjeffryes, line 593: Theoretically, version "2021-01-01" should also show up in the list of "Other available API versions" right?

provider/pkg/gen/schema_test.go
```diff
@@ -462,3 +464,134 @@ func TestIsApiVersionQueryParam(t *testing.T) {
...
+				getAllVersions: func(moduleName openapi.ModuleName, resource string) []openapi.ApiVersion {
+					return []openapi.ApiVersion{"2023-03-03", "2024-04-04"}
+				},
+			},
+		}
+
+		desc := g.formatDescription("This is a wonderful resource.", "SomeResource", "2022-02-22", "2021-01-01", pulumi.StringRef("Additional docs."))
+		assert.Equal(t, `This is a wonderful resource.
+
+Azure REST API version: 2022-02-22. In version 2.x of the Azure Native provider, it used API version 2021-01-01.
```
mjeffryes, line 591:
```suggestion
Uses Azure REST API version: 2022-02-22. In version 2.x of the Azure Native provider, it used API version 2021-01-01.
```
Nit. I think making the first a complete sentence makes the second sentence read better.

mjeffryes (approved):
This looks pretty reasonable to me.

provider/pkg/gen/schema_test.go
thomas11: Yes, in practice, the previous major version's API version will be available. I just hadn't added it to the test data here.

PR #4058, opened 2025-03-25
Update and standardize on Gradle 8.13

thomas11:


thomas11:
> Looks fine .. how did we decide on this specific version? Are we keeping this in sync with something in the ci-mgmt or pulumi-java repo? Perhaps worth adding to the description.

It was simply the latest at the time. I don't want to get in to the business of manually looking up version numbers in other repos.

PR #4062, opened 2025-03-27
Only update ref's that point to the parameterized module, and simplify type tokens

thomas11:
It's not guaranteed that `$ref`'s only point to the same module. For instance, there's `specification/common-types`.


danielrbradley (approved):
Looks good - the diff in the test is :chefkiss: to review

thomas11:
> Looks good - the diff in the test is :chefkiss: to review

Turns out I mixed up two changes that I wanted to keep in separate PRs. Oh well, pushed the rest of the other one to this PR.

PR #4068, opened 2025-03-31
Factor API version handling more cleanly

thomas11:


danielrbradley (commented):
Looking much better 👍  Can then simplify the `getApiVersionFromInputs` further to expect to only be passed the inputs object, not the state.

danielrbradley (approved):
:shipit:

PR #4074, opened 2025-04-02
GetSchema should return only default API versions; also zip schemas to reduce provider size

### LINKED ISSUE #4072: Get Schema should return the default-versions only schema for V3
## Hello!

## Issue details

`pulumi package get-schema azure-native@v3.0.0` returns the full 77 MB schema while I expect it to return the smaller one that wouldn't include any API versions. YAML consumption is very slow right now, presumably because of the schema size. Also things like AI tools won’t get confused.

thomas11:
This PR combines two changes that could be separate but I haven't made the effort yet to disentangle them:
- Return only default versions from GetSchema
- Zip both full and default versions schemas to reduce provider size

Return the default API versions schema is more correct, since it reflects the actually available resources that are in the SDKs, and speeds up YAML consumption. Also things like AI tools won’t get confused.

The effect of zipping the schemas on the size of the provider binary `pulumi-resource-azure-native`: 387M down to 165M.

Unzipping of the smaller default versions schema takes ~70ms on my machine.

Resolves #4072

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -312,10 +312,10 @@ dist/pulumi-azure-native_$(PROVIDER_VERSION)_checksums.txt: dist/$(PROVIDER)-v$(
 	cp bin/metadata-compact.json provider/cmd/$(PROVIDER)
 	cp -v versions/v${MAJOR_VERSION}.yaml provider/pkg/versionLookup/default-versions.yaml
 
-	gzip --keep bin/schema-full.json
+	gzip --keep --force bin/schema-full.json
 	cp bin/schema-full.json.gz provider/cmd/$(PROVIDER)
 
-	gzip --keep bin/schema-default-versions.json
+	gzip --keep --force bin/schema-default-versions.json
 	cp bin/schema-default-versions.json.gz provider/cmd/$(PROVIDER)
 
 	@touch $@


iwahbe (approved):
Beyond using gzip instead of zip, everything else looks good.

.gitignore
```diff
@@ -15,7 +15,8 @@ package-lock.json
 provider/coverage.txt
 provider/cmd/**/metadata.go
 provider/cmd/**/schema.go
-provider/cmd/**/schema-full.json
+provider/cmd/**/schema-full.json.zip
+provider/cmd/**/schema-default-versions.json.zip
```
mikhailshilkov, line 19: Is it zip or gz?

provider/pkg/provider/provider_parameterize.go
```diff
@@ -149,13 +154,17 @@ func (p *azureNativeProvider) Parameterize(ctx context.Context, req *rpc.Paramet
...
+		return nil, status.Errorf(codes.Internal, "failed to gzip new schema: %v", err)
+	}
 
 	newMetadata, err = updateMetadataRefs(newMetadata, newPackageName, args.Module, args.Version)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to update metadata $refs: %v", err)
 	}
 
-	p.schemaBytes = s
+	p.defaultSchemaZipped = newSchemaZipped
```
mikhailshilkov, line 167: Given we override it on parametrization, should we call it something different - not `defaultSchemaZipped`? It's a full current schema, can be default or can be an API version.

provider/pkg/provider/provider_parameterize.go
```diff
@@ -149,13 +154,17 @@ func (p *azureNativeProvider) Parameterize(ctx context.Context, req *rpc.Paramet
 		return nil, status.Errorf(codes.Internal, "failed to marshal schema: %v", err)
 	}
 	s = updateRefs(s, newPackageName, args.Module, args.Version)
+	newSchemaZipped, err := util.GzipBytes(s)
```
mikhailshilkov, line 157: This is slightly wasteful - we zip the in-memory schema only to unzip it later, if I understand it correctly. Should we store the raw JSON schema instead?

provider/pkg/provider/provider_parameterize.go
```diff
@@ -120,7 +121,11 @@ func (p *azureNativeProvider) Parameterize(ctx context.Context, req *rpc.Paramet
 	logging.V(9).Infof("Creating parameterized Azure Native for %s %s", args.Module, args.Version)
 
 	var schema pschema.PackageSpec
-	err = json.Unmarshal(p.schemaBytes, &schema)
+	fullSchemaUnzipped, err := util.GunzipBytes(p.fullSchemaZipped)
```
mikhailshilkov, line 124: Any hint on how much time this takes?

mikhailshilkov (commented):
Looks good overall, a few tactical questions

EronWright:
The fact is that the v3 provider does support both the explicitly-versioned types and the unversioned types.  It simply generates a subset into the published SDK.  In an "extension" type of parameterization, one would simply need a generated SDK while continuing to use the standard provider. 

Also this would be a breaking change.  For these reasons, am closing.

PR #4174, opened 2025-05-31
new provider option: useDefaultAzureCredential

### LINKED ISSUE #4172: Support AZURE_* environment variables in addition to ARM_*
## Hello!

## Issue details

It would be very helpful if the environment variables populated by Azure Workload Identity (AZURE_CLIENT_ID, AZURE_TENANT_ID, etc...) were use for configuration, without having to copy them to their ARM_* equivalents. This is particularly significant in the Pulumi Kubernetes Operator, where there is no easy way to dynamically set the ARM_ variables to their AZURE_ counterparts.

EronWright:
This PR enables the ["Default Credential Chain"](https://learn.microsoft.com/en-us/azure/developer/go/sdk/authentication/credential-chains#defaultazurecredential-overview) feature of the Azure SDK for Go:

> [DefaultAzureCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential) is an opinionated, preconfigured chain of credentials. It's designed to support many environments, along with the most common authentication flows and developer tools.

This feature is enabled with a new provider option, `useDefaultAzureCredential` and the environment variable `ARM_USE_DEFAULT_AZURE_CREDENTIAL`. When enabled, the provider looks for `AZURE_*` environment variables, then a Azure workload or managed identity, then an active CLI account.

### Set the Azure environment

To configure the active Azure environment, set the provider option `environment` or the environment variable `ARM_ENVIRONMENT` or `AZURE_ENVIRONMENT`  to `public`, `usgov`, or `china`.

### Exclude a credential type category

To exclude all "Developer tool" or "Deployed service" credentials, set environment variable `AZURE_TOKEN_CREDENTIALS` to `prod` or `dev`, respectively.

### Use a specific credential

To exclude all credentials except for one, set environment variable `AZURE_TOKEN_CREDENTIALS` to one of: 

- `AzureCLICredential`
- `AzureDeveloperCLICredential`
- `EnvironmentCredential`
- `ManagedIdentityCredential`
- `WorkloadIdentityCredential`

### Manual Setup

```
curl -fsSL https://get.pulumi.com | sh
pulumi plugin install resource azure-native 3.8.0-alpha.1756942349
pulumi config set azure-native:useDefaultAzureCredential true
pulumi config set azure-native:subscriptionId <subscription-id>
```

Closes #4172

### DIFF
--- a/provider/pkg/gen/__snapshots__/gen_aliases_test_v2.snap
+++ b/provider/pkg/gen/__snapshots__/gen_aliases_test_v2.snap
@@ -437,6 +437,10 @@
     "description": "The Tenant ID which should be used.",
     "type": "string"
    },
+   "useDefaultAzureCredential": {
+    "description": "Use the default credential chain of the Azure SDK (see https://learn.microsoft.com/en-us/azure/developer/go/sdk/authentication/credential-chains#defaultazurecredential-overview).",
+    "type": "boolean"
+   },
    "useMsi": {
     "description": "Allow Managed Service Identity to be used for Authentication.",
     "type": "boolean"
--- a/provider/pkg/gen/__snapshots__/gen_aliases_test_v3.snap
+++ b/provider/pkg/gen/__snapshots__/gen_aliases_test_v3.snap
@@ -437,6 +437,10 @@
     "description": "The Tenant ID which should be used.",
     "type": "string"
    },
+   "useDefaultAzureCredential": {
+    "description": "Use the default credential chain of the Azure SDK (see https://learn.microsoft.com/en-us/azure/developer/go/sdk/authentication/credential-chains#defaultazurecredential-overview).",
+    "type": "boolean"
+   },
    "useMsi": {
     "description": "Allow Managed Service Identity to be used for Authentication.",
     "type": "boolean"
--- a/provider/pkg/gen/__snapshots__/gen_vnet_test.snap
+++ b/provider/pkg/gen/__snapshots__/gen_vnet_test.snap
@@ -11402,6 +11402,10 @@
     "description": "The Tenant ID which should be used.",
     "type": "string"
    },
+   "useDefaultAzureCredential": {
+    "description": "Use the default credential chain of the Azure SDK (see https://learn.microsoft.com/en-us/azure/developer/go/sdk/authentication/credential-chains#defaultazurecredential-overview).",
+    "type": "boolean"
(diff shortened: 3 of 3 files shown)


EronWright:
This fix is blocked on a deeper decoupling of the old/new sdks:

https://github.com/pulumi/pulumi-azure-native/compare/master...issue-4047

EronWright:

- #4174 ◀

provider/pkg/gen/schema.go
```diff
@@ -161,6 +161,11 @@ func PulumiSchema(rootDir string, modules openapi.AzureModules, versioning Versi
 					Description: "The URL to initiate the OIDC token exchange. ",
 				},
 
+				"useDefaultAzureCredential": {
+					TypeSpec:    pschema.TypeSpec{Type: "boolean"},
+					Description: "Use the default credential chain of the Azure SDK (see https://learn.microsoft.com/en-us/azure/developer/go/sdk/authentication/credential-chains#defaultazurecredential-overview).",
```
mjeffryes, line 166: I wonder if we shouldn't have a doc of our own to point at that explains how this strategy layers in with our other auth strategies. (IIUC, the existing strategies try to mimic the TF provider env variables, while this new strategy will respect the AZURE_* env variables used by the azure cli. But I imagine there's a lot more nuance involved.)

provider/pkg/provider/auth_azidentity.go
```diff
@@ -98,15 +98,31 @@ func NewAzCoreIdentity(ctx context.Context, authConf *authConfiguration, baseCli
...
+//     2. Service Principal with client certificate
+//     3. Service Principal with client secret
+//     4. OIDC
+//     5. Managed Identity
+//     6. Azure CLI
 //   - When a method is configured but instantiating the credential fails, we return an error and do not fall through to
 //     the next method.
 //   - Auxiliary or additional tenants are supported for SP with client secret and CLI authentication, not for others.
 func newSingleMethodAuthCredential(authConf *authConfiguration, baseClientOpts azcore.ClientOptions) (azcore.TokenCredential, error) {
+	if authConf.useDefault {
```
mjeffryes, line 111: Would we ever want to enable this by default (say in a future major version)? It seems like it's designed to support a set of auth strategies that overlap with existing strategies for this provider.

Issue #4176, opened 2025-06-03
userAssignedIdentities Issue With Typescript


johnhaczewski:
### What happened?

When attempting the code below I get an error due to a mismatch on how Typescript is handling typing vs what the Azure control plane API expects

### Example

```
const funtionManagedIdentity = new UserAssignedIdentity(`${stackPrefix}-app-mi`, {
  resourceGroupName: ticketApiResourceGroup.name,
});

const functionApp = new web_v20240401.web.v20240401.WebApp(
  `functionapp`,
  {
    ...
    identity: {
      type: "UserAssigned",
      userAssignedIdentities: [funtionManagedIdentity.id]
    },
    ...
);
```

Error:

```
  pulumi:pulumi:Stack ():
    error: update failed

  azure-native_web_v20240401:web/v20240401:WebApp ():
    error: Status=400 Code="InvalidRequestContent" Message="The request content was invalid and could not be deserialized: 'Cannot populate JSON array onto type 'System.Collections.Generic.Dictionary`2[System.String,Microsoft.WindowsAzure.Governance.PolicyService.Models.Policy.UserAssignedResourceIdentity]'. Path 'identity.userAssignedIdentities'.'."
```

-----

The following works:

```
identity: {
      type: "UserAssigned",
      userAssignedIdentities: output(funtionManagedIdentity.id).apply((id) => ({
        [id]: {},
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
      })) as any,
    },
```

### Output of `pulumi about`

```
CLI          
Version      3.173.0
Go Version   go1.24.3
Go Compiler  gc

Plugins
KIND      NAME    VERSION
language  nodejs  3.173.0

Host     
OS       ubuntu
Version  24.04
Arch     x86_64

This project is written in nodejs: executable='/home/user/.nvm/versions/node/v23.10.0/bin/node' version='v23.10.0'

Current Stack: organization/CoreInfraBuilder-Hub/stack

Found no pending operations associated with {stack}

Backend        
Name           hostname
URL            azblob://{path}
User           johnh
Organizations  
Token type     personal

Dependencies:
NAME                                VERSION
@eslint/js                          9.28.0
@types/node                         18.19.81
@typescript-eslint/eslint-plugin    8.33.1
@typescript-eslint/parser           8.33.1
typescript-eslint                   8.33.1
typescript                          5.8.2
eslint                              9.28.0
base64-js                           1.5.1
[...]

Issue #3324, opened 2024-06-03
Fix the `identity` field in the Golang `dbforpostgresql.Server` example documentation


zbuchheit:
### What happened?

When attempting to user a UserAssignedIdentity with `dbforpostgresql.Server` in Golang, a user ran into issues and wasn't able to get the example from ServerCreatewithDataEncryptionEnabled to work properly. 

### Example

This example from the documentation doesn't work as intended

```
    identity: {
        type: azure_native.dbforpostgresql.IdentityType.UserAssigned,
        userAssignedIdentities: {
            "/subscriptions/ffffffff-ffff-ffff-ffff-ffffffffffff/resourceGroups/testresourcegroup/providers/Microsoft.ManagedIdentity/userAssignedIdentities/test-usermanagedidentity": {},
        },
    },
```

while this is what I was able get to work for the user
```
			Identity: &dbforpostgresql.UserAssignedIdentityArgs{
				Type: pulumi.String(dbforpostgresql.IdentityTypeUserAssigned),
				UserAssignedIdentities: userAssignedIdentity.ID().ApplyT(func(id string) (dbforpostgresql.UserIdentityMap) {
					return dbforpostgresql.UserIdentityMap{
						id: dbforpostgresql.UserIdentityArgs{},
					}
				}).(dbforpostgresql.UserIdentityMapOutput),
			},
```

### Output of `pulumi about`

```
CLI          
Version      3.117.0
Go Version   go1.22.3
Go Compiler  gc

Plugins
KIND      NAME          VERSION
resource  azure-native  2.43.0
resource  azure-native  2.43.0
resource  azure-native  2.43.0
resource  azure-native  2.42.1
resource  azure-native  2.42.1
language  go            unknown

Host     
OS       darwin
Version  14.2.1
Arch     arm64

This project is written in go: executable='/opt/homebrew/bin/go' version='go version go1.22.3 darwin/arm64'

Current Stack: zbuchheit-pulumi-corp/azure-postgresql-go/dev

Found no resources associated with dev

Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/zbuchheit-pulumi-corp
User           zbuchheit-pulumi-corp
Organizations  zbuchheit-pulumi-corp
Token type     personal

Dependencies:
NAME                                                          VERSION
github.com/pulumi/pulumi-azure-native-sdk/dbforpostgresql/v2  v2.43.0
[...]

PR #4185, opened 2025-06-11
[dbforpostgresql] rename UserAssignedIdentity to ServerUserAssignedIdentity

EronWright:
This PR renames `dbforpostgresql.UserAssignedIdentity` to `dbforpostgresql.ServerUserAssignedIdentity` ([spec](https://github.com/Azure/azure-rest-api-specs/blob/88652738431f51137c31860eb70a17cbce12a6d4/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2024-08-01/FlexibleServers.json#L1019-L1054)), because there are multiple definitions of `UserAssignedIdentity` in the REST API specs, and renaming this variant is the least disruptive approach.

#4186 will then re-add a type named `dbforpostgresql.UserAssignedIdentity` that is based on the [common-types schema](https://github.com/Azure/azure-rest-api-specs/blob/88652738431f51137c31860eb70a17cbce12a6d4/specification/common-types/resource-management/v5/managedidentity.json#L95-L112).

Note how the `UserAssignedIdentityResponse` type is changing shape, losing some properties.  These properties were present due to "merging" logic; essentially the dbforpostgresql variant was being merged with the common variant.  Now they've been separated.

Part of: #4176

See also: https://github.com/pulumi/pulumi-azure-native/issues/3324


EronWright:

- #4185 ◀
    - #4186
        - #4187

EronWright:
Closing this because it was needed only in support of https://github.com/pulumi/pulumi-azure-native/pull/4186.

Issue #2645, opened 2023-08-10
Missing examples for VirtualMachineImageTemplate using v2 user-assigned identities


thomas11:
In v2 of the provider, user-assigned identities were overhauled and got a new API. In the process, the generated examples seem to have been lost, like on [VirtualMachineImageTemplate](https://www.pulumi.com/registry/packages/azure-native/api-docs/virtualmachineimages/virtualmachineimagetemplate/#azure-native-virtualmachineimages-virtualmachineimagetemplate).

See also #1258

Issue #1495, opened 2022-02-09
UserAssignedIdentities for AKS is almost impossible fill out in Go


TechnicallyJoe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

I'm attempting to create a ManagedIdentity and assign it to AKS in Go.
The problem is getting the Identity ID and converting that to a string to be used by "&containerservice.ManagedClusterIdentityArgs"

I've found several issues describing this exact problem, but for other languages. I'm not sure if i'm doing something wrong but I believe that the Go type system is preventing me from using them.
See #812 

There's also a lively discussion on this in the Slack Golang channel - https://pulumi-community.slack.com/archives/CCWP5TJ5U/p1644346766552949

As a last attempt i've attempted to "manually" build the Identity using something like this:

```
//In this case "clusterArgs.UserAssignedIdentityId" is a string
Identity: &containerservice.ManagedClusterIdentityArgs{
			Type: containerservice.ResourceIdentityTypeUserAssigned,
			UserAssignedIdentities: pulumi.Map{
				clusterArgs.UserAssignedIdentityId: nil,
			},
		},
```

However i'm left with an error saying
```
 error: Code="MissingIdentityIds" Message="The identity ids must not be null or empty for 'UserAssigned' identity type."
```

I have confirmed by outputting it using fmt.Println that the ID is in fact filled out.

Last side note: The example here: https://www.pulumi.com/registry/packages/azure-native/api-docs/containerservice/managedcluster/#createupdate-managed-cluster-with-enableahub
Specifies to use `pulumi.AnyMap` however this does not exist `github.com/pulumi/pulumi/sdk/v3 v3.23.2`

### Steps to reproduce

See #812  - Its been described a few times already.

TechnicallyJoe:
So based on this issue on a different repo - https://github.com/pulumi/pulumi-azure-nextgen/issues/141

And trying what feels like half a million different things, it turns out if you pass
`pulumi.ToStringMap(map[string]string{}),` instead of `nil` - it works. Like:

`clusterArgs.UserAssignedIdentityId: pulumi.ToStringMap(map[string]string{}),`

This can't seriously be the intended way of doing this.

Still can't pass in Identity.ID() because I can't turn a pulumi.ID into a string, nor a pulumi.StringOutput into a string. This does seemingly only apply to the Go SDK though.

[...]

Issue #1001, opened 2021-07-19
[question] What's the correct way of providing a managed identity to an Azure VMSS created by a ManagedCluster?


markphillips100:
Apologies if this has been asked before but I'm not seeing specific examples of what I'm trying to do.  I'll explain.

Some background first. I'm new to azure-native having used the azure wrapper a year or so ago with time off till now.  So in the past I would have created an Azure AKS cluster and installed aad-pod-identity to provide managed identity to pods.  At the time aad-pod-identity was installed in what is now called "standard mode".  This required creation of user managed identities that would be assigned to VMSS by a MIC controller.  And in my particular use case I would create a user managed identity per application that I deploy to the cluster.  This identity would be given whatever other AAD roles are required for it to access other Azure resources the app is dependent on.  In short, the application can only access what is should.

Now the preference is to use "managed" mode with aad-pod-identity, and as I understand things this installation is handled automatically by  enabling the feature when creating the ManagedCluster.

My question comes in here as to how application-specific user managed identities should be created and assigned to the cluster?  I understand there's an azure cli command for setting against a vmss but as azure-native is supposed to be close to rest api as possible I figured there would be something I could use here.

I did see [this](https://github.com/pulumi/pulumi-azure-native/issues/812) issue which describes setting the "identity" property but I'm assuming that's specifying the property on the ManagedCluster resource as part of its creation.  This made me think that there's only meant to be one user managed identity for the whole cluster?  If so, that would mean all apps use the same identity which feels wrong.

In my case, I could have several different application deployments each with their own identity, so how to manage those identity assignments against the cluster/vmss without effecting any other deployment's identity?

mikhailshilkov:
Do you have an example of what you are trying to achieve expressed in ARM Templates or Azure CLI?

markphillips100:
I've only seen an example of this done using Azure Cli  as described [here](https://azure.github.io/aad-pod-identity/docs/configure/pod_identity_in_managed_mode/#values) on the aad-pod-identity docs, although I've not tried this myself yet. 

This is the specific command:
```
az vmss identity assign -g <VM resource group name> -n <VMSS name> --identities <resource ID of managed identity>
```

I can make the VMSS name and VM resource group name outputs of my cluster deployment, assuming there's a way of obtaining the name of the cluster's resource group using @pulumi/azure-native?

Assuming the above is possible, is it possible to achieve the cli command's behaviour purely with @pulumi/azure-native?

[...]

PR #4186, opened 2025-06-11
correctly generate userAssignedIdentities property

EronWright:
This PR makes a breaking change to the `userAssignedIdentities` properties of many resources to use the correct type,
and is the main fix for #4176. The change resembles:

```diff
- userAssignedIdentities?: pulumi.Input<pulumi.Input<string>[]>;
+ userAssignedIdentities?: pulumi.Input<{[key: string]: pulumi.Input<inputs.web.UserAssignedIdentityArgs>}>;
```

To explain, the `userAssignedIdentities` property as seen on many resources has a `UserAssignedIdentities` type (see [spec](https://github.com/Azure/azure-rest-api-specs/blob/88652738431f51137c31860eb70a17cbce12a6d4/specification/common-types/resource-management/v5/managedidentity.json#L86-L94)) that resembles:

```json
    "UserAssignedIdentities": {
      "type": "object",
      "title": "User-Assigned Identities",
      "description": "The set of user assigned identities associated with the resource. The userAssignedIdentities dictionary keys will be ARM resource ids in the form: '/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}. The dictionary values can be empty objects ({}) in requests.",
      "additionalProperties": {
        "$ref": "#/definitions/UserAssignedIdentity",
        "x-nullable": true
      }
    }
```

Notice the use of `additionalProperties` here to create a keyed structure. For example, to assign two identities to a particular resource, the input would be:
```yaml
userAssignedIdentities:
  "arm-id-foo": {}
  "arm-id-bar": {}
```

In the output value, the identities become "filled in" by the system:
```yaml
userAssignedIdentities:
  "arm-id-foo": {
     "principal-id": "...",
     "client-id": "..."
  }
  "arm-id-bar": {
     ...
  }
```

The schema generator was failing to generate the correct property type for two reasons in combination:
1. The type generator suppresses the generation of "empty" input types (see [code](https://github.com/pulumi/pulumi-azure-native/blob/7d74c23973a253ddb3dd430c1971a78d13d495f6/provider/pkg/gen/types.go#L91-L95)). The rationale may be that such types are noisy when they exist only to contain output values.  In this case, the `UserAssignedIdentity` type.
2. The property generator for properties that use "additionalProperties" doesn't expect (1) and so it applies the wrong type.

The solution is to not suppress the generation of the empty type in this case.  A possible follow-up would be to not suppress generation in any other case either.

Part of: https://github.com/pulumi/pulumi-azure-native/issues/4176


EronWright:

- #4185
    - #4186 ◀
        - #4187

EronWright:
It did not work before, the workaround was to bypass the SDK's type system by whatever means.  https://github.com/pulumi/pulumi-azure-native/issues/4176 has an workaround for typescript:
```ts
identity: {
      type: "UserAssigned",
      userAssignedIdentities: output(funtionManagedIdentity.id).apply((id) => ({
        [id]: {},
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
      })) as any,
    },
```

EronWright:
Related: https://github.com/pulumi/pulumi-azure-native/issues/1001 https://github.com/pulumi/pulumi-azure-native/issues/812 https://github.com/pulumi/pulumi-azure-native/issues/1495 https://github.com/pulumi/pulumi-azure-native/issues/2645

EronWright:
Closing this as by-design per https://github.com/pulumi/pulumi-azure-native/pull/1833/.

PR #4187, opened 2025-06-11
[dbformysql] schema fix for userAssignedIdentity property

EronWright:
This PR makes a fix to the API for setting user-assigned identities on the ["azure-native.dbformysql.Server"](https://www.pulumi.com/registry/packages/azure-native/api-docs/dbformysql/server/#azure-native-dbformysql-server)  resource. The `userAssignedIdentities` property within the `identity` block will now have the correct type.

There's an underlying problem with the Azure REST API specs; the property is defined to have _an array of_ UserAssignedIdentity objects (see [spec](https://github.com/Azure/azure-rest-api-specs/blob/88652738431f51137c31860eb70a17cbce12a6d4/specification/mysql/resource-manager/Microsoft.DBforMySQL/FlexibleServers/stable/2023-12-30/FlexibleServers.json#L1387-L1397)). It should not be an array.

```json
        "userAssignedIdentities": {
          "type": "object",
          "additionalProperties": {
            "type": "object",
            "description": "",
            "items": {
              "$ref": "#/definitions/UserAssignedIdentity"
            }
          },
          "description": "Metadata of user assigned identity."
        }
   
```

The solution is to add a special-case such that the value is not treated as an array.  

The input continues to have an array-like syntax (as per https://github.com/pulumi/pulumi-azure-native/pull/1833):
```diff
userAssignedIdentities?: pulumi.Input<pulumi.Input<string>[]>;
```

The outputs are changing from an array to a non-array:
```diff
- userAssignedIdentities?: {[key: string]: outputs.dbformysql.UserAssignedIdentityResponse[]};
+ userAssignedIdentities?: {[key: string]: outputs.dbformysql.UserAssignedIdentityResponse};
```

The results seem good, for example as a stack output:
```
Current stack outputs (1):
    OUTPUT    VALUE
    identity  {"principalId":"00000000-0000-0000-0000-000000000000","tenantId":"00000000-0000-0000-0000-000000000000","type":"UserAssigned","userAssignedIdentities":{"/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourcegroups/rg5c50fa27/providers/Microsoft.ManagedIdentity/userAssignedIdentities/server-mi8894c455":{"principalId":"074ad052-68b5-467d-9cfc-f1e848e901a1"}}}
```

### DIFF
--- a/provider/pkg/azure/cloud/metadata/client_test.go
+++ b/provider/pkg/azure/cloud/metadata/client_test.go
@@ -128,7 +128,7 @@ func metadataStubServer(ctx context.Context) net.Addr {
 		<-ctx.Done()
 		err := server.Shutdown(ctx)
 		if err != nil {
-			log.Fatalf("failed to gracefully shut down ARM Metadata stub server: %v", err)
+			log.Printf("failed to gracefully shut down ARM Metadata stub server: %v", err)
 		}
 	}()
 


EronWright:

- #4185
    - #4186
        - #4187 ◀

blampe:
The difference from the API spec makes me a little nervous - seems worth covering with an e2e test in case they ever change their behavior to match the spec.

EronWright:
Note that the Java SDK is not checked in, so we don't see a change there.  However, I double-checked that generated SDK looks correct:

```java
package com.pulumi.azurenative.dbformysql.outputs;
@CustomType
public final class MySQLServerIdentityResponse {
    public Map<String,UserAssignedIdentityResponse> userAssignedIdentities() {
        return this.userAssignedIdentities == null ? Map.of() : this.userAssignedIdentities;
    }
}
```

blampe (approved):

PR #4199, opened 2025-06-27
fix for makefile (install_provider)

EronWright:
Taking a dependency on `.make/provider_prebuild` ensures that the schema file is copied from bin/ to cmd/, where it becomes embedded into the binary.  I was finding that `make install_provider` would _build_ but not _use_ the schema, due to lack of this dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -436,6 +436,6 @@ export FAKE_MODULE
 	; fi
 	@touch $@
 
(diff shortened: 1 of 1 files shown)


EronWright:

- #4199 ◀

blampe (approved):

Issue #3962, opened 2025-02-18
Using `TagAtScope` to add a tag to a resource where one already exists fails


pierskarsenbarg:
### What happened?

If a resource already has a tag (that's either been set using Pulumi or in the console) and you use the `TagAtScope` [resource](https://www.pulumi.com/registry/packages/azure-native/api-docs/resources/tagatscope/) it fails with the following error (I've removed the subscription ID):

```
error: cannot create already existing subresource '/%2Fsubscriptions%2F{subscription_id}%2FresourceGroups%2Fresource_groupc9bbe73a/providers/Microsoft.Resources/tags/default'
```

### Example

Using the following code works if there are no tags on the resource group

```python
from pulumi_azure_native import resources

resource_group = resources.ResourceGroup("resource_group")

tag_at_scope = resources.TagAtScope(
    "tag", properties={"tags": {"owner": "piers"}}, scope=resource_group.id
)
```

But then if you add the following:

```python
tag_at_scope2 = resources.TagAtScope(
    "tag2", properties={"tags": {"owner2": "piers"}}, scope=resource_group.id
)
```

This is fine generally because you can just add multiple tags in one place, but if you have a resource that you don't necessarily control with Pulumi but you want to add a tag, you can't

### Output of `pulumi about`

```
CLI
Version      3.147.1-dev.0
Go Version   go1.23.6
Go Compiler  gc

Plugins
KIND      NAME          VERSION
resource  azure-native  2.88.0
language  python        3.147.1-dev.0

Host
OS       darwin
Version  15.1.1
Arch     arm64

This project is written in python: executable='/Users/piers/go/src/github.com/pierskarsenbarg/scratchpad/azure-tagatscope-python/venv/bin/python' version='3.13.2'

Current Stack: pierskarsenbarg/azure-tagatscope-python/dev

TYPE                                  URN
pulumi:pulumi:Stack                   urn:pulumi:dev::azure-tagatscope-python::pulumi:pulumi:Stack::azure-tagatscope-python-dev
pulumi:providers:azure-native         urn:pulumi:dev::azure-tagatscope-python::pulumi:providers:azure-native::default_2_88_0
azure-native:resources:ResourceGroup  urn:pulumi:dev::azure-tagatscope-python::azure-native:resources:ResourceGroup::resource_group
azure-native:resources:TagAtScope     urn:pulumi:dev::azure-tagatscope-python::azure-native:resources:TagAtScope::tag


[...]

PR #4200, opened 2025-06-27
fix for TagAtScope for empty tags 

EronWright:
This PR addresses a bug of the `TagAtScope` resource, causing the following error message when the target resource (scope) has no tags:

```
error: cannot create already existing subresource '/%2Fsubscriptions%2F{subscription_id}%2FresourceGroups%2Fresource_groupc9bbe73a/providers/Microsoft.Resources/tags/default'
```

The message indicates that the target resource (scope) cannot be adopted by the TagAtScope resource, which would otherwise overwrite the tags of that resource.  When there are existing tags, it is appropriate to get an error message, and the remedy is to use the `import` tag.  This PR does not change this behavior, it fixes an edge case where the target resource has no tags.

The simplest repro is to create an RG with an empty set of tags (as opposed to nil), and a TagAtScope for that same resource. 

```yaml

name: azure-rm-yaml
runtime: yaml
description: A demonstration of tagging using TagAtScope.
resources:
  resource_group_2:
    type: azure-native:resources:ResourceGroup
    properties:
      # issue-3962 - Set the tags to an empty map to distinguish from a nil map.
      tags: {}
    options:
      ignoreChanges:
        - tags # issue-4196
  tag_2:
    type: azure-native:resources:TagAtScope
    properties:
      properties:
        tags:
          owner: tag_2
      scope: ${resource_group_2.id}
outputs:
  rg_2_name: ${resource_group_2.name}
  rg_2_tags: ${resource_group_2.tags}
```

Related to #3962

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -436,6 +436,6 @@ export FAKE_MODULE
 	; fi
 	@touch $@
 
-.make/install_provider: .make/provider_mod_download provider/cmd/$(PROVIDER)/* $(PROVIDER_PKG)
+.make/install_provider: .make/provider_mod_download provider/cmd/$(PROVIDER)/* .make/provider_prebuild $(PROVIDER_PKG)
 	cd provider && go install $(VERSION_FLAGS) $(PROJECT)/v2/provider/cmd/$(PROVIDER)
 	@touch $@


EronWright:

- #4200 ◀

provider/pkg/convert/isDefaultResponse.go
```diff
@@ -12,46 +12,55 @@ import (
...
-							return false
-						}
-						defaultValue, has := defaultMap[mk]
-						if !has || defaultValue != mv {
-							return false
-						}
-					}
-				case reflect.Bool:
-					b := value.(bool)
-					if b && defaultBody[key] != value {
```
EronWright, line 43: The comparison on this line causes a panic if the value isn't comparable, such as in the case of sub-maps. The refactored code simply recurses at this point; no other logic changes are intended here.

provider/pkg/convert/isDefaultResponse.go
```diff
@@ -12,46 +12,55 @@ import (
...
-						return false
-					}
-				default:
-					// `*` default body means that we want to accept any value there.
-					// It's used for values that are determined dynamically by Azure API.
-					if defaultBody[key] != value && defaultBody[key] != "*" {
```
EronWright, line 49: The comparison on this line (`defaultBody[key] != value`) causes a panic if the value isn't comparable, such as in the case of sub-maps. The refactored code simply recurses at this point; no other logic changes are intended here.

provider/pkg/convert/isDefaultResponse.go
```diff
@@ -12,46 +12,55 @@ import (
...
+			defaultValue, has := defaultMap[mk]
+			if !has || !isDefaultValue(mv, defaultValue) {
+				return false
+			}
+		}
+	case reflect.Bool:
+		b := value.(bool)
+		if b && defaultValue != value {
+			return false
+		}
```
blampe, line 57: Should this be `return value == defaultValue`?

Does it matter that explicitly setting something to the default value will be functionally equivalent to if that thing had been omitted?

EronWright: I think no, because this code is designed to match well-known subresources that are in an "unset" state and thus overwritable. Otherwise, the subresource is considered to be unmanaged and one must use import to bring it under management.

EronWright:
See also, the original fix that dealt with TagAtScope: https://github.com/pulumi/pulumi-azure-native/pull/1323

I believe that this PR's fix is a superset of #1323 because the logic of 'isDefaultResouce' allows for a partial match.  That is, whether the response has `properties={}` or has `properties.tags={}`, it is considered a default response.

danielrbradley (approved):

PR #4201, opened 2025-06-27
Fix formatting of REST API paths given "x-ms-skip-url-encoding" directive

EronWright:
In the REST API specs, certain path parameters are marked with `x-ms-skip-url-encoding` because their values are sub-paths containing slashes. The provider should honor this directive in order to produce a correct REST API path, but currently it doesn't, for example it produces:
```
/%2Fsubscriptions%1234567890-0000-424b-0000-96babd57a8a1%2FresourceGroups%2Fresource_groupc9bbe73a/providers/Microsoft.Resources/tags/default
```

Such a path does still work inside the Azure SDK (due to some unescaping logic), but improving the correctness of `PrepareAzureRESTIdAndQuery` has benefits for other PRs (https://github.com/pulumi/pulumi-azure-native/pull/4202).  The ideal output is:

```
/subscriptions/1234567890-0000-424b-0000-96babd57a8a1/resourceGroups/resource_groupb04391fe/providers/Microsoft.Resources/tags/default
```

For example, [Tags_CreateOrUpdateAtScope](https://github.com/Azure/azure-rest-api-specs/blob/6ba380cf89dadde10a72781d4e39b51bf1d24cd7/specification/resources/resource-manager/Microsoft.Resources/stable/2025-04-01/resources.json#L1905-L1916) has a `scope` parameter that is marked in this way, and refers to either a resource group or a subscription.

```json
    "/{scope}/providers/Microsoft.Resources/tags/default": {
      "put": {
        "tags": [
          "Tags"
        ],
        "operationId": "Tags_CreateOrUpdateAtScope",
        "summary": "Creates or updates the entire set of tags on a resource or subscription.",
        "description": "This operation allows adding or replacing the entire set of tags on the specified resource or subscription. The specified entity can have a maximum of 50 tags.",
        "parameters": [
          {
            "$ref": "#/parameters/ScopeParameter"
          },
```

```json
  "parameters": {
    "ScopeParameter": {
      "name": "scope",
      "in": "path",
      "required": true,
      "type": "string",
      "description": "The resource scope.",
      "x-ms-parameter-location": "method",
      "x-ms-skip-url-encoding": true
    },
```

Since users expect to set the scope to `resource_group.id`, the value is likely to have a leading slash, so the code seeks to give `path.Join`-like behavior to avoid producing a double-slash.

For reference, see the Azure SDK's handling of such parameters:
- https://github.com/Azure/azure-sdk-for-go/blob/f3cbdffacfe2c536416d3e953a0141111bb07741/sdk/resourcemanager/resources/armresources/tags_client.go#L158

### DIFF
--- a/provider/pkg/provider/crud/crud_test.go
+++ b/provider/pkg/provider/crud/crud_test.go
@@ -99,6 +99,77 @@ func TestPathParamsErrorHandling(t *testing.T) {
 	})
 }
 
+func TestPathParamEncoding(t *testing.T) {
+
+	t.Run("path encoding", func(t *testing.T) {
+		tests := []struct {
+			p1       string
+			expected string
+		}{
+			{
+				p1:       "my-value",
+				expected: "/path/my-value",
+			},
+			{
+				p1:       "my-value with space and /",
+				expected: "/path/my-value%20with%20space%20and%20%2F",
+			},
+		}
+
+		for _, tt := range tests {
+			t.Run(tt.p1, func(t *testing.T) {
+				id, _, err := PrepareAzureRESTIdAndQuery("/path/{p1}",
+					[]resources.AzureAPIParameter{
+						{
+							Name:            "p1",
+							Location:        "path",
+							Value:           &resources.AzureAPIProperty{Type: "string"},
+							SkipUrlEncoding: false,
+						},
+					}, map[string]any{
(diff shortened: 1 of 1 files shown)


EronWright:

- #4201 ◀

blampe (approved):
Nice!

PR #4202, opened 2025-06-27
Canonicalize the path parameters in Check and Diff

EronWright:
This PR seeks to canonicalize the value of path parameters during `Check`, so that the values are equal to what would be produced by an import of that same resource. This is immensely helpful for importing a resource without causing replacement, since most path parameters have replace-on-change semantics.

For example, when you import a `TagAtScope` resource, the `scope` parameter is imported from the resource's ID based on the operation's pattern, which is:

```
"/{scope}/providers/Microsoft.Resources/tags/default"
```

Given an ID of `/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/resource_groupfde2092d/providers/Microsoft.Resources/tags/default`, we get the following state. Observe that the `scope` parameter has no leading slash.

```json
"urn": "urn:pulumi:dev::issue-3962::azure-native:resources:TagAtScope::tag_imported",
"custom": true,
"id": "/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/resource_groupfde2092d/providers/Microsoft.Resources/tags/default",
"type": "azure-native:resources:TagAtScope",
"inputs": {
    "properties": {
        "tags": {
            "owner": "external"
        }
    },
    "scope": "subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/resource_groupfde2092d"
},
```

Meanwhile, when you create a `TagAtScope` resource from scratch, you're likely to use `scope=resource_group.id` and hence give the scope a slash.

```py
resource_group = resources.ResourceGroup("resource_group"})
tag_at_scope = resources.TagAtScope(
    "tag_at_scope", properties={"tags": {"owner": "tag_at_scope"}}, scope=resource_group.id
)
```

```json
"urn": "urn:pulumi:dev::issue-3962::azure-native:resources:TagAtScope::tag",
"custom": true,
"id": "/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/resource_group31d34185/providers/Microsoft.Resources/tags/default",
"type": "azure-native:resources:TagAtScope",
"inputs": {
    "azureApiVersion": "2024-03-01",
    "properties": {
        "tags": {
             "owner": "tag_at_scope"
        }
    },
    "scope": "/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/resource_group31d34185"
},
```

This leads to problems during import because scope changes cause replacement, and so you can't simply use the expression `resource_group.id`, you need to trim it yourself as seen in [this example](https://github.com/pulumi/pulumi-azure-native/issues/3962#issuecomment-2674146080). The goal of this PR is to make `scope` have the trimmed value as seen above, by canonicalizing it during Check.

For migration purposes, we need to canonicalize the old inputs too in Diff, to avoid producing a spurious replacement. Of course, one still wants to detect a true change to the scope, so ignoring it isn't an option.

Related to #3962

### DIFF
--- a/provider/pkg/provider/crud/crud.go
+++ b/provider/pkg/provider/crud/crud.go
@@ -24,6 +24,7 @@ import (
 // It operates in the context of a specific kind of Azure resource of type resources.AzureAPIResource.
 type AzureRESTConverter interface {
 	// PrepareAzureRESTIdAndQuery prepares the ID and query parameters for an Azure REST API request.
+	// Unknowns are represented as plugin.UnknownStringValue in the returned values.
 	PrepareAzureRESTIdAndQuery(inputs resource.PropertyMap) (string, map[string]any, error)
 
 	// PrepareAzureRESTBody prepares the body of an Azure REST API request.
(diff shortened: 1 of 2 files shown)


EronWright:

- #4202 ◀

EronWright:
This is in draft because it needs test code, e.g. an upgrade test, and to verify how empty values are handled.

provider/pkg/provider/crud/crud.go
```diff
@@ -24,6 +24,7 @@ import (
 // It operates in the context of a specific kind of Azure resource of type resources.AzureAPIResource.
 type AzureRESTConverter interface {
 	// PrepareAzureRESTIdAndQuery prepares the ID and query parameters for an Azure REST API request.
+	// Unknowns are represented as plugin.UnknownStringValue in the returned values.
```
EronWright, line 27: This method previously did not handle unknowns because it was called only in non-preview mode.  Now it is called during Check, even in preview, and so it returns sentinel values for the unknowns.

PR #4207, opened 2025-07-01
SDK compatibility with latest specs

blampe:
These were all the changes necessary I needed to unblock local SDK generation from the latest specs.

* Omit examples from spec discovery.
* Handle a panic due to missing max length validation.

### DIFF
--- a/versions/v3-config.yaml
+++ b/versions/v3-config.yaml
@@ -594,18 +594,8 @@ ProfessionalService:
   notes: Only single preview version
 ProviderHub:
   expectTracking: preview
-  exclusions:
-    AuthorizedApplication: "2024-09-01"
-    CustomRollout: "2024-09-01"
-    DefaultRollout: "2024-09-01"
-    NewRegionFrontloadRelease: "2024-09-01"
-    NotificationRegistration: "2024-09-01"
-    OperationByProviderRegistration: "2024-09-01"
-    ProviderRegistration: "2024-09-01"
-    ResourceTypeRegistration: "2024-09-01"
-    Sku: "2024-09-01"
   notes: |
-    2024-09-01 stable is excluded due to incompatibilities with ResourceAccessRoles.
+    Can't find any docs
 Purview:
 Quantum:
   expectTracking: preview


blampe:

- #4207 ◀
    - #4208

danielrbradley (commented):
The code and config change all look good to me.

Should we be expecting to see some changes to the schema & SDKs?

blampe:
> The code and config change all look good to me.
> 
> Should we be expecting to see some changes to the schema & SDKs?

Yes that's broken out into #4208 to make the review a little easier.

danielrbradley:
> Exclude latest API from providerhub due to incompatibilities. The previous API had items: {} but the latest API makes this a ResourceAccessRole type. This leads to an error refs do not match: pulumi.json#/Any vs #/types/azure-native:providerhub:ResourceAccessRoleResponse which I'm not sure how to resolve, so I'm excluding the latest providerhub API for now.

This happens during the "type merging" phase where we're trying to resolve two, possibly different types with the same name into one type which is usable with both resources - most likely from different API versions.

1. Excluding one of the versions is generally fine. 
2. We could also add a condition at the beginning, given "any" is the most permissive type, then we could say that if either allows "any", then just make the type "any". This looses some of the nice typing from one side, but technically allows any valid value to be passed in.
3. Alternatively, we can add specific overrides where we want here before calling the merge function: https://github.com/pulumi/pulumi-azure-native/blob/c1920f4202f8a71a3620b58dde8ecc732a7e0469/provider/pkg/gen/types.go#L108-L133

blampe:
> 2. We could also add a condition at the beginning, given "any" is the most permissive type, then we could say that if either allows "any", then just make the type "any". This looses some of the nice typing from one side, but technically allows any valid value to be passed in.

Could we not go the other way and always prefer the concrete type over any?

> 3. Alternatively, we can add specific overrides where we want here before calling the merge function: https://github.com/pulumi/pulumi-azure-native/blob/c1920f4202f8a71a3620b58dde8ecc732a7e0469/provider/pkg/gen/types.go#L108-L133

@danielrbradley I'll be honest I tried doing roughly this but I have no idea how to override the relevant bits. The entire error chain is very nested:

> panic: failed to generate 'azure-native:providerhub:DefaultRollout': response type: failed to generate property "properties": failed to generate type spec for property "properties": failed to generate allOf properties 0: failed to generate property "specification": failed to generate type spec for property "specification": failed to generate allOf properties 0: failed to generate property "providerRegistration": failed to generate type spec for property "providerRegistration": failed to generate allOf properties 0: failed to generate property "properties": failed to generate type spec for property "properties": failed to generate allOf properties 0: failed to generate allOf properties 0: failed to generate property "management": failed to generate type spec for property "management": incompatible type "azure-native:providerhub:ResourceProviderManifestPropertiesResponseManagement" for resource "DefaultRollout" ("azure-native:providerhub:DefaultRollout"): refs do not match: pulumi.json#/Any vs #/types/azure-native:providerhub:ResourceAccessRoleResponse

I made progress with this override but it just led to more errors:
```
switch tok {
case
        "azure-native:providerhub:ResourceProviderManifestPropertiesManagement",
        "azure-native:providerhub:ResourceTypeRegistrationPropertiesManagement":
        props.specs["resourceAccessRoles"].TypeSpec.Items.Ref = "#/types/azure-native:providerhub:ResourceAccessRole"
case
        "azure-native:providerhub:ResourceProviderManifestPropertiesResponseManagement",
        "azure-native:providerhub:ResourceTypeRegistrationPropertiesResponseManagement":
        props.specs["resourceAccessRoles"].TypeSpec.Items.Ref = "#/types/azure-native:providerhub:ResourceAccessRoleResponse"
}
```

> panic: failed to generate 'azure-native:providerhub:DefaultRollout': response type: failed to generate property "properties": failed to generate type spec for property "properties": failed to generate allOf properties 0: failed to generate property "specification": failed to generate type spec for property "specification": failed to generate allOf properties 0: failed to generate property "resourceTypeRegistrations": failed to generate type spec for property "resourceTypeRegistrations": failed to generate type spec for array element: failed to generate type spec for property "resourceTypeRegistrations": incompatible type "azure-native:providerhub:ResourceTypeRegistrationResponse" for resource "DefaultRollout" ("azure-native:providerhub:DefaultRollout"): refs do not match: #/types/azure-native:providerhub:ResourceTypeRegistrationResponseProperties vs #/types/azure-native:providerhub:ResourceTypeRegistrationPropertiesResponse

Hence why I opted to just ignore this API.

danielrbradley:
What about the following:

```diff
diff --git a/provider/pkg/gen/merging.go b/provider/pkg/gen/merging.go
--- a/provider/pkg/gen/merging.go
+++ b/provider/pkg/gen/merging.go
@@ -93,2 +92,2 @@
-		// Refs don't match - fail with error.
-		return nil, errors.Errorf("refs do not match: %s vs %s", t1.Ref, t2.Ref)
+		// Refs don't match provide a union of the two types.
+		return unionOneOfTypes(t1, t2)

```

danielrbradley:
A bit more playing locally with this shows that the two versions are more incompatible and the type names have also changed. Can we just upgrade ProviderHub to 2024-09-01 as our new v3 strategy is to allow minor breaking changes intra-version as users can always generate custom API version SDKs.

danielrbradley:
Looking at the `reports/pending.json` diff, there's also a lot of other modules which could be updated to a newer version, but perhaps we address each of those in separate PRs.

danielrbradley (approved):

PR #4281, opened 2025-08-24
Add new SKU names for RedisEnterprise

smkhalsa:
Update the sku list based on 

https://learn.microsoft.com/en-us/cli/azure/redisenterprise?view=azure-cli-latest#az-redisenterprise-create


smkhalsa:
I realized the new version of the sdk already has these.

PR #4284, opened 2025-08-25
Remove legacy auth code

### LINKED ISSUE #4047: Remove legacy auth code
Authentication was updated to use azidentity in release v2.69 in October 2024. There are no open issues that require users to fall back to the previous auth stack. Therefore, we should remove the legacy auth which would be a significant simplification of the provider code.

EronWright:
This PR removes the autorest library and the ability to enable its use with the `PULUMI_ENABLE_AZCORE_BACKEND` environment variable.

Closes #4047 

_Holding in draft until after the next release (v3.8), specifically after https://github.com/pulumi/pulumi-azure-native/commit/6f231b5f9d4e5ed2da3f249da092cb36a253a829._

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -239,6 +239,14 @@ func (k *azureNativeProvider) Configure(ctx context.Context,
 		return nil, errors.New("Azure Native provider requires Pulumi CLI v3.74.0 or later")
 	}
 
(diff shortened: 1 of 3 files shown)


EronWright:
Updated to throw an error (not just a warning) if PULUMI_ENABLE_AZCORE_BACKEND=false, since one may otherwise have unexpected behavior.

PR #4286, opened 2025-08-26
auth: support for Custom Environments and Disabling Instance Discovery

tombuildsstuff:
This PR includes support for both disabling Instance Discovery (which checks that the Tenant is known to Microsoft before sending credentials over the wire) and Custom Environments (using custom endpoints) when using the AzCore backend.

### DIFF
--- a/provider/pkg/gen/schema.go
+++ b/provider/pkg/gen/schema.go
@@ -248,6 +248,23 @@ func PulumiSchema(rootDir string, modules openapi.AzureModules, versioning Versi
 					Description: "The URL to initiate the `oidcRequestToken` OIDC token exchange.",
 				},
 
+				"disableInstanceDiscovery": {
+					TypeSpec:    pschema.TypeSpec{Type: "boolean"},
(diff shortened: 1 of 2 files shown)


EronWright:
@tombuildsstuff thanks for sharing! The timing is good, given https://github.com/pulumi/pulumi-azure-native/pull/4242 and https://github.com/pulumi/pulumi-azure-native/pull/4174. 

If you don't mind, I'll push some edits to this PR to help get it merged.

EronWright:
@tombuildsstuff I opened a PR based on this one, sorry for the extra hurdle.  As you will see, we need to solve for a few additional endpoints/suffixes, and so I took some effort to wrap the SDK's `cloud.Configuration` structure. 
https://github.com/pulumi/pulumi-azure-native/pull/4288

Would you mind reviewing this and giving it a try?  Thanks!

tombuildsstuff:
@EronWright that PR seems to work fine, however I've left a comment inline about the configuration - as you're requiring 6 different configuration values already, it'd probably be worth sourcing it from the environment instead

tombuildsstuff:
@EronWright worth adding there's a related PR for the other provider too: https://github.com/pulumi/pulumi-azure/pull/3243

PR #4289, opened 2025-08-27
Skip OIDC test if OIDC_ARM_CLIENT_ID is not set

EronWright:
The "master" workflow lacks an OIDC identity that is needed to run OIDC tests, but such a test was added in https://github.com/pulumi/pulumi-azure-native/pull/4242. This PR removes the "fail-fast" behavior for the OIDC test, so that it skips on the master workflow.

The limitation is documented here:
https://github.com/pulumi/pulumi-azure-native/blob/7ed14fcdfeaff45ea8295cb693b8a36d0494fd74/.github/workflows/master.yml#L35-L38

Closes https://github.com/pulumi/pulumi-azure-native/issues/4283

### DIFF
--- a/provider/pkg/provider/provider_e2e_test.go
+++ b/provider/pkg/provider/provider_e2e_test.go
@@ -195,9 +195,6 @@ func TestAzidentity(t *testing.T) {
 	t.Run("OIDC", func(t *testing.T) {
 		oidcClientId := os.Getenv("OIDC_ARM_CLIENT_ID")
 		if oidcClientId == "" {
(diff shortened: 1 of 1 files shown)


EronWright:
The OIDC tests use the value of `OIDC_ARM_CLIENT_ID` as a parameter to configure the provider's OIDC authentication scheme. Apparently, the master workflow doesn't support OIDC due to a detail of identity federation, so it blanks the `OIDC_ARM_CLIENT_ID` to force the test to be skipped.  The acceptance-test workflow is more capable and does run the test.

PR #4356, opened 2025-10-15
Extract canonical ID when importing

danielrbradley:
Prefer to read the canonical ID from the response if inputs are nil.


EronWright:
Possibly related: https://github.com/pulumi/pulumi-azure-native/pull/4202

Issue #4388, opened 2025-10-30
Cannot create containerservice:ManagedCluster with preview API and custom autoscalerProfile values


etiennetremel:
### What happened?

When creating a `containerservice.ManagedCluster` using the azure-native v3 preview APIs the fields under `autoscalerProfile` are not recognized by the Azure API.

> Invalid request body. Converting request body to a managed cluster encountered error: json: unknown field \"scaleDownUtilizationThreshold\" Check that the field is in the right location, is spelled correctly, and is supported in the API version. Visit aka.ms/aks/apiversions for more API version information..

### Example

Install the Azure 2024-10-02-preview API:

```
pulumi package add azure-native containerservice 2024-10-02-preview
```

Then a minimal managed cluster configuration:

```typescript
import * as containerservice from "@pulumi/azure-native_containerservice_v20241002preview/containerservice";

new containerservice.ManagedCluster("my-cluster", {
	// ...

	autoScalerProfile: {
		expander: "least-waste",
		scaleDownUtilizationThreshold: "0.75",
		skipNodesWithSystemPods: "true",
	},
});
```

You will get the following error:

```
  azure-native_containerservice_v20241002preview:containerservice:ManagedCluster (my-cluster):
    error: Status=400 Code="UnmarshalError" Message="{
      "code": "UnmarshalError",
      "details": null,
      "message": "Invalid request body. Converting request body to a managed cluster encountered error: json: unknown field \"scaleDownUtilizationThreshold\" Check that the field is in the right location, is spelled correctly, and is supported in the API version. Visit aka.ms/aks/apiversions for more API version information..",
      "subcode": ""
     }"
```

### Output of `pulumi about`

```
CLI
Version      3.204.0
Go Version   go1.25.3
Go Compiler  gc

Plugins
KIND      NAME          VERSION
resource  azure-native  3.8.0
resource  azure-native  3.8.0
resource  azure-native  3.8.0
language  nodejs        3.204.0

Host
OS       darwin
Version  15.6.1
Arch     arm64

This project is written in nodejs: executable='/Users/me/.local/share/mise/installs/node/23.11.0/bin/node' version='v23.11.0'

// ...

Dependencies:
NAME                                                    VERSION
typescript                                              5.9.3
@pulumi/azure-native_containerservice_v20241002preview  3.8.0
@pulumi/azure-native                                    3.8.0
@pulumi/pulumi                                          3.205.0
@types/node                                             18.19.130
```

### Additional context

[...]

PR #4391, opened 2025-10-31
Fix regression about field case conversion in ManagedCluster.autoScalerProfile

etiennetremel:
This is an attempt to fix a regression in the ManagedCluster resource.

For `ManagedCluster.autoScalerProfile`, the Azure API expect keys to be formatted in [kebab-case](https://www.pulumi.com/registry/packages/azure-native/api-docs/containerservice/managedcluster/) instead of [camelCase](https://www.pulumi.com/registry/packages/azure-native/api-docs/containerservice/managedcluster/). Refer to the issue marked below to see the steps to reproduce this failure.

This only target fields nested under `ManagedCluster.autoScalerProfile`, the rest is fine.
Obviously it's concerning that Azure doesn't agree on a convention so I suppose this is an exception?

I'm still getting familiar with the codebase, so please correct me if this isn't the right place for exceptions. It feels like a workaround, and I'm wondering if there is a more appropriate place to manage these per-resource exceptions.

Also, here I'm using a mapping but maybe we could just do a simple conversion instead, any keys under `autoScalerProfile` will be converted to kebab-case, what's your opinion on this?

Fix: https://github.com/pulumi/pulumi-azure-native/issues/4388

### DIFF
--- a/provider/pkg/convert/autoScalerProfile_test.go
+++ b/provider/pkg/convert/autoScalerProfile_test.go
@@ -1,0 +1,98 @@
+package convert
+
+import (
+	"testing"
+
+	"github.com/pulumi/pulumi-azure-native/v2/provider/pkg/resources"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestRewriteAutoScalerProfileKeys verifies outbound mapping camelCase -> kebab-case.
+func TestRewriteAutoScalerProfileKeys(t *testing.T) {
+	in := map[string]interface{}{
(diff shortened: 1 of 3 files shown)


EronWright:
Thanks @etiennetremel, taking a look.

EronWright:
Thanks @etiennetremel for the issue report and the suggested fix!  Your fix would be fine, but it turns out it can be fixed in a more general way: https://github.com/pulumi/pulumi-azure-native/pull/4455

PR #4414, opened 2025-11-14
Model dependencies through mise

pose:
_A journey of a thousand miles begins with a single step_. Let's make this PR that first step 🥂

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,0 +1,11 @@
+[tools]
+
+go = '1.24.7'
+pulumi = '3.207'
+nodejs = '20'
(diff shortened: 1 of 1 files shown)


EronWright (commented):
Would like to compare with:
https://gist.github.com/EronWright/9ccb3949be97294e71c6c8917e57e4ea

I see that dotnet goes 8->9, why?

guineveresaenger (commented):
Yay good start! Let's roll CI and local version handling out at the same time maybe?

blampe (approved):

Issue #3720, opened 2024-11-20
Retry on TooManyRequests


danielrbradley:
Example error that's not retried:
```
    --------------------------------------------------------------------------------
    RESPONSE 429: 429 Too Many Requests
    ERROR CODE: RetryableError
    --------------------------------------------------------------------------------
    {
      "error": {
        "code": "RetryableError",
        "message": "A retryable error occurred.",
        "details": [
          {
            "code": "RetryableErrorDueToAnotherOperation",
            "message": "Operation PutSubnetOperation (xxxxx) is updating resource /subscriptions/xxxxxx/resourceGroups/resourceGroup123456789/providers/Microsoft.Network/virtualNetworks/virtualNetworke12345679. The call can be retried in 14 seconds."
          }
        ]
      }
    }
    --------------------------------------------------------------------------------
```

Ideally waiting the right amount of time before retrying, though this might be a little tricky to parse out of the message.

vogonistic:
I get 429 Too Many Requests on all deployments now. Both destroying App Services and creating namespaces in Azure ServiceBus. It makes all automated creation and destruction of environments impossible for us.

I don't keep logs, but it seems to have started sometime in November 2024 for me and persists.

dragonlipz:
I've been hitting this when creating resources against Azure from a list of names.
So when webAppNames is brought in as a `List<string>` this all works great, except, it often overloads the `AppServicePlan` object getting 429 errors because the resources is locked by another operation. In this case a different web app create or delete depending on what's going on.

I've also tried doing a `foreach` loop with a sleep between iterations, but that has ZERO effect, other than making each instance longer. It still behaves as though it's running in a `parallel.foreach`.

Essentially, it appears the only throttling that functions is doing `pulumi up/down --parallel #`, no other flow control can be exerted. (other thanDependsOn and Parent) that I can find.

```cs

   var svc = new AppServicePlan {...}

    var webapps = webAppNames.ToDictionary(name => name, name => new LinuxDockerWebAppComponent(name,
      new LinuxDockerWebAppArgs
      {
        ResourceGroup = resourceGroup.Name,
        AppServicePlanId = svc.LinuxAppServicePlanId,
        SubscriptionId = SubscriptionId,
        AppConfig = appConfigConnectionString,
        RegistryRg = registry_rg,
        RegistryName = registry,
      })
    );
```

[...]

PR #4458, opened 2025-12-11
Add serialization for Web App resources

### LINKED ISSUE #4461: Serialize WebApp resource CRUD operations
WebApps can cause a 429 "Cannot acquire exclusive lock" error when too many of them share the same App Service Plan. Our existing attempts at retrying requests are insufficient, because multiple web apps compete for the same lock at the same time, including during retries.

The way recommended by Azure Support is to create/update/delete each WebApp sequentially, rather than in parallel.

Read more here: https://github.com/pulumi/pulumi-azure-native/issues/3720#issuecomment-3610836629

guineveresaenger:
~This pull request is a DRAFT.~ Update: added integration test!

According to https://github.com/pulumi/pulumi-azure-native/issues/3720#issuecomment-3610836629, resources such as Web Apps should be serialized.

This pull request does so for Create/Update, and Delete. In order to be able to parallelize WebApp operations on different App Serive Plans, there's additional logic to make sure that different App Service Plans can run in parallel.

It slows down these operations but it does prevent the "Cannot acquire exclusive lock" errors.

Fixes #4461 .


blampe (approved):
Sounds good, thank you for adding the regression test!

This gets into a weird area where historically we've wanted providers to be stateless, but there are instances like this where a little bit of state is necessary (e.g. p-k keeps track of CRDs it's going to create).

_Technically_ speaking we can still run into this error because there's no guarantee the same provider process will handle each resource (the engine could decide to spin up multiple processes for some reason, or the user could do something weird) but this should fix 99% of the problem.

PR #4500, opened 2026-01-27
Remove dead code using "required explicit resources"

Zaid-Ajaj:
Found dead code that is not used anywhere that reads "required explicit resources" from a text file but the data is then not used anywhere as far as I can tell. This PR removes that code and the text file


guineveresaenger (approved):
🔥 LGTM

PR #4518, opened 2026-02-18
chore: Update API specs to 2026-02-18

JasonWhall:
Updates the API specs submodule reference to latest merge commit from 18-02-2026. This in preperation to support #4475 as a new spec version has been released.

Re-ran SDK generation to update code

I had a bit of a miserable time running integration tests locally as there is a lot of internal knowledge/references I have not untangled yet so may need a steer or triggering actions a few times to work out what may or may not be broken.


Zaid-Ajaj:
API specs updated in #4574 which also updated the MachineLearningServices to a stable version that allows for setting  `managedResourceGroupSettings`

PR #4534, opened 2026-02-23
Sequentially handle network.PrivateEndpoint operations

### LINKED ISSUE #4509: `network.PrivateEndpoint`:  Intermittent HTTP 429 error upon creation
### What happened?

When creating a `network.PrivateEndpoint`, a customer is intermittently receiving an HTTP 429:

The error message says:

```
Status=429 Code="RetryableError" Message="A retryable error occurred."
```

Given that HTTP 429 is "too many requests", it would be awesome if we could automatically retry the (presumable) write that is failing with an exponential backoff. Even better if this is done across the board for all Azure resources.

### Example

n/a - the error is intermittent

### Output of `pulumi about`

n/a - will enter later if I get the info from the customer

### Additional context

_No response_

### Contributing

Zaid-Ajaj:
Fixes #4509

### Description

This PR updates how the azure client handles resource operations for resources of type `network.PrivateEndpoint` and makes them sequential. Uses the the private endpoint name for the mutex key.

Also enables logging for retry events so that we can tell why a retry-able error is returned to the user.


provider/pkg/azure/serailize_test.go
guineveresaenger: typo in filename - let's have it be `serialize_test.go`

PR #4559, opened 2026-03-02
Add oidcTokenFilePath to Provider InputProperties

MaayanLaav:
The `oidcTokenFilePath` config option was already defined in the config variables section but was missing from the Provider InputProperties, preventing it from being used as an explicit provider constructor argument in the SDKs.

It's useful for running different stacks in the same environment.


MaayanLaav:
Hey @Zaid-Ajaj!

I forgot to include the `oidcTokenFilePath` as a parameter in the provider properties, this should address the issue

Zaid-Ajaj:
@MaayanLaav We did need to sdk-generate these inputs so that the languages have access to it. I will do that myself before I release this

PR #4586, opened 2026-03-12
Fix ManagedCluster identityProfile losing objectId and resourceId after v3.10.0

### LINKED ISSUE #4385: containerservice:ManagedCluster 3.10.0 removes properties in state from identityProfile.kubeletidentity
### What happened?

we have the following code `aks.IdentityProfile.Apply(p => p!["kubeletidentity"].ObjectId!)` that is used to set an rbac permission. We noticed after upgrading to 3.10.0 that the ObjectId property was no longer present causing exceptions. We investigated the state and saw it has been removed.

After restoring an earlier state we tried again with 3.10 and noticed the removal again. Reverted to 3.8.0 and this version keeps it intact

### Example

see above

### Output of `pulumi about`

This is on my local machine. We also run with GitHub actions

CLI          
Version      3.204.0
Go Version   go1.25.3
Go Compiler  gc

Plugins
resource  azure-native  3.8.0
resource  azuread       6.6.0
language  dotnet        unknown
resource  mongodbatlas  3.36.1
resource  random        4.18.4

Host     
OS       darwin
Version  26.0.1
Arch     arm64

This project is written in dotnet: executable='/usr/local/share/dotnet/dotnet' version='9.0.304'

Current Stack: dev

### Additional context

_No response_

### Contributing

nelsonfigueroa:
## Summary

Fixes #4385

After upgrading to provider v3.10.0, `identityProfile.kubeletidentity.objectId` and `.resourceId` disappeared from `ManagedCluster` outputs/state, while `clientId` continued to work. Multiple users confirmed the regression and had to pin to v3.8.0.

## Root Cause

The bug is in the codegen's type metadata handling in `provider/pkg/gen/types.go`.

When two different Azure OpenAPI specs define a type that maps to the **same Pulumi type token**, there are two parallel data structures that need to stay in sync:

1. **Schema types** (`pkg.Types`) — used for SDK generation. These were **correctly merged** (union of all properties) via the `mergeTypes` function at line 172.
2. **Metadata types** (`metadata.Types`) — used at **runtime** by `ResponseBodyToSdkOutputs` to determine which properties to extract from Azure API responses. These were **unconditionally overwritten** with whichever definition was processed last.

### The triggering change

v3.10.0 added the `NodeCustomization` resource (API version `2025-08-02-preview`), whose `identityProfile` references **common-types** `UserAssignedIdentity`:
- Properties: `{clientId, principalId}`

ManagedCluster's `identityProfile` references the **containerservice-specific** `UserAssignedIdentity`:
- Properties: `{clientId, objectId, resourceId}`

Both produce the same type token: `azure-native:containerservice:UserAssignedIdentityResponse`.

Since `visitedTypes` is per-generator-instance but `metadata.Types` is shared across generators:
1. Generator A (ManagedCluster) sets metadata = `{clientId, objectId, resourceId}`
2. Generator B (NodeCustomization) **overwrites** metadata = `{clientId, principalId}`

At runtime, `ResponseBodyToSdkOutputs` iterates over metadata properties. Azure returns `objectId` and `resourceId` in the API response, but they're silently dropped because they're no longer in the metadata.

This explains why `clientId` still worked (present in both definitions) while `objectId` and `resourceId` did not (only in the first definition, which got overwritten).

### Why it's a latent bug

The overwrite-instead-of-merge pattern has existed since the metadata was introduced, but it only manifests when:
- Two specs in the same module define the same-named type
- With **different** property sets
- And a module generator processes them in separate passes

This is the first time it caused a user-visible regression because the common-types `UserAssignedIdentity` has a significantly different property set from the containerservice-specific one.

### Why `disambiguateTypeToken` (PR #4496) does not fix this

In v3.13.0, `disambiguateTypeToken` was added to handle token collisions when `mergeTypes` **fails** (incompatible types). It creates a new token (e.g., `FooV1`) so each type gets its own entry.

However, `containerservice:UserAssignedIdentityResponse` types are **compatible** — `mergeTypes` succeeds (confirmed: the committed schema.json has all four properties, and no `V1`-suffixed type exists). When merge succeeds, `disambiguateTypeToken` is never called, `tok` stays the same, and the metadata overwrite on line 189 still occurs.

The two fixes address different branches of the same if/else at `types.go:171-184`. Verified on master (v3.15.0) — the unconditional metadata overwrite is still present at `types.go:189`:

```go
m.metadata.Types[tok] = resources.AzureAPIType{
    Properties:         props.properties,
    RequiredProperties: props.requiredProperties.SortedValues(),
}
```

## Fix

When setting `metadata.Types[tok]`, check if an entry already exists. If so, merge properties (union) instead of overwriting — mirroring the schema merge behavior directly above it. Existing properties are preserved, and new properties from the second definition are added.

```go
// Before (line 189): unconditional overwrite
m.metadata.Types[tok] = resources.AzureAPIType{...}

// After: merge if existing
if existingMeta, has := m.metadata.Types[tok]; has {
    // Union of properties from both definitions
    mergedProps := merge(existingMeta.Properties, props.properties)
    m.metadata.Types[tok] = resources.AzureAPIType{Properties: mergedProps, ...}
} else {
    m.metadata.Types[tok] = resources.AzureAPIType{...}
}
```

After the fix, regardless of processing order:
- metadata = `{clientId, objectId, resourceId, principalId}` (union of both)
- `ResponseBodyToSdkOutputs` sees all properties and `objectId` and `resourceId` are included in output

## Test plan

- [x] Existing unit tests pass (`mergeTypes`, enum tests, etc.)
- [ ] Run `make schema` and verify `metadata-compact.json` contains `objectId` and `resourceId` for `containerservice:UserAssignedIdentityResponse`
- [ ] Run `make provider` and verify ManagedCluster read returns `identityProfile.kubeletidentity.objectId`
- [ ] Verify no schema diff / SDK diff beyond the expected metadata fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/examples_dotnet_test.go
+++ b/examples/examples_dotnet_test.go
@@ -33,6 +33,7 @@ func TestAccSimpleDotnet(t *testing.T) {
 }
 
 func TestAccSql(t *testing.T) {
+	t.Skip("TODO https://github.com/pulumi/pulumi-azure-native/issues/4595")
 	test := getCsharpBaseOptions(t).
 		With(integration.ProgramTestOptions{
 			Dir: filepath.Join(getCwd(t), "cs-sql"),


Zaid-Ajaj:
Hi @nelsonfigueroa thanks a lot the contribution! I am looking at the fix now, in azure-native [v3.13.0](https://github.com/pulumi/pulumi-azure-native/releases/tag/v3.13.0), specifically in the PR [Update azure-rest-api-specs to latest 2026-01-27](https://github.com/pulumi/pulumi-azure-native/pull/4496) I added a function `disambiguateTypeToken(token string)` which creates a new type token when encountering a second time.

The PR here comments in "Triggering change" section: 
> Both produce the same type token: azure-native:containerservice:UserAssignedIdentityResponse.

So this could already be fixed in latest v3.13.0+

Were you able to confirm this fix works or whether the issue repros in latest version?

nelsonfigueroa:
@Zaid-Ajaj thanks for taking a look! Yeah this is still happening in v3.15.0. I think the summary could have been better about specifying that it seems like this bug happens in v3.10.0 and onwards.

Zaid-Ajaj:
thanks @nelsonfigueroa for confirming that the issue persists in latest! We will need some tests for this, I will see if I can take care of it

PR #4591, opened 2026-03-12
Upgrade Web API version from 2024-04-01 to 2024-11-01

### LINKED ISSUE #4403: Upgrade WebApp API version to 2024-11-01 to support clientAffinityProxyEnabled
## Description

The `clientAffinityProxyEnabled` property is available in Azure's WebApp API version `2024-11-01` but is not currently available in the Pulumi Azure Native provider because the default API version for WebApp resources is `2024-04-01`.

## Property Details

- **Property Name**: `clientAffinityProxyEnabled`
- **Location**: `siteConfig.clientAffinityProxyEnabled`
- **Type**: `boolean`
- **Description**: "true to override client affinity cookie domain with X-Forwarded-Host request header. false to use default domain. Default is false."
- **Available Since**: Azure API version `2024-11-01` (released September 30, 2025)

## Current State

- **Current Default API Version**: `2024-04-01`
- **Spec Location**: `azure-rest-api-specs/specification/web/resource-manager/Microsoft.Web/AppService/stable/2024-11-01/CommonDefinitions.json:3321`
- **Version Config File**: `versions/v3.yaml`

## Requested Change

Update the default API version for WebApp resources from `2024-04-01` to `2024-11-01` in the version configuration (`versions/v3-config.yaml` or `versions/v3-spec.yaml`) to include this and other new properties introduced in the latest stable API version.

This would require:
1. Updating the version configuration to use `2024-11-01` for WebApp
2. Regenerating the schema with `make schema`
3. Regenerating the SDKs with `make generate`

## Workaround

Use `pulumi package add` to generate an SDK for the `2024-11-01` version.

## Related

- Azure REST API Specs submodule commit: `2caaf9e30d` (App Service Split #37232)
- Spec file: `specification/web/resource-manager/Microsoft.Web/AppService/stable/2024-11-01/WebApps.json`

## Impact

This affects all users who need to configure client affinity proxy settings for their Azure Web Apps. The property is particularly useful for applications behind proxies that need to override the affinity cookie domain based on the X-Forwarded-Host header.

pgavlin:
Updates the Web namespace tracking version to 2024-11-01 in versions/v3-spec.yaml.
Removes SiteCertificate and SiteCertificateSlot from additions as they are now
covered by the tracking version. Regenerates v3.yaml, all SDKs, and docs schema.

Fixes #4403.

Note that this _is_ a breaking change, but it's one I think that we should take:
we have users and customers waiting to take advantage of the new shape of this
API, and users that need the older version can use `pulumi package add`.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/examples/examples_py_test.go
+++ b/examples/examples_py_test.go
@@ -7,12 +7,9 @@ package examples
 import (
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 
 	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 func TestAccSimplePython(t *testing.T) {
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Test `TestAccWebAppWithSiteConfigPython` from `./examples/examples_py_test.go` is failing because the resources registered don't seem to have a `siteConfig` output property at runtime, failing this assertion
```go
require.Contains(t, webApp.Outputs, "siteConfig", kind)
```
I checked the schema for both `WebApp` and `WebAppSlot` via 
```bash
jq '.resources["azure-native:web:WebApp"]' bin/schema-default-versions.json
jq '.resources["azure-native:web:WebAppSlot"]' bin/schema-default-versions.json
```
Both contain an input property `siteConfig` but not in _output_ properties
```json
"siteConfig": {
      "type": "object",
      "$ref": "#/types/azure-native:web:SiteConfig",
      "description": "Configuration of an App Service app. This property is not returned in response to normal create and read requests since it may contain sensitive information."
},
```
The description mentions:
> This property is not returned in response to normal create and read requests since it may contain sensitive information

So I guess this is expected by default 🤔 I will adjust the test so that it doesn't assert things on `siteConfig` in outputs.

PR #4640, opened 2026-04-02
fix: emit null for removed IsStringSet entries and use PATCH for removal

### LINKED ISSUE #4635: CosmosDB DatabaseAccount: Removing a user-assigned identity from `user_assigned_identities` does not remove it in Azure (merge-patch semantics not handled)
### What happened?

When updating a `azure-native:cosmosdb:DatabaseAccount` resource to **remove** a user-assigned managed identity from `user_assigned_identities`, Pulumi detects the diff correctly and sends a PUT request to Azure, but **Azure does not remove the identity**. The identity remains attached to the CosmosDB account.

### Root cause

The Pulumi azure-native provider converts the `user_assigned_identities` list (an `IsStringSet` property) into a dict `{"<arm-resource-id>": {}}` in the PUT body. When an identity is removed from the list, it is simply omitted from the dict.

**This works correctly for most Azure resources** (e.g., `azure-native:sql:Server`), because Azure treats the PUT body as the complete desired state — omitted identities are removed.

**However, Azure CosmosDB uses merge-patch semantics for `userAssignedIdentities`**, even on the PUT (`CreateOrUpdate`) endpoint. Omitting an identity key means "no change" — **not** "remove". To actually remove a user-assigned identity from a CosmosDB account, you must explicitly send `{"<identity-id>": null}` in the request body.

The provider has no mechanism to generate `null` values for removed `IsStringSet` entries, so the removal is silently ignored by Azure.

### Expected behavior

Removing an identity from `user_assigned_identities` should result in that identity being removed from the Azure CosmosDB account, the same way it works for `azure-native:sql:Server`.

### Actual behavior

- Pulumi preview shows the correct diff: `- [1]: "/subscriptions/.../userAssignedIdentities/myIdentity"`
- Pulumi sends the PUT request to Azure **without** the removed identity
- Azure responds 200 OK but **does not remove the identity** (merge-patch semantics)
- `pulumi up` succeeds, but the identity is still attached to the CosmosDB account
- A subsequent `pulumi refresh` shows the identity reappearing in the state

### Affected resource

- `azure-native:cosmosdb:DatabaseAccount` (API version `2025-10-15`, also confirmed on earlier versions)

### Not affected

- `azure-native:sql:Server` — works correctly because Azure SQL treats PUT as full replacement for `userAssignedIdentities`

### Azure documentation reference

The CosmosDB `CreateOrUpdate` (PUT) docs explicitly state:
> "The 'Update' method is preferred when performing updates on an account."

The separate `Update` endpoint (`PATCH`) exists at:
`PATCH /subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.DocumentDB/databaseAccounts/{name}`

This PATCH endpoint supports sending `{"identity": {"userAssignedIdentities": {"<id>": null}}}` to remove identities.

### Workaround

We work around this by making a direct Azure REST API PATCH call after `pulumi up` to explicitly send `null` for the removed identities.

### Example

Minimal reproduction program (Python):

```python
import pulumi
import pulumi_azure_native as azure_native

# Step 1: Create a CosmosDB account with 2 user-assigned identities
# Run `pulumi up` — both identities are attached. ✅

# Step 2: Remove one identity from the list
# Run `pulumi up` — Pulumi shows the diff, update succeeds,
# but the removed identity is still attached in Azure. ❌

MSI_ID = "/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/msi-primary"
UMI_ID = "/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/umi-to-remove"

cosmosdb = azure_native.cosmosdb.DatabaseAccount(
    "my-cosmosdb",
    account_name="my-cosmosdb-account",
    resource_group_name="my-rg",
    location="westeurope",
    database_account_offer_type=azure_native.cosmosdb.DatabaseAccountOfferType.STANDARD,
    locations=[{
        "failover_priority": 0,
        "location_name": "westeurope",
    }],
    identity=azure_native.cosmosdb.ManagedServiceIdentityArgs(
        type=azure_native.cosmosdb.ResourceIdentityType.SYSTEM_ASSIGNED_USER_ASSIGNED,
        # Step 1: Both identities present
        user_assigned_identities=[MSI_ID, UMI_ID],
        # Step 2: Remove UMI_ID — change to:
        # user_assigned_identities=[MSI_ID],
    ),
)
```

### Steps to reproduce

1. Deploy with both identities in `user_assigned_identities` → `pulumi up` succeeds, both attached ✅
2. Remove `UMI_ID` from the list → `pulumi up` succeeds with diff showing removal
3. Check Azure Portal or CLI → **the removed identity is still attached** ❌
4. Run `pulumi refresh` → the removed identity reappears in Pulumi state

### Pulumi diff output (step 2)

```
~ azure-native:cosmosdb:DatabaseAccount: (update)
    ~ identity: {
        ~ userAssignedIdentities: [
            - [1]: "/subscriptions/.../userAssignedIdentities/umi-to-remove"
          ]
      }
```

The diff is correct, but Azure ignores the removal.

### What the PUT body looks like (sent by Pulumi)

```json
{
  "identity": {
    "type": "SystemAssigned,UserAssigned",
    "userAssignedIdentities": {
      "/subscriptions/.../msi-primary": {}
    }
  }
}
```

### What Azure needs to actually remove the identity

```json
{
  "identity": {
    "type": "SystemAssigned,UserAssigned",
    "userAssignedIdentities": {
      "/subscriptions/.../msi-primary": {},
      "/subscriptions/.../umi-to-remove": null
    }
  }
}
```

---

## Environment

- Pulumi CLI version: `v3.224.0`
- Pulumi Python SDK: `3.224.0`
- `pulumi-azure-native` SDK: `3.14.0`
- Python version: `3.10.12`
- OS: Ubuntu Linux (x86_64)

---

### Output of `pulumi about`

❯ pulumi about
CLI
Version      3.224.0
Go Version   go1.26.0
Go Compiler  gc

Host
OS       ubuntu
Version  22.04
Arch     x86_64

Backend
Name           rijo-vm2
URL            file:///home/coder/extollo-infra-definition/.pulumi
User           coder
Organizations
Token type     personal

Pulumi locates its logs in /tmp by default

## Potential fix suggestions

1. **Track previous `IsStringSet` values during diff/update**: When the provider detects that an `IsStringSet` entry was removed, include it in the PUT body with a `null` value instead of omitting it. This would require comparing old vs new state during `CreateOrUpdate`.

2. **Use PATCH for CosmosDB updates**: If the resource metadata for `cosmosdb:DatabaseAccount` set `UpdateMethod: "PATCH"`, the provider would use the PATCH endpoint which natively supports merge-patch semantics with explicit `null` for removals. However, this may have other side effects.

3. **Resource-specific handling**: Add CosmosDB-specific logic that detects removed `IsStringSet` entries and includes them as `null` in the body. This is similar to the existing `MaintainSubResourceIfUnset` pattern.

### Contributing

rijojohn85:
# Fix: Emit null for removed IsStringSet entries and use PATCH for removal

Fixes #4635

## Problem

When a user removes a user-assigned managed identity from a CosmosDB `DatabaseAccount` (or any resource using `IsStringSet` properties), Pulumi correctly detects the diff and sends an update — but Azure silently ignores the removal. The identity remains attached after `pulumi up`.

### Root cause

`IsStringSet` properties (e.g. `userAssignedIdentities`) are SDK arrays of strings (`["id1", "id2"]`) that the provider converts to ARM maps (`{"id1": {}, "id2": {}}`) before sending to Azure. When an entry is removed, the provider simply **omits** it from the map.

For most Azure resources, PUT is a full replacement, so omission is sufficient. However, **CosmosDB (and potentially other resources) use merge-patch semantics** on their PUT/CreateOrUpdate endpoint — omitting a key means "no change", not "remove". To remove an identity, Azure requires an **explicit `null`**:

```json
{
  "identity": {
    "userAssignedIdentities": {
      "/subscriptions/.../msi-to-keep": {},
      "/subscriptions/.../msi-to-remove": null
    }
  }
}
```

This is documented in the [Azure ARM managed identity docs](https://learn.microsoft.com/en-us/rest/api/cosmos-db/database-accounts/create-or-update) and confirmed by the error message Azure returns when null is sent via PUT (see investigation below).

### Attempt 1: Emit null in the PUT body

We threaded `previousInputs` (already available during `Update()`) through the conversion chain to the `IsStringSet` branch. When an entry is present in previous inputs but absent in new inputs, we emit `nil` (→ JSON `null`). This correctly produced the body:

```json
{
  "identity": {
    "type": "SystemAssigned,UserAssigned",
    "userAssignedIdentities": {
      "/subscriptions/.../msi_rij-3003-local61e93d23": {},
      "/subscriptions/.../rijoumi0401da0484a9": null
    }
  }
}
```

**Result:** Azure rejected with HTTP 400:

```
Status=400 Code="InvalidIdentityValues"
Message="Invalid value for the identities '...'. The 'UserAssignedIdentities'
property keys should only be empty null during PATCH requests."
```

**Conclusion:** Azure's PUT endpoint does not accept `null` for identity keys — only PATCH does.

### Attempt 2 (final): Emit null + auto-switch to PATCH

We kept the null emission logic from Attempt 1 and added automatic detection: when the request body contains `nil` values (from removed IsStringSet entries), the provider switches from PUT to PATCH.

**Why this is safe:**

1. **CosmosDB supports PATCH** — confirmed in the OpenAPI spec (`DatabaseAccounts_Update` operation) and the PATCH body schema uses the same `ManagedServiceIdentity` definition as PUT.

2. **Pulumi is declarative** — the body always contains ALL properties, so PATCH with a complete body is semantically equivalent to PUT. No properties are lost.

3. **Only triggers for identity removal** — the PATCH switch only activates when the body actually contains `nil` values, which only happens during updates where `IsStringSet` entries were removed. Creates always use PUT (no previous state → no nulls).

4. **`IsStringSet` is almost exclusively `userAssignedIdentities`** — 964 of 1006 `isStringSet: true` properties in the provider metadata are `userAssignedIdentities` on `ManagedServiceIdentity` types. This is a standard ARM pattern where PATCH is universally supported.

### Alternatives considered

| Approach | Verdict |
|----------|---------|
| **Two-call: PUT without nulls, then follow-up PATCH with nulls** | Rejected — two API calls, partial failure risk, complex body reconstruction |
| **Change metadata to `updateMethod: "PATCH"` for CosmosDB** | Rejected — too broad (affects all updates), body prepared using PUT schema |
| **Send null only in PUT** | Rejected — Azure rejects null in PUT for identity properties |

## Live testing

Tested against a real Azure CosmosDB account with two user-assigned managed identities:

| Step | Action | Result |
|------|--------|--------|
| **STEP 1** | `pulumi up` — create CosmosDB with both MSI_1 and MSI_2 | ✅ `+ 2 created`, `az cosmosdb show` confirms both identities |
| **STEP 2** | `REMOVE_SECOND_IDENTITY=true pulumi up` — remove MSI_2 | ✅ `~ 1 updated [diff: ~identity]`, `az cosmosdb show` confirms only MSI_1 remains |
| **STEP 3** | `pulumi refresh` — verify no drift | ✅ `2 unchanged` — state matches Azure |
| **Cleanup** | `pulumi destroy` + `pulumi stack rm` | ✅ Resources cleaned up |

Test environment: Azure West Europe, CosmosDB with CMK (Key Vault), SystemAssigned+UserAssigned identity type.

### `provider/pkg/convert/sdkInputsToRequestBody.go`

**What:** `SdkInputsToRequestBody` and its private helpers gain a `previousValues` parameter. In the `IsStringSet` branch, entries present in `previousValues` but absent in the new values are emitted as `nil` (→ JSON `null`).

**Why:** This is the core of the fix — it produces the explicit removal signal that Azure requires. The `previousValues` parameter is `nil` for creates (no nulls emitted) and populated during updates.

### `provider/pkg/convert/convertTypedObjects.go`

**What:** Added `convertTypedObjectWithPrev` callback type alongside the existing `convertTypedObject`.

**Why:** The recursive conversion callback needs to carry `previousValues` through nested objects so that deeply nested `IsStringSet` properties (like `identity.userAssignedIdentities`) receive the correct previous state.

### `provider/pkg/convert/sdkInputsToRequestBody_test.go`

**What:** Added `previousInputs` field to test struct; added 3 new test cases:
- `"string set removal emits null for removed entries"` — flat removal
- `"string set with no previous values emits no nulls (create path)"` — create safety
- `"nested string set removal emits null for removed entries"` — nested removal (mirrors CosmosDB structure)

**Why:** Validates the null emission logic for both removal and create paths.

### `provider/pkg/provider/crud/crud.go`

**What:**
1. `PrepareAzureRESTBody` interface method and implementation gain `previousInputs resource.PropertyMap` parameter.
2. Standalone `PrepareAzureRESTBody` function gains `previousMethodInputs map[string]any`.
3. `CreateOrUpdate` now calls `containsNilValues(bodyParams)` — if true, switches from PUT to PATCH.
4. New `containsNilValues` helper recursively checks a map for `nil` values.

**Why:** The interface change threads previous inputs to the conversion layer. The PATCH switch is needed because Azure rejects `null` in PUT for identity properties — only PATCH accepts the removal signal.

### `provider/pkg/provider/crud/crud_test.go`

**What:** Added `TestContainsNilValues` with 5 sub-tests (empty, no nils, top-level nil, nested nil, deeply nested no nil).

**Why:** Validates the nil detection logic that drives the PUT→PATCH switch.

### `provider/pkg/provider/provider.go`

**What:**
1. `defaultUpdate` signature gains `previousInputs resource.PropertyMap` and passes it to `PrepareAzureRESTBody`.
2. `Update()` call site passes `previousInputs` (already computed) to `defaultUpdate`.
3. `defaultCreate` passes `nil` for `previousInputs`.
4. Invoke call site (line ~375) passes `nil` for `previousMethodInputs`.
5. Singleton delete path (line ~1693) passes `nil`.

**Why:** `previousInputs` is already available in the `Update()` method but was never forwarded to the body conversion layer. These changes thread it through without affecting create/invoke/delete paths.

### `provider/pkg/resources/customresources/custom_pim.go`

**What:** Two `PrepareAzureRESTBody` calls updated to pass `nil` for `previousInputs`.

**Why:** PIM resources manage their own state logic. Passing `nil` preserves existing behavior.

### `provider/pkg/openapi/defaults/defaultResourceStateConvert_test.go`

**What:** One `crud.PrepareAzureRESTBody` call updated to pass `nil` for `previousMethodInputs`.

**Why:** Compile fix for the changed function signature.

## What does NOT change

- **Create path** — `nil` previousInputs means no nulls emitted, PUT used as before
- **Resources without IsStringSet removal** — no nulls in body, PUT used as before
- **Schema / metadata / versions** — no regeneration needed
- **SDK code** — no changes to any language SDK
- **Custom resources** — PIM passes `nil`, behavior unchanged

## Screenshots from local testing:
Created CosmosDB with 2 Managed identites:
image

Then removed one:
image


Zaid-Ajaj:
Hi there @rijojohn85 thanks a lot for taking the time the to contribute and raising a PR that fixes your own issue! We really appreciate it 🙇 I am looking through the implementation and added unit tests. They look like they are covering the issue scenario properly. 

We have a few issues right now running acceptance tests by external contributors, so as soon as I can, I will be running tests locally and confirm if everything is in order. 

Will keep you posted! Thanks again for raising the PR 🙏

rijojohn85:
@Zaid-Ajaj i think the acceptance tests are stuck. Can you take a look?

pose:
> @Zaid-Ajaj i think the acceptance tests are stuck. Can you take a look?

FYI, the tests have passed successfully. Here's the triggered run result: https://github.com/pulumi/pulumi-azure-native/actions/runs/24223644219 Don't look at the PR checks since they are out of sync. Thanks!

rijojohn85:
> > @Zaid-Ajaj i think the acceptance tests are stuck. Can you take a look?
> 
> 
> 
> FYI, the tests have passed successfully. Here's the triggered run result: https://github.com/pulumi/pulumi-azure-native/actions/runs/24223644219 Don't look at the PR checks since they are out of sync. Thanks!

Oh excellent. Thanks. Can we merge this if everything is fine?

Zaid-Ajaj (approved):

PR #4696, opened 2026-05-04
Retry TransientError failures when deleting `cognitiveservices:ProjectConnection`

### LINKED ISSUE #4693: cognitiveservices: ETag 409 'TransientError' on ProjectConnection delete is not retried, fails destroy
## What happened

`pulumi destroy` on a Foundry (Azure AI Services) project that owns multiple `ProjectConnection` children fails with a non-retried `409 TransientError` ETag conflict. The ARM API itself flags the response as `TransientError` (i.e. the caller is expected to retry), but `pulumi-azure-native` surfaces it as a hard failure and aborts the rest of the dependent deletions.

Re-running `pulumi destroy` succeeds. So the underlying Azure operation works fine on retry; only the provider's classification of the 409 is wrong.

## Reproduction

Stack: a Foundry account with several `azure-native:cognitiveservices:ProjectConnection` resources hanging off the same `Project` (e.g. an ACR connection, a Search connection, an APIM/Application Insights connection, etc.). Then:

```bash
pulumi destroy --yes
```

Observed during the cascading delete of the children:

```
 - azure-native:cognitiveservices:ProjectConnection search-connection
   deleting failed
   error: Status=409 Code=\"TransientError\"
          Message=\"Etag conflict on <subscriptionId>/<resourceId> with etag '<etag>'.\"
```

11 resources deleted cleanly in the same run, 2 errored on the 409, the run aborted, and a second `pulumi destroy` then completed cleanly with 54 resources removed.

Correlation ID from one occurrence: `b741cab2-6c60-4118-8368-95cce34cd745`

## Why this is a provider bug

Microsoft documents `TransientError` as an explicitly retryable failure class. The cognitive-services control plane uses an optimistic-concurrency ETag on the parent `Project` resource, and Pulumi's default parallel destroy fans out concurrent `DELETE` calls on multiple children. Concurrent deletes mutate the parent's ETag → second writer's stale ETag → 409 `TransientError`. ARM expects the caller to refresh the ETag and retry; well-behaved Azure SDKs do this transparently with backoff.

The auto-generated `pulumi-azure-native` cognitiveservices resource handler does not classify `Code=\"TransientError\"` (or the broader 409-with-ETag-message pattern) as retryable, so the failure escapes to the user.

## Expected behaviour

For 409 responses where `Code == \"TransientError\"` (or equivalently any 412 / 409 with an ETag-conflict message), the resource handler should:
1. Refresh the resource's ETag from a `GET`.
2. Retry the `DELETE` (or `PUT`/`PATCH`) with bounded exponential backoff.
3. Only surface the error after retries are exhausted.

This matches both ARM's documented retry contract and the behaviour of the Microsoft Azure SDKs that wrap the same APIs.

## Workarounds today

- `pulumi destroy --parallel=1` — serializes the children, sidesteps the race.
- Add explicit `depends_on` chains across the `ProjectConnection` resources so Pulumi orders them.
- Re-run `pulumi destroy` after the failure (the second run nearly always succeeds).

## Environment

- `pulumi-azure-native` plugin: latest at time of report
- `pulumi` CLI: 3.231+
- Azure region: `swedencentral`
- Affected resource: `azure-native:cognitiveservices:ProjectConnection`
- Parent: `azure-native:cognitiveservices:Project` on a Foundry-enabled account

cc @Zaid-Ajaj — flagging for triage. Happy to share the full Pulumi log + correlation IDs over email if helpful.

Zaid-Ajaj:
Fixes #4693

### DIFF
--- a/provider/pkg/azure/client_azcore.go
+++ b/provider/pkg/azure/client_azcore.go
@@ -138,6 +138,13 @@ func shouldRetryConflict(resp *http.Response) bool {
 			if responseErr.ErrorCode == "StorageAccountOperationInProgress" {
 				return true
 			}
+
+			// Handle deletion conflicts for resources of type cognitiveservices:ProjectConnection
+			// where deleting multiple connections at the same time can cause conflicts.
+			// see pulumi/pulumi-azure-native#4693
+			if responseErr.ErrorCode == "TransientError" {
+				return true
+			}
 		}
 	}
 	return false


dirien:
Thanks @Zaid-Ajaj, much appreciated.

One small nit: the inline comment scopes the branch to `cognitiveservices:ProjectConnection`, but the check itself fires on any 409 with `ErrorCode == "TransientError"`. That's the right behavior (other Foundry-account children hit the same ETag race), but a future reader will wonder why the comment is narrower than what the code does. Worth a wording tweak in a follow-up so the comment matches the actual scope.

PR #4711, opened 2026-05-08
Restore v3.16.0 source-compatible type-token layout

### LINKED ISSUE #4385: containerservice:ManagedCluster 3.10.0 removes properties in state from identityProfile.kubeletidentity

pgavlin:
These changes restore the v3.16.0 source-compatible token layout that #4639 broke while preserving the metadata-merge correctness fix that PR introduced.

#4639 fixed #4385 by moving types that collide across API versions into per-version `commontypesvN` namespaces. That avoided the silent property merge but moved roughly 1,500 type tokens out of their originating namespaces, breaking source compatibility for anyone on the v3.16.0 SDK. The runtime gain is the metadata-table merge, not the token relocation; these changes keep the former and revert the latter.

The codegen:

- Drops the `commontypesvN` token-rewriting path. Tokens stay in the namespace they belong to.
- Restores the original V[N] disambiguation when two API versions of the same type token disagree on shape.
- Keeps the metadata-merge fix that preserves the union of fields in the runtime metadata table.

The resulting schema is bit-identical to v3.16.0; SDK differences against that baseline are limited to package-metadata churn from intervening pulumi/pulumi codegen updates.

Note that this leaves us in essentially the same situation #4622 flagged: the schema-level pointer on a colliding property still references the default-version shape, which can be the smaller of the two. The `webAppRouting.identity` property of `ManagedCluster` exposes the 2-property `UserAssignedIdentityResponse` rather than the 3-property AKS shape; the full shape is reachable via `UserAssignedIdentityResponseV1`. The metadata merge means runtime decoding still sees the union of fields, so wire data is not lost. A proper fix needs upstream type-alias support in the schema-to-SDK generators.


Zaid-Ajaj (approved):
LGTM

PR #4720, opened 2026-05-14
Implement List in the provider protocol

Zaid-Ajaj:
### Description

This PR addresses #4716 implementing the `List(...)` in the provider server protocol. The implementation is two-fold:
 - API discovery and embedding the listable operations' metadata
 - Runtime implementation that makes use of the operations

### API Discovery
The PR implements API discovery for Listable resources from the OpenAPI specifications. Listable resources are those which have API operations with an extension `x-ms-pageable`. During API discovery, we embed the metadata of these operations into the Azure resource API metadata which are available at runtime from the `List(...)` implementation. The API metadata also gets embedded into the Pulumi schema in the `listInputs` field of the resources.

### Runtime implementation
The PR implements the streaming API in `List(...)` in the provider server. The list inputs are translated into the API query, path and request body based on API metadata from the list operation. 

Many of these listable operations require a `subscriptionId` parameter, especially in the operation path. This parameter is intentionally made _optional_ in the specs then in `List(...)` we add it ourselves based on the configured subscription ID from the provider. 

The `pageSize` provided in `*rpc.ListRequest` gets translated the OData-based parameter `$top` or one of its variants, inferred from the list parameter metadata. 

Continuation token handling uses a combination of `nextLink` returned from Azure responses of the listable operations, as well as keeping track of the number of the remaining items to be streamed. When `nextLink` is available in the continuation token from the request, we use it as is without including any more query parameters. 

The `Query` in the list request is expected to contain _plain_ values: unknowns and assets are rejected when un-marshalling.

🚧 Only the unit tests are AI-generated in this PR 🚧

- [ ] TODO: end-to-end testing from the CLI when it's available

### Remarks / Follow Ups
- The API discovery from the Azure specs includes operations for all versioned modules. Should we only include the operations for resources with the current default version?


Zaid-Ajaj:
Investigating these breaking changes now, they are not expected as part of this PR

EDIT: reverted schema breaking changes. GitHub comment now shows no changes because it doesn't yet understand the `listInputs` field in resources.

corymhall (commented):
Reviewed with Codex.

provider/pkg/provider/provider.go
```diff
@@ -468,6 +470,299 @@ func (k *azureNativeProvider) invokeResponseToOutputs(response any, res resource
...
+	if continuationToken.NextLink != "" {
+		// nextLink here is an absolute URL, parse and strip the host part
+		nextLinkURL, err := url.Parse(continuationToken.NextLink)
+		if err != nil {
+			return fmt.Errorf("parsing nextLink URL for token %s: %w", token, err)
+		}
+
+		// no need to pass an API version or query parameters when following the nextLink,
+		// as it is a fully formed URL provided by Azure
+		response, err = k.azureClient.Get(k.context, nextLinkURL.RequestURI(), "", nil)
```
corymhall: This does not preserve Azures `nextLink` as-is. `azCoreClient.Get` always starts with a fresh query map containing `api-version`, then `initRequest` rewrites `RawQuery`, so the query string already embedded in `nextLinkURL.RequestURI()` gets dropped. That loses skip tokens and can repeat the first page or fetch the wrong page.

Can we add a dedicated next-link request path, or parse the `nextLink` query back into the query params passed to the client?

created by codex

provider/pkg/provider/provider.go
```diff
@@ -468,6 +470,299 @@ func (k *azureNativeProvider) invokeResponseToOutputs(response any, res resource
...
+			return fmt.Errorf("sending list response for token %s: %w", token, err)
+		}
+		sentElementsCount++
+	}
+
+	// Send a continuation token if Azure has more pages and the caller's limit (if any) is not yet reached.
+	// When a limit is active we always set newRemaining (even to 0) so isEmptyListContinuationToken
+	// can distinguish "limit exactly met" from "no limit was set".
+	var newRemaining *int64
+	if effectiveLimit > 0 {
```
corymhall, line 766: This can still emit a continuation token after Azure has no next page. If `limit=10`, Azure returns 3 items, and `nextLinkURL == ""`, then `newRemaining` is `7`; `isEmptyListContinuationToken` treats `{NextLink: "", Remaining: 7}` as non-empty, so line 747 sends a token. The next call has no `NextLink`, falls back to the original list request, and can duplicate the same page.

Can we only carry `Remaining` when there is also a real `nextLink`, and add a regression test for `limit > len(finalPage)` with no `nextLink`?

created by codex

provider/pkg/provider/provider.go
```diff
@@ -468,6 +470,299 @@ func (k *azureNativeProvider) invokeResponseToOutputs(response any, res resource
...
+		if err != nil {
+			return fmt.Errorf("requesting list operation for token %s: %w", token, err)
+		}
+	} else if resourceAPI.ListMetadata.Method == "GET" {
+		response, err = k.azureClient.Get(k.context, id, resourceAPI.APIVersion, query)
+		if err != nil {
+			return fmt.Errorf("requesting list operation for token %s: %w", token, err)
+		}
+	} else if resourceAPI.ListMetadata.Method == "POST" {
+		responseValue, err := k.azureClient.Post(k.context, id, inputs.Mappable(), query)
```
corymhall: This sends every list input as the POST body, including path/query fields like `subscriptionId`, `resourceGroupName`, and injected page-size params. That seems different from the existing body construction path used elsewhere and can produce malformed bodies for POST pageable operations.

Can we split path/query/body parameters before calling `Post`, and add a runtime test for a POST list operation?

created by codex

provider/pkg/provider/provider.go
```diff
@@ -468,6 +470,299 @@ func (k *azureNativeProvider) invokeResponseToOutputs(response any, res resource
...
+	// continuation token as Remaining so the total across all pages stays correct.
+	var effectiveLimit int64
+	if continuationToken.Remaining != nil {
+		effectiveLimit = *continuationToken.Remaining
+	} else {
+		effectiveLimit = limit
+	}
+
+	sentElementsCount := int64(0)
+	for _, element := range elements {
```
corymhall, line 738: `page_size` is currently only forwarded as a best-effort Azure query parameter. The provider still streams until `limit`/`Remaining`, so if Azure ignores the page-size hint or the operation has no recognized page-size parameter, this can return more than the requested `PageSize`.

The provider protocol says the provider should not return more than `page_size`; can we enforce that while streaming, or otherwise avoid advertising support for operations where we cannot respect it?

created by codex

provider/pkg/provider/provider.go
```diff
@@ -468,6 +470,299 @@ func (k *azureNativeProvider) invokeResponseToOutputs(response any, res resource
...
+			return true
+		}
+	}
+	return false
+}
+
+// given an object response such as { "value": [ { ... }, { ... } ] }
+// returns the array of elements from "value"
+// assumes there is _one_ key in the response which is the array of objects returned
+func extractObjectElementsFromListResponse(response map[string]any) ([]map[string]any, error) {
```
corymhall: This scans every top-level array and treats each object in those arrays as a listed resource. `x-ms-pageable` can specify the item field, and Azure responses are not guaranteed to have only one top-level array. That makes this sensitive to unrelated array properties and map iteration order.

Can discovery carry the pageable item name and have runtime extract that specific collection instead?

created by codex

corymhall (approved):

PR #4764, opened 2026-07-15
Verify and fix Go SDK is consumable downstream

### LINKED ISSUE #4763: Go SDK v3.20.0/v3.21.0 publish a broken go.mod: pre-split google.golang.org/genproto pin breaks `go mod tidy` for every consumer
### What happened?

Starting with v3.20.0, the published `go.mod` of the Go SDK modules (verified on `resources/v3` and `storage/v3`; still present in **v3.21.0**) requires the ancient monolithic `google.golang.org/genproto v0.0.0-20200608115520-7c474a2e3482` (June 2020). That version predates the genproto module split, so it still contains the `googleapis/rpc/*` and `googleapis/api/*` packages. Any consumer whose dependency graph also pulls the split modules — which every Pulumi Go program does, via `pulumi/sdk/v3` → grpc/otel — fails `go mod tidy` with ambiguous-import errors. This means `pulumi new azure-go` fails during dependency installation (it broke CI on pulumi/templates#1160), as does any project upgrading to v3.20.0+.

v3.19.0 correctly pinned the split modules instead:

```diff
-	google.golang.org/genproto/googleapis/api v0.0.0-20260226221140-a57be14db171 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20260226221140-a57be14db171 // indirect
+	google.golang.org/genproto v0.0.0-20200608115520-7c474a2e3482 // indirect
```

`go mod graph` on a consumer project shows the azure-native-sdk modules are the *only* source of the old genproto edge.

### Example

`main.go`:

```go
package main

import (
	resources "github.com/pulumi/pulumi-azure-native-sdk/resources/v3"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		_, err := resources.NewResourceGroup(ctx, "rg", nil)
		return err
	})
}
```

`go.mod`:

```
module repro

go 1.25

require (
	github.com/pulumi/pulumi-azure-native-sdk/resources/v3 v3.21.0
	github.com/pulumi/pulumi/sdk/v3 v3.253.0
)
```

`go mod tidy` exits 1 with five errors of this shape:

```
go: repro imports
	github.com/pulumi/pulumi/sdk/v3/go/pulumi imports
	google.golang.org/grpc/status imports
	google.golang.org/genproto/googleapis/rpc/status: ambiguous import: found package google.golang.org/genproto/googleapis/rpc/status in multiple modules:
	google.golang.org/genproto v0.0.0-20200608115520-7c474a2e3482 (.../googleapis/rpc/status)
	google.golang.org/genproto/googleapis/rpc v0.0.0-20260401024825-9d38bb4040a9 (.../rpc@v0.0.0-20260401024825-9d38bb4040a9/status)
```

Changing the require to v3.19.0 makes `go mod tidy` and `go build` succeed.

### Workarounds

- Stay on v3.19.0.
- Or, before `go mod tidy`, run `go mod edit -require=google.golang.org/genproto@v0.0.0-20240213162025-012b6fc9bca9` (any post-split monolithic version); tidy then succeeds and prunes the pin.

### Output of `pulumi about`

Not relevant — this fails at `go mod tidy`, before any Pulumi engine involvement. Reproduced on macOS (darwin/arm64) and all three OS runners in pulumi/templates CI.

Zaid-Ajaj:
### Fix broken Go SDK go.mod

Fixes #4763

Since v3.20.0, the published Go SDK modules pin the pre-split, monolithic `google.golang.org/genproto` instead of the split `genproto/googleapis/{api,rpc}` modules that grpc/otel need. Any consumer whose own graph also needs the split modules (i.e. every Pulumi Go program) fails `go mod tidy` with ambiguous-import errors.

**Root cause:** the Go SDK's `go.mod` is regenerated from scratch every release with no prior `go.sum`, starting from a hardcoded `pulumi/sdk/v3` version in the codegen template that hadn't been touched since 2022 (`v3.37.2`). `go mod tidy` then has to re-derive the entire modern grpc/otel/genproto dependency graph from that 4-year-old anchor against whatever's on the proxy at that moment - bridging that large a gap is exactly where `go mod tidy`'s package resolution can non-deterministically land on the wrong (pre-split) `genproto` module instead of the split ones. Nothing in this repo changed between the working v3.19.0 and broken v3.20.0 releases; the drift is entirely external.

**CI didn't catch it** because `make build_go` only tidies each module against itself via local `replace` directives to sibling directories in this checkout - that stays green even when the go.mod is broken for real consumers, since the ambiguity only shows up once a downstream consumer's own requirements combine with ours during module resolution.

### Changes

- Bump the hardcoded `pulumi/sdk/v3` baseline in the `goModTemplate` (codegen) from `v3.37.2` to `v3.253.0`, a current release. This closes the gap `go mod tidy` has to bridge on every release, so it never needs to "rediscover" a resolution for the ambiguous genproto packages - `v3.253.0`'s own `go.mod` already requires the split modules directly. Verified this alone (no genproto pinning needed) produces a clean `go.mod` with no trace of the monolithic `genproto` module, across all 255 generated modules.
- Add `make verify_go_sdk_consumable`, which builds a throwaway module requiring the generated SDK the way a real downstream consumer would (fresh `go.mod`, no shared `go.sum`) and runs `go mod tidy && go build` against it, as a regression guard. Wired into `build-test.yml` after `make build_go`.

Verified locally: the template bump produces a clean `go.mod` across all 255 modules and all build successfully, and the new consumer check reproduces the issue's exact ambiguous-import errors when tested against a broken SDK, and passes against the fixed one.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -133,6 +133,14 @@ jobs:
         if: ${{ matrix.language != 'yaml' }}
         run: make build_${{ matrix.language }}
 
+      # Regression check for https://github.com/pulumi/pulumi-azure-native/issues/4763. `make build_go`
+      # above only tidies/builds the generated modules against themselves (via local `replace`
+      # directives), which can stay green even when the generated go.mod is broken for real consumers.
+      # This builds a throwaway module that requires the generated SDK the way a real user would.
+      - name: Verify Go SDK is consumable downstream
+        if: ${{ matrix.language == 'go' }}
+        run: make verify_go_sdk_consumable
+
       - name: Check worktree clean
         uses: pulumi/git-status-check-action@0d90c81496aa8d6a31d9c6a0d6297feea2f54057 # v2
         with:
--- a/.github/scripts/go-sdk-consumer-check-main.go
+++ b/.github/scripts/go-sdk-consumer-check-main.go
@@ -1,0 +1,16 @@
+// Used by `make verify_go_sdk_consumable` to check that the generated Go SDK can be tidied and built
+// by a downstream consumer. See the target's comment in the Makefile and
+// https://github.com/pulumi/pulumi-azure-native/issues/4763 for why this exists.
+package main
+
+import (
+	resources "github.com/pulumi/pulumi-azure-native-sdk/resources/v3"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func main() {
+	pulumi.Run(func(ctx *pulumi.Context) error {
+		_, err := resources.NewResourceGroup(ctx, "rg", nil)
+		return err
+	})
+}
--- a/Makefile
+++ b/Makefile
@@ -16,16 +16,6 @@ endif
 
 JAVA_GEN        := pulumi-java-gen
 
-# Seed version for google.golang.org/genproto used before `go mod tidy` runs on the generated Go SDK.
-# The genproto module was split into google.golang.org/genproto/googleapis/{api,rpc} - any version of
-# genproto older than the split still ships those packages too, and if `go mod tidy` ever picks the
-# pre-split module for one of them (which it can do non-deterministically, since the SDK's go.mod/go.sum
-# are regenerated from scratch on every release with no prior lockfile to anchor the resolution), every
-# downstream consumer's own `go mod tidy` fails with "ambiguous import" errors. Pinning to a version of
-# genproto published after the split guarantees `go mod tidy` only ever resolves the split modules. See
-# https://github.com/pulumi/pulumi-azure-native/issues/4763.
-GENPROTO_POST_SPLIT_PIN := v0.0.0-20240213162025-012b6fc9bca9
-
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 GOEXE ?= $(shell go env GOEXE)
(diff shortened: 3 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -133,6 +133,14 @@ jobs:
         if: ${{ matrix.language != 'yaml' }}
         run: make build_${{ matrix.language }}
 
+      # Regression check for https://github.com/pulumi/pulumi-azure-native/issues/4763. `make build_go`
+      # above only tidies/builds the generated modules against themselves (via local `replace`
+      # directives), which can stay green even when the generated go.mod is broken for real consumers.
+      # This builds a throwaway module that requires the generated SDK the way a real user would.
+      - name: Verify Go SDK is consumable downstream
+        if: ${{ matrix.language == 'go' }}
+        run: make verify_go_sdk_consumable
```
pose, line 142: Could we use https://github.com/pulumi/verify-provider-release instead and avoid modifying the Makefile? Thanks!

pose (changes requested):

.github/workflows/build-test.yml
Zaid-Ajaj: I still need to modify the Make file to ensure we are using a pinned version of `genproto` (the cause of the issue), this only refactors the verifiication logic, correct?

pose:
Discussed offline and @Zaid-Ajaj found a way to not require the `Makefile` dependency pinning. I am approving, so it can be merged once the change is pushed.

pose (approved):

PR #4783, opened 2026-08-03
Fix #4782: don't send siteConfig.http20ProxyFlag's spec default of 0

### LINKED ISSUE #4782: Http20ProxyFlag Being incorrectly set when null
### What happened?

Currently we're using the [Functions on Container Apps](https://learn.microsoft.com/en-us/azure/container-apps/functions-overview) capability, but the [V1 version](https://learn.microsoft.com/en-us/azure/container-apps/migrate-functions) that lives under the `Microsoft.Web` namespace which means using the [`Pulumi.AzureNative.Web.WebApp`](https://www.pulumi.com/registry/packages/azure-native/api-docs/web/webapp/) resource to create the Azure resource.

Using Pulumi.AzureNative provider version <= `3.15.0` we were successfully able to create this resource type and deploy to Azure. However since upgrading to >= `3.16.0` of the provider, we are now seeing the following error being returned:

```text
  azure-native:web:WebApp (my-webapp):
    error: Status=400 Message="{"Code":"BadRequest","Message":"Http20ProxyFlag is invalid.  Http20ProxyFlag is not supported for Azure Functions on Azure Container apps. Please retry the operation without Http20ProxyFlag.","Target":null,"Details":[{"Message":"Http20ProxyFlag is invalid.  Http20ProxyFlag is not supported for Azure Functions on Azure Container apps. Please retry the operation without Http20ProxyFlag."},{"Code":"BadRequest"},{"ErrorEntity":{"ExtendedCode":"51021","MessageTemplate":"{0} is invalid.  {1}","Parameters":["Http20ProxyFlag","Http20ProxyFlag is not supported for Azure Functions on Azure Container apps. Please retry the operation without Http20ProxyFlag."],"Code":"BadRequest","Message":"Http20ProxyFlag is invalid.  Http20ProxyFlag is not supported for Azure Functions on Azure Container apps. Please retry the operation without Http20ProxyFlag."}}],"Innererror":null}"
```

This suggests we're setting the [Http20ProxyFlag](https://www.pulumi.com/registry/packages/azure-native/api-docs/web/webapp/#http20proxyflag_csharp) when creating the resource, but upon review of our code this property is not being specified so should be defaulting to a null value.

Unfortunately it looks like even if we set this to null the provider today defaults this to `0` and submits this as part of the REST API request, triggering this error to occur. 

This code corroborates this theory that it is being set on the request.
https://github.com/pulumi/pulumi-azure-native/blob/9236e100b698d6ee24b36644c76d94f0bf57a03a/sdk/dotnet/Web/Inputs/SiteConfigArgs.cs#L513

The REST API expects this property to be omitted or set to null to consider it not being set. Is there a way that this can be set? Or does this require a change to support this ability?

### Example

```csharp
// Some settings omitted for brevity
new Web.WebApp("myapp", new Web.WebAppArgs
        {
            Location = "westeurope",
            ResourceGroupName = "my-resourcegroup"
            Kind = "functionapp,linux,container,azurecontainerapps",
            ManagedEnvironmentId = containerAppEnvId,
            SiteConfig = new WebInputs.SiteConfigArgs
            {
                // Http20ProxyFlag = null, // Explicitly set to null also has no effect
                MinimumElasticInstanceCount = 1,
                LinuxFxVersion = $"DOCKER|mcr.microsoft.com/mcr/hello-world",
            },
        });
```

### Output of `pulumi about`

N/A

### Additional context

_No response_

### Contributing

MeganYTan:
## Summary
- `siteConfig.http20ProxyFlag` on `WebApp`/`WebAppSlot` carries an OpenAPI spec default of `0`, which every SDK then sent on create/update even when the user left the property unset.
- Azure Functions on Container Apps (kind `functionapp,linux,container,azurecontainerapps`) rejects the property outright, so those apps failed to deploy with a `Http20ProxyFlag is invalid` 400 error, and users had no workaround since an explicit `null` input is indistinguishable from an omitted one.
- Adds a `noDefaultMap` in `provider/pkg/gen/replacement.go` (parallel to the existing `caseInsensitiveDiffMap` pattern) so specific module/resource/property combos can have their spec-declared default dropped during codegen, and wires it into `properties.go`. Scoped to `Web.WebApp`/`Web.WebAppSlot` `http20ProxyFlag` only — no change to default-handling behavior elsewhere.
- Regenerated the docs schema and Node.js/Python/.NET SDKs to drop the `0` default for this property.

Fixes #4782.

## Test plan
- [x] `go test ./provider/pkg/gen/...` passes, including new `TestNoDefault` unit test
- [x] `go build ./...` succeeds for the provider
- [x] Verified regenerated schema/SDK diffs are scoped to `http20ProxyFlag`'s default only (schema.json, dotnet `SiteConfigArgs.cs`, nodejs `input.ts`/`output.ts`, python `_inputs.py`/`outputs.py`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/provider/pkg/gen/properties.go
+++ b/provider/pkg/gen/properties.go
@@ -414,6 +414,13 @@ func (m *moduleGenerator) genProperty(name string, schema *spec.Schema, context 
 		}
 	}
 
+	// Some spec-declared defaults are values Azure itself rejects, so drop them and leave the
+	// property unset unless the user assigns it explicitly (see noDefaultMap).
+	if defaultValue != nil && m.noDefault(name) {
+		logging.V(5).Infof("Dropping default value '%v' for property %q per noDefaultMap\n", defaultValue, name)
+		defaultValue = nil
+	}
+
 	// If there's no object value type, then it's just a set of strings which we'll represent as a string
 	// array in the SDK, but leave the metadata to indicate we need to convert it.
 	isStringSet := typeSpec.Type == "object" && typeSpec.AdditionalProperties == nil && typeSpec.Ref == ""
@@ -596,6 +603,17 @@ func (m *moduleGenerator) enumOverride(propertyName string) (EnumOverride, bool)
 // notRequiredInputsMap.
 func (m *moduleGenerator) notRequired(propertyName string) bool {
 	if resourceMap, ok := notRequiredInputsMap[m.moduleName]; ok {
+		if properties, ok := resourceMap[m.resourceName]; ok {
+			return properties.Has(propertyName)
+		}
+	}
+	return false
+}
+
+// noDefault returns true if the OpenAPI spec's default value for propertyName should be dropped
+// on the resource currently being generated, per noDefaultMap.
+func (m *moduleGenerator) noDefault(propertyName string) bool {
+	if resourceMap, ok := noDefaultMap[m.moduleName]; ok {
 		if properties, ok := resourceMap[m.resourceName]; ok {
 			return properties.Has(propertyName)
 		}
--- a/provider/pkg/gen/properties_test.go
+++ b/provider/pkg/gen/properties_test.go
@@ -114,6 +114,20 @@ func TestCaseInsensitiveDiff(t *testing.T) {
 	assert.False(t, otherModule.caseInsensitiveDiff("backendPoolType"))
 }
 
+func TestNoDefault(t *testing.T) {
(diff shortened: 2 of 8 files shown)


Zaid-Ajaj (approved):
Looks good to me. I was already looking at a similar fix when you opened the PR 😄

